PR #624, opened 2014-06-03
allow DCAP configuration to be set per flavour

calestyo:
Currently, the defaults file forbid to set per-protocol-flavour options for
“dcap.authz.mover-queue-overwrite”, “dcap.mover.queue” and
“dcap.limits.clients”.

I do not see any good reason why this shouldn’t be allowed, actually this is
highly desired, since admins for example want to default all their mover queues
for GSI DCAP to “gsidcap” while all others should perhaps default to “dcap”.
When those options are marked immutable, an admin can’t set e.g. site-wide
defaults in dcache.conf.

• Drop the “immutable” flag for those DCAP per-protocol-flavour options where
  it seems approppriate.

Require-book: no
Require-notes: yes

Signed-off-by: Christoph Anton Mitterer mail@christoph.anton.mitterer.name

### DIFF
--- a/skel/share/defaults/dcap.properties
+++ b/skel/share/defaults/dcap.properties
@@ -68,10 +68,10 @@ dcap.authz.staging=${dcache.authz.staging}
 # Whether a dcap client is allowed to override which mover queue on
 # the pool is used to schedule a request.
 (one-of?true|false|${dcap.authz.mover-queue-overwrite.${dcap.authn.protocol}})dcap.authz.mover-queue-overwrite=${dcap.authz.mover-queue-overwrite.${dcap.authn.protocol}}
-(immutable)dcap.authz.mover-queue-overwrite.plain=${dcap.authz.mover-queue-overwrite-when-${dcapIoQueueOverwrite}}
-(immutable)dcap.authz.mover-queue-overwrite.auth=${dcap.authz.mover-queue-overwrite-when-${dcapIoQueueOverwrite}}
-(immutable)dcap.authz.mover-queue-overwrite.gsi=${dcap.authz.mover-queue-overwrite-when-${gsidcapIoQueueOverwrite}}
-(immutable)dcap.authz.mover-queue-overwrite.kerberos=${dcap.authz.mover-queue-overwrite-when-${kerberosdcapIoQueueOverwrite}}
+dcap.authz.mover-queue-overwrite.plain=${dcap.authz.mover-queue-overwrite-when-${dcapIoQueueOverwrite}}
+dcap.authz.mover-queue-overwrite.auth=${dcap.authz.mover-queue-overwrite-when-${dcapIoQueueOverwrite}}
+dcap.authz.mover-queue-overwrite.gsi=${dcap.authz.mover-queue-overwrite-when-${gsidcapIoQueueOverwrite}}
+dcap.authz.mover-queue-overwrite.kerberos=${dcap.authz.mover-queue-overwrite-when-${kerberosdcapIoQueueOverwrite}}
 (immutable)dcap.authz.mover-queue-overwrite-when-allowed=true
 (immutable)dcap.authz.mover-queue-overwrite-when-denied=false
 (deprecated,one-of?allowed|denied)dcapIoQueueOverwrite=denied
@@ -94,20 +94,20 @@ dcap.net.listen=${dcache.net.listen}
 
 # Pool mover queue to use for transfers by this door
 dcap.mover.queue=${dcap.mover.queue.${dcap.authn.protocol}}
-(immutable)dcap.mover.queue.plain=${dcapIoQueue}
-(immutable)dcap.mover.queue.auth=${dcapIoQueue}
-(immutable)dcap.mover.queue.gsi=${gsidcapIoQueue}
-(immutable)dcap.mover.queue.kerberos=${kerberosdcapIoQueue}
+dcap.mover.queue.plain=${dcapIoQueue}
+dcap.mover.queue.auth=${dcapIoQueue}
+dcap.mover.queue.gsi=${gsidcapIoQueue}
+dcap.mover.queue.kerberos=${kerberosdcapIoQueue}
 (deprecated)dcapIoQueue=
 (deprecated)gsidcapIoQueue=
 (deprecated)kerberosdcapIoQueue=
 
 # Maximum number of concurrent DCAP transfers on this door
 dcap.limits.clients=${dcap.limits.clients.${dcap.authn.protocol}}
-(immutable)dcap.limits.clients.plain=${dcapMaxLogin}
-(immutable)dcap.limits.clients.auth=${dcapMaxLogin}
-(immutable)dcap.limits.clients.gsi=${gsidcapMaxLogin}
-(immutable)dcap.limits.clients.kerberos=${kerberosdcapMaxLogin}
+dcap.limits.clients.plain=${dcapMaxLogin}
+dcap.limits.clients.auth=${dcapMaxLogin}
+dcap.limits.clients.gsi=${gsidcapMaxLogin}
+dcap.limits.clients.kerberos=${kerberosdcapMaxLogin}
 (deprecated)dcapMaxLogin=1500
 (deprecated)gsidcapMaxLogin=1500
 (deprecated)kerberosdcapMaxLogin=1500


calestyo:
Hey.

I've just stumbled over that again...

The current state seems to be still quite inconsistent...

For the movers, the old per-flavour settings are apparently gone now, but:

- for the cell names we have per-flavour settings for dcap/ftp, but not for webdav (which is kinda annoying, when one wants to add two webdav doors http/https in one domain)
- same for the ports, we have them for ftp/dcap, but not for webdav...

- it looks similar for options that give authn/authz rights (things like webdav.authn.require-client-cert. webdav.authz.readonly, webdav.authz.anonymous-operations, webdav.authz.anonymous-listing, dcap.authz.readonly, dcap.authz.anonymous-operations)... no per flavour-settings for these, but there are some for ftp read-only.


I've also reconsidered your original arguments, and I'm not longer sure, whether I agree, at least not from a practical point of view:
If we really say, all the different protocol flavours are basically the same, and there is not reason why they should have per-flavour settings for say the mover queue... then the same must in principle apply to the port, cell name, etc.
Sure, different ports are used for different flavours, but this is just convention, and even there we don't follow those conventions in most cases (webdav plain and webdav TLS port, FTP plain and KrbFTP port, etc.)

When there is reason for the ports to have per flavour settings, then IMHO the same would need to apply to most of the authn/z settings from above, because those you also most likely want to set, depending on your flavour (plain dcap => read only yes, anon-oper=read-only,... but gsi: readonly no, anon-oper=NONE).
And as for the movers, at least I know many sites which use different queues based on the flavour, based on the argument that e.g. gsi is more resource intensive than plain... so they want to limit the queues there a bit.


Right now I've made per-flavour options for all of these, but I think it's a bit unhandy and it seems to me that this is a general use case.
Of course there's a difficult question: where to draw the line.

Maybe one needs a syntactical way to specify the arguments dependent on other options value, e.g. something like
case(dcap.authn.protocol)
{
  plain:
  dcap.authz.readonly=yes
  ;;
  gsi:
  dcap.authz.readonly=no
}

gbehrmann:
This is only an issue because you want more than one instance of a flavour on a single host, right? Otherwise you could just specify the per instance settings in the layout file (like we do at NDGF).

gbehrmann:
As for the naming, I think it would make sense to redo the naming scheme for doors to include the port number rather than some flavour specific string. That way it will automatically be unique on the given host.

calestyo:
Well I guess it's a matter of how one wants to design the site local configuration.
My idea is basically that, ... uhm... difficult to explain ^^ ... I guess I want to follow several paradigms:
- config files should be small and readable (which they are not, if I have to repeat everything per service instance)

- the option should be set at the "canonical place",.. with that I basically mean: when I configure webdav.plain=readonly but webdav.tls=readwrite ... than this is a global decision and not something that I individually decide again for each door (of course there can be situations where people really do just that, decide things individually per door, say one door is publicly visible and has allowed-paths=/public while one in the intranet has allowed-paths=/ ... in such case, I really do it *for* specific doors).
But for most of mover-queue/access control settings/port numers and for most doors it's more a globally/site-wide motivated decision,i.e. the decision "if you're unauthenticated, you should only get RO access".
For that reason, such settings belong to the dcache.conf and not to a per door setting.

- layout files, which are per se a rather unfortunate construct (design wise) since I feel we should have only one dache.conf and this hast just include statements to include other config files (which could then be typically layouts,... layout files are for me like Lego Parts, which I use to build up my site.
With that I basically mean, simple things without much additional knowledge (config-wise) that per default just do what they should most likely do (with respect to our site), i.e. chose the right ports, and so on.
The only things I'd like to set there are node specific settings (e.g. memory constraints) or when I really want to set something for that specific instance of service... e.g. my personal ftp door which allows access to Chris' secret /pnfs/dcache-C++-fork
Ideally they layouts should be so simple, that I can tell my mother "add a new gsiftp door" and what she does works out.


As for your question, no these problems are when I have multiple flavours of the same protocol on one host (e.g. dcap + gsidcap).



And then there's the bad guy bug #8611:
Your original suggestion was "just make up your own per flavour settings", which seemed like a good idea back then (even if I'd say in the meantime it's rather just work intensive, since many sites will do that over and over again)... but as for now I did just that, e.g.:
webdav.authn.require-client-cert=${lmu.webdav.authn.require-client-cert.${webdav.authn.protocol}}
lmu.webdav.authn.require-client-cert.http=false
lmu.webdav.authn.require-client-cert.https=true
lmu.webdav.authn.require-client-cert.https-jglobus=${lmu.webdav.authn.require-client-cert.https}
#TODO: “/robots.txt” shouldn’t be writable (and thus not added to the value of the option “lmu.access-control.accepted-paths”), but as of now this can’t be configured for WebDAV doors without also making it non-readable. Once WebDAV doors support separately configured acceptable read and write paths, make “/robots.txt” non-writable.
webdav.authz.allowed-paths=${lmu.access-control.accepted-paths}:/robots.txt
webdav.authz.readonly=${lmu.webdav.authz.readonly.${webdav.authn.protocol}}
lmu.webdav.authz.readonly.http=true
lmu.webdav.authz.readonly.https=false
lmu.webdav.authz.readonly.https-jglobus=${lmu.webdav.authz.readonly.https}
webdav.authz.anonymous-operations=${lmu.webdav.authz.anonymous-operations.${webdav.authn.protocol}}
lmu.webdav.authz.anonymous-operations.http=READONLY
lmu.webdav.authz.anonymous-operations.https=NONE
lmu.webdav.authz.anonymous-operations.https-jglobus=${lmu.webdav.authz.anonymous-operations.https}
webdav.authz.anonymous-listing=${lmu.webdav.authz.anonymous-listing.${webdav.authn.protocol}}
lmu.webdav.authz.anonymous-listing.http=true
lmu.webdav.authz.anonymous-listing.https=false
lmu.webdav.authz.anonymous-listing.https-jglobus=${lmu.webdav.authz.anonymous-listing.https}
(looks a bit complicated, but the "lmu." is just our "personal namespace" to avoid any possible future conflicts)

But of course I got hit by the bug I've already reported my self some days ago and what happens with the above is:
# dcache check-config 
[ERROR] dcache.conf:292: Property webdav.authn.require-client-cert:
                         "${lmu.webdav.authn.require-client-cert.${webdav.authn.protocol}}"
                         is not a valid value. Must be one of "true",
                         "false"
[ERROR] dcache.conf:298: Property webdav.authz.readonly:
                         "${lmu.webdav.authz.readonly.${webdav.authn.protocol}}"
                         is not a valid value. Must be one of "true",
                         "false"
[ERROR] dcache.conf:302: Property webdav.authz.anonymous-operations:
                         "${lmu.webdav.authz.anonymous-operations.${webdav.authn.protocol}}"
                         is not a valid value. Must be one of "NONE",
                         "READONLY", "FULL"
[ERROR] dcache.conf:306: Property webdav.authz.anonymous-listing:
                         "${lmu.webdav.authz.anonymous-listing.${webdav.authn.protocol}}"
                         is not a valid value. Must be one of "true",
                         "false"
[ERROR] dcache.conf:368: Property dcap.authz.readonly:
                         "${lmu.dcap.authz.readonly.${dcap.authn.protocol}}"
                         is not a valid value. Must be one of "true",
                         "false"
[ERROR] dcache.conf:373: Property dcap.authz.anonymous-operations:
                         "${lmu.dcap.authz.anonymous-operations.${dcap.authn.protocol}}"
                         is not a valid value. Must be one of "NONE",
                         "READONLY", "FULL"
Found 6 errors.


So basically one is really back to: Having to set all these options per flavour, per door instance.
Kinda error-prone, and some effort whenever one value is to be changed (for all).

calestyo:
As for your second comment:

I've already started thinking about this as well... actually I think all cell names should be freshed up (especially SrmSpaceManager), following consistent rules (e.g. all lower case or so), I have now:
spacemanager.cell.name=SpaceManager
nfs.cell.name=nfs_${host.name}
webdav.cell.name=${lmu.webdav.cell.name.${webdav.authn.protocol}}
lmu.webdav.cell.name.http=webdav.plain_${host.name}
lmu.webdav.cell.name.https=webdav.tls_${host.name}
lmu.webdav.cell.name.https-jglobus=${lmu.webdav.cell.name.https}
ftp.cell.name.plain=ftp.plain_${host.name}
ftp.cell.name.kerberos=ftp.kerberos_${host.name}
ftp.cell.name.gsi=ftp.gsi_${host.name}
dcap.cell.name.plain=dcap.plain_${host.name}
dcap.cell.name.auth=dcap.auth_${host.name}
dcap.cell.name.kerberos=dcap.kerberos_${host.name}
dcap.cell.name.gsi=dcap.gsi_${host.name}
xrootd.cell.name=xrootd_${host.name}
srm.cell.name=srm_${host.name}

So for the doors I have protocol.flavour_<host>.

(And I did away with the "Domain" of all my domains)

Adding the port came to my mind as well, since I could have more than one e.g. http.plain door per node, for now I decided against, but mainly because we never have the case that multiple doors of the same flavour run on the same host on different ports - we just have different flavours per host, where each already has it's well-known port... and I'm to lazy to write the port every time in the admin interface ;)

But for default values I'd agree that the port should be included.

Whether the flavour should be dropped instead is another question: It's simply quite handy to see the flavour just by the name.

calestyo:
btw: wrt bug #8611, dCache is really cheeky here.
I cannot even redfine the annotations in dcache.conf to include the custom option, because then dCache complains:
[ERROR] dcache.conf:293: Property webdav.authn.require-client-cert: remove
                         "(one-of?true|false|${lmu.webdav.authn.require-client-cert.${webdav.authn.protocol}})";
                         annotated assignments are not allowed

:-P
:-P

PR #1210, opened 2014-12-22
Db split defaults

calestyo:


gbehrmann:
As I mentioned in email before, I don't see much point in defaulting to multiple user accounts for the different databases. I suggest you introduce a common dcache.db.user and dcache.db.password property that all services default to.

Before the patch can be merged, a transition document has to be published (or included in the release notes). That's why this change has not been done before - it's not like changing the defaults is difficult. It is telling everybody how to change their setup (or tell them to add additional configuration to keep the old values) that is the challenge.

calestyo:
Well, using only one user/owner, has the disadvantage that we have yet another place where we either divert from the current schema (chimera/chimera) or where people would have to adapt there (chimera) as well.
Apart from that as dCache uses DBs I don't really see any disadvantage in using different user names, and in the end (e.g. if unix sockets should ever be allowed) it would probably be the more powerful solution because one can easily map the same POSIX user (dcache) to different PostgreSQL user.

As for the transition document... I've posted my guidelines on the list one one can refer to them in the release notes.
And even if there wouldn't be such document, it's probably better to give new installations the better setup and tell legacy sites that they need to adapt the config to the legacy layout.

gbehrmann:
Thanks.

Anybody else having an opinion on user accounts? I don't have a strong opinion on this. It's just that I don't really see a need for the separation of accounts when there are no barriers internally in dCache (dCache services do not authenticate with each other).

kofemann:
In general, I will prefer to have a single db for all databases used by
dcache separated by different  schema. Such setup will allow to decrease
the overhead on managements as well as clear separation of tables used by
different components.

Related to patch.
To me it looks like we introduce more overhead for fresh installation. If
an admin wants to have db/user/pass per service he/she still can do that.
We can document it and describe the advantages, but keep basic installation
as simple as possible.

Tigran.

-kofemann

gbehrmann:
Which would be nice if different RDBMS would treat schemas the same. Specifically in Oracle schema is a different thing.

kofemann:
And postgres JDBC driver does not support it at all :)

-kofemann

calestyo:
@gbehrmann: It's not that I'd have a strong opinion on the owner/user issue... if I'd have to choose from both, I'd just find the solution with separate owners a tiny bit cleaner.


@kofemann: And having everything in on DB, with tables/triggers/functions etc. all mixed, where users have no idea which belongs to what service?
At least in the past this didn't work out quite well, since the naming was so unfortunate, that one cannot really tell. And it makes it basically impossible for the average user to safely move their services (with the related DBs) to different hosts, if they wish or need to.
Moving a full DB however is something that seems to be very simple pg_dump + psql ... no hidden secrets.

What's the reason that you prefer having everything in one DB?

calestyo:
@kofemann: Oh and related to difficulties in fresh installations... it's just creating more users/DBs in the beginning,... nothing that any user should fail with (they already create chimera user and DB).

gbehrmann:
@calestyo, regarding the suggestion from @kofemann, using a different schema for each service would solve that. Each service would have it's own schema, thus grouping all the tables, indexes, functions, triggers, etc that belongs to that service. It would have the added benefit that you can easily do queries involving tables from multiple schemas, while doing this for multiple databases is quite difficult.

gbehrmann:
@calestyo, check out http://www.postgresql.org/docs/9.4/static/ddl-schemas.html if you haven't used schemas before.

calestyo:
No, I(!) know schemas,... and dumping is not really that much more difficult with it (pg_dump -n foo)... I just thought it might be less obvious for novice users.

Anyway,.. I still think that splitting the DBs has other advantages which I've laid out in the split guidelines on user-forum.
And at least dCache itself cannot rely on advantages  (like the queries you've mentioned  or  automatic locking or making atomic transactions on tables affecting multiple services) anyway, since they might as well be on different nodes.

So I'd welcome if you introduce schemas,... but I think that's independent from the discussions of whether the DBs should be split or not.

gbehrmann:
Yes, obviously dCache cannot rely on the cross schema queries, but it is useful for the admin. E.g. I have used it to get rid of orphaned entries in the space manager database. Anyway, given that @kofemann reported that schemas are not supported by the JDBC driver, the discussion about schemas is mute anyway.

calestyo:
So... the only open question is probably whether to use different owners or not.
As said, it's not that I'd have a very strong opinion on that either, since changing the owner is rather easy doable for people - the only thing more complicated is the split itself, which is why I think we should get *this* through ASAP - actually I think you should also encourage people to do during 2.6.

Anyway, wrt the owner question... I've just though it's subtly more "powerful" for possible future setups to have different owners per DB - but nothing really concrete, so we could also change that - but then we should do so for chimera/chimera as well.

calestyo:
+1

PR #1255, opened 2015-01-22
webdav: Alternative to fixing return code of DELETE of absent file

gbehrmann:
Patch https://rb.dcache.org/r/7614 fixed an issue with the return code
we return when attempting to delete files that do not exist. That patch
had the side effect that TLS authenticated session used the string form
of the DN as the tag value of the Milton Auth object. When backported
to 2.6, this caused problems with the old LoggingFilter as it assumes
the tag is a Subject.

This patch implements an alternative to patch 7614. It avoids the use
of a dummy X509AuthenticationHandler and instead sets a pre-fabricated
Auth object no matter whether Milton has created such an object
already or not. This is in line with our architecture in which we
handle all authentication outside of Milton.

Target: trunk
Request: 2.11
Request: 2.10
Request: 2.9
Request: 2.8
Request: 2.7
Request: 2.6
Require-notes: yes
Require-book: no
Ticket: http://rt.dcache.org/Ticket/Display.html?id=8584
Ticket: http://rt.dcache.org/Ticket/Display.html?id=8583
Acked-by: Dmitry Litvintsev <litvinse@fnal.gov>
Patch: https://rb.dcache.org/r/7709/
(cherry picked from commit 49298736b6bb51440b16a1597ad23d86e15fb74b)


paulmillar:
Patch currently broken

paulmillar:
retest this please

PR #1278, opened 2015-02-06
Fix/2.11/jquery selector

alrossi:


alrossi:
You will need to apply 1277 first.

PR #1363, opened 2015-03-09
spacemanager: Controlled shutdown

gbehrmann:
Since the thread pool was shut down asynchronously, requests in the queue could
be processed after the database connection pool was closed. The was a
potentially large number of stack traces in the logs when shutting down space
manager while it was busy.

This refactors the thread pool handling such that the tasks of the space
manager service can be drained before the connection pool is closed. The space
manager classifies requests into important requests and unimportant requests.
The latter are returned to the sender with an error, while the former are
processed as loosing them is undesirable.

The patch also fixes the use of an srm property inside the space manager
configuration.

Target: trunk
Require-notes: yes
Require-book: yes
Request: 2.12
Request: 2.11
Request: 2.10
Patch: https://rb.dcache.org/r/7947/
(cherry picked from commit 5e56f3832bf4535badb6d2c1d2468a629b83dd8b)


paulmillar:
retest this please

[Note to Gerd; please ignore this -- it is for Jenkins]

PR #1464, opened 2015-04-17
admin: Fix connection shutdown

gbehrmann:
It is desirable that the shell completes the last command even when the
input stream shuts down. This patch ensures that the shell thread is not
prematurely interrupted.

Target: trunk
Request: 2.12
Require-notes: yes
Require-book: no
Acked-by: Christian Bernardt <christian.bernardt@desy.de>
Patch: https://rb.dcache.org/r/8120/
(cherry picked from commit 1ee6afe17e980fe867cdf6f6d16afa4e3e25a0c8)


paulmillar:
retest this please

PR #1465, opened 2015-04-17
admin: Avoid division by zero when the client reports a zero sized terminal

gbehrmann:
The jline error generates an arithemtic error when faced with a terminal
of width 0. The ssh client reports such a width when forced to allocate
a pseudo TTY without having a TTY (i.e. when using the -t -t options with
redirected stdin).

This patch avoids the error by reinterpreting the 0 width/height as meaning
an arbitrarily large terminal (i.e. no line breaks etc).

Target: trunk
Request: 2.12
Request: 2.11
Request: 2.10
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/8121/
(cherry picked from commit 085c335446b9efeab9120a322ee0ba8b574f0577)


paulmillar:
retest this please

PR #1608, opened 2015-06-03
srm: Fix cache invalidation of space meta data

gbehrmann:
Motivation:

The SRM caches space meta data. This cache needs to be invalidated if a client
does something that modifies the meta data. Releasing spaces is done
asynchronously with a callback, yet the cache invalidation happens in the
forward call and thus may happen before the space is actually released. Thus
there is a race in which the cache could be repopulated with the old data
before the space is released in space manager.

Modification:

Intercept the callback and invalidate the cache inside the callback.

Result:

The cache is invalidated after the space is released, thus resolving the
described race condition.

Target: trunk
Require-notes: yes
Require-book: no
Request: 2.12
Request: 2.11
Request: 2.10
Patch: https://rb.dcache.org/r/8273/
(cherry picked from commit c15a133be42385cbe0c2c8d986231b016f410b77)


paulmillar:
please retest this

paulmillar:
retest this please

PR #1616, opened 2015-06-04
pool: reduce load on back-end file system

kofemann:
the nfs read request uses file size and current position to detect and
report EOF marker. As a result any file READ request will issue File#length().
Reduce the load on backend file system by caching file size for read-only
movers.

Small refactoring to use FileChannel only instead of FileChannel + File + RandomAccessFile.

acked-by: Gerd Behrmann
Target: master, 2.12
Require-notes: no
Require-book: no
(cherry picked from commit d1c977dee1cb8bf973e2d10a8b5ae78b02600fd2)


paulmillar:
Patch currently breaks at least FTP transfers; tested on sisyphus and with system-test.  The cause is currently not known, but results in pool disabling itself during the initial transfer.  The symptoms are consistent with the pool creating the file then attempting to exclusive-create the file, which then fails.

Issue #1638, opened 2015-06-10
In web admin interface, max stores and max restores columns are zero instead of the real value


onnozweers:
In dCache 2.10, when we look at the web admin interface pool request queues, for instance here:
http://dcmain.grid.sara.nl:2288/webadmin/queueinfo
... we always see zeroes in the 'max stores' and 'max restores' columns, whatever the value in the pool is. The 'act' and 'que' columns seem to have the correct values.

The old style web interface has the same issue:
http://dcmain.grid.sara.nl:2288/queueInfo

In 2.6 this worked as expected.

gbehrmann:
Not a bug. The max values no longer exist as far as dCache is concerned. Queues for nearline storage are unbounded. Any limit is an implementation detail of a specific nearline storage plugin. This was noted in the 2.9.0 release notes:

> Since nearline storage request queuing is now no longer part of dCache, the maximum active requests limits are no longer reported to poolmanager. For some drivers, such a limit may not even exist. Pool selection has been updated to no longer rely on these values.

Thus the web pages also do not pick those values up anymore.

onnozweers:
Hi Gerd, 

Thanks for explaining. Then may I suggest the columns are removed instead of showing misleading information? Even showing a dash instead of a zero would be better, IMHO.

Kind regards
Onno

gbehrmann:
Okay, marking as a request for enhancement instead.

gbehrmann:
Al, could you have a look at applying the same fix to webadmin?

alrossi:
@gbehrmann Sure, Gerd.  But not immediately.  I can perhaps have a look this weekend.  OK?

gbehrmann:
Sure, no hurry.

alrossi:
patch https://rb.dcache.org/r/8338 @gbehrmann

gbehrmann:
@alrossi please submit pull requests for 2.10 to 2.13.

alrossi:
I have added null guards to a few places.

PR #1703, opened 2015-06-22
(2.10) webadmin: do not display numerical value for max restores or s…

alrossi:
…tores

        webadmin complement to https://rb.dcache.org/r/8319

        Follows same solution (does not remove column).

        Added: null pointer fixes.

        Target: 2.11
        Ticket: #1638
        Require-notes: yes
        Require-book: no
        Acked-by:  Gerd Behrmann


alrossi:
Sorry, that should read Target: 2.10

PR #1724, opened 2015-06-30
pool: Fix race condition in migration module

### LINKED ISSUE #1722: Pool migration command fails with 'No targets' if destination already contains the file
The following affects at least 2.12. When copying files like this:

    migration copy -target=pool -- pool_read

the command fails and sleeps whenever it runs into a file that already exists on the target pool:

    12:10:05 [4] 000007A86FED654D4664A0FE65CC968195AE: No targets

The expected behaviour is that the file is skipped/updated.

In case multiple targets are available, the migration module will create superfluous replicas instead of producing an error.

gbehrmann:
Motivation:

A regression was introduced by the addition of the -replicas option. A race
condition will in some cases cause additional replicas to be created. In other
cases it will cause failures with the error "No targets". The trigger is when
the file already exists on the target pool in combination with a particular
message ordering.

Modification:

The cause is that we add the target pool to the list of confirmed replicas
after receiving the PoolMigrationCopyReplicaMessage reply. If
PoolMigrationCopyFinishedMessage overtakes the reply, the subsequent check
whether additional replicas are needed will be erroneous.

The fix is to move the addition to the list to the point at which we receive
PoolMigrationCopyFinishedMessage. The state machine only checks the result
after receiving this message.

Result:

Fixes #1722.

Target: trunk
Request: 2.13
Request: 2.12
Request: 2.11
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/8369/
(cherry picked from commit cd7637d239906c85990eb28aceb527e38c7159fc)

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/pool/migration/Task.java
+++ b/modules/dcache/src/main/java/org/dcache/pool/migration/Task.java
@@ -197,6 +197,7 @@ public class Task
         messageArrived(PoolMigrationCopyFinishedMessage message)
     {
         if (_uuid.equals(message.getUUID())) {
+            _replicas.add(message.getPool());
             _fsm.messageArrived(message);
         }
     }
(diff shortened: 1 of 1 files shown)


paulmillar:
retest this please

paulmillar:
retest this please

paulmillar:
retest this please

paulmillar:
retest this please

paulmillar:
retest this please

PR #1768, opened 2015-07-17
set a default path for admin.paths.history + introduce dcache.host.http option+ enable DH modes again

calestyo:
Now that the admin shell detects whether a terminal is connected or not and that
the history file is automatically created if it doesn’t exist, it makes no
longer sense to have the history disabled per default.

• Set “/var/lib/dcache/admin/history” as default value for the option
  “admin.paths.history”.
• Adapt the documentation to describe how the history can be disabled.


calestyo:
There seems to be no nice way to reuse an existing ${dcache.paths.XXX} here :-(

calestyo:
+ Can one of the admins reconfigure dcache-ci to no spam github?

paulmillar:
@calestyo could you update this pull request?  It no longer applies cleanly to master.

The pull request contains three unrelated requests, which is sub-optimal: one patch is no longer needed; one is acceptable as-is (factoring out HTTP host); one needs further work (admin path history).  By placing unrelated work in the same pull request the part that could be taken is blocked by unrelated work.

Suggest you close this pull request and create two new ones: one for each of the patches that are still valid, updating

PR #1804, opened 2015-08-12
cells: Fix dumpster and default route removal

gbehrmann:
Motivation:

Route removal for dumpster and default routes fail to work as compareAndSet
used in the code is by reference rather than equality.

Modification:

Explicitly call equals to compare the existing route with the route to be
removed.

Result:

Route removal for -dumpster and -default routes works.

Target: trunk
Require-notes: yes
Require-book: no
Request: 2.13
Request: 2.12
Request: 2.11
Request: 2.10
Acked-by: Femi Adeyemi <olufemi.segun.adeyemi@desy.de>
Patch: https://rb.dcache.org/r/8466/
(cherry picked from commit decafd2f94f50b0b3884c6375e77ccdb14fe9438)


paulmillar:
retest this please

PR #1831, opened 2015-09-08
srm: Cache the vomsdir and ca stores

gbehrmann:
Motivation:

The SRM uses a generic utility class in gPlazma2 to extract FQANs from a
certificate chain. This utility class repeatedly reloads the files in vomsdir
and the ca certificates. In particular the latter is associated with
significant overhead.

This is a regression in 2.13.

Modification:

Embeds the FQAN extraction code directly into the relevant places in the SRM,
which allows us to cache the store material. The utility class is removed
entirely as the only users were the SRM and the DCAP doors.

Result:

Faster SRM and fewer dependencies between our modules.

Target: trunk
Require-notes: no
Require-book: no
Request: 2.13
Patch: https://rb.dcache.org/r/8474/
(cherry picked from commit 56edf342e7bfcfd51f184acf42703c8f5088eb7b)


paulmillar:
retest this please

PR #1878, opened 2015-10-12
(2.13) webadmin: improve user-friendliness of alarms page on load

alrossi:
Motivation:

When the alarms page is opened, refresh() is immediately called because Ajax autorefresh
is set for this service; owever, no default limits are given to the query range,
and if there are many entries in the database, the page can take rather long to load.
This is not user friendly and could be mistaken for a network issue of some sort.

Modification:

This patch adopts two small changes which should ameliorate this.

First, an autorefresh option has been incorporated into the BasePage API,
and exposed on Alarms page as a query checkbox option which is by default
off.   Hence, the page will load without automatically refreshing.

Second, we have also provided a default limit:

    to (range): a limit of 100 results is initially returned

The default end date is undefined (meaning the most recent event); events
are sorted by timestamp descending.

The type default is still set to Alarms, but this does not affect the query issued
to the underlying store, only the in-memory filter.

Result:

A more modest amount of time is initially required to connect, and
the page refresh need not refetch by default a maximum of 10000 entries.

Target: 2.13
Acked-by: Gerd
Require-book: yes
Require-notes: yes

### DIFF
--- a/modules/dcache-webadmin/src/main/java/org/dcache/webadmin/view/pages/alarms/AlarmsPage.java
+++ b/modules/dcache-webadmin/src/main/java/org/dcache/webadmin/view/pages/alarms/AlarmsPage.java
@@ -106,7 +106,7 @@ public class AlarmsPage extends BasePage implements AuthenticatedWebPage {
             }
         };
 
-        Form<?> form = getAutoRefreshingForm("alarmsPageForm");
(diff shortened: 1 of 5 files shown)


alrossi:
Paul,

There is a slight modification to the user interface.   I hadn't considered this before.  Perhaps backport is not correct here.   If so, we can just discard.

Al

PR #2145, opened 2016-01-31
webdav: mapping of sudden owncloud-related paths to specific hardcode…

IvanUgrin:
…d json dictionaries

Motivation:

The OwnCloud Sync Client requers sudden endpoints to indicate properties, capabilities and status of the server. To simulated OwnCloud server on dCache these endpoints are required in order for  the client to establish a successful connection.

Modification:

Add a jetty handler to the server, which detects the user agents and responds with an OwnCloud-specific response if the agent is OwnCloud's.

Result:

OwnCloud can establish a successful connection on the WebDAV door of dCache.

Target: trunk
Require-notes: no
Require-book: no
Patch:

### DIFF
--- a/modules/dcache-webdav/pom.xml
+++ b/modules/dcache-webdav/pom.xml
@@ -96,5 +96,10 @@
         <groupId>org.italiangrid</groupId>
         <artifactId>voms-api-java</artifactId>
     </dependency>
+    <dependency>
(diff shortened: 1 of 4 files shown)


paulmillar:
ok to test

paulmillar:
retest this please

Issue #1635, opened 2015-06-09
storage group names with '/'-sign break statistics service


ahaupt:
Dear dCache support,

we created some storage groups with a pattern like this (from poolmanager.conf):

```
psu create unit -store  ifh:magic15/testosm:testgroup@osm
```

This seems to break the statistics service (dCache 2.10.31) as it does not seem to properly mask the '/'. Excerpt from httpDomain (runs inside that domain here):

```
 [puck] /var/log/dcache % tail -n 12 httpdDomain.log
 09 Jun 2015 14:05:52 (PoolStatistics) [] Can't prepare Html directory: /var/lib/dcache/statistics/2015/06/09
 java.io.FileNotFoundException: /var/lib/dcache/statistics/2015/06/09/class-ifh:magic15/testosm:testgroup@osm.html (No such file or directory)
    at java.io.FileOutputStream.open(Native Method) ~[na:1.7.0_79]
    at java.io.FileOutputStream.<init>(FileOutputStream.java:221) ~[na:1.7.0_79]
    at java.io.FileOutputStream.<init>(FileOutputStream.java:171) ~[na:1.7.0_79]
    at java.io.FileWriter.<init>(FileWriter.java:90) ~[na:1.7.0_79]
    at diskCacheV111.services.PoolStatisticsV0.prepareDailyHtml(PoolStatisticsV0.java:1266) [dcache-core-2.10.31.jar:2.10.31]
    at diskCacheV111.services.PoolStatisticsV0.prepareDailyHtmlFiles(PoolStatisticsV0.java:1126) [dcache-core-2.10.31.jar:2.10.31]
    at diskCacheV111.services.PoolStatisticsV0.access$1100(PoolStatisticsV0.java:119) [dcache-core-2.10.31.jar:2.10.31]
    at diskCacheV111.services.PoolStatisticsV0$HourlyRunner.run(PoolStatisticsV0.java:511) [dcache-core-2.10.31.jar:2.10.31]
    at dmg.cells.nucleus.CellNucleus$5.run(CellNucleus.java:742) [cells-2.10.31.jar:2.10.31]
    at java.lang.Thread.run(Thread.java:745) [na:1.7.0_79]
```

Cheers,
Andreas

gbehrmann:
There are likely many other places in which this would break. The easy solution would be to disallow `/` in storage classes. Opinions?

paulmillar:
[...]

PR #2172, opened 2016-02-11
statistics: encode '/' in filenames

paulmillar:
The statistics service creates static HTML pages that describe dCache
usage over time as simple files that the webadmin service can serve.
Currently the files used to store summary information about pools and
store-units derive their filename directly from the pool or store-unit
names.

Unfortunately pools and store-units may have characters that cannot be
represented as filenames in the underlying storage system.  With
POSIX-like filesystems, the '/' character is not allowed.

This patch adds a simple encoding: any '/' character is mapped to the
"^s^" sequence and any '^' is mapped to a "^o^" sequence.

NB: this patch results in the statistics module loosing track of pools
or store-units that have a '^' in the name.

Target: master
Require-notes: yes
Require-book: no
Request: 2.14
Request: 2.13
Request: 2.12
Request: 2.11
Request: 2.10
Bug: https://github.com/dCache/dcache/issues/1635
Patch: https://rb.dcache.org/r/8315/
Acked-by: Gerd Behrmann

### DIFF
--- a/modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
+++ b/modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
@@ -2,6 +2,9 @@
 
 package  diskCacheV111.services ;
 
+import com.google.common.escape.Escaper;
+import com.google.common.escape.Escapers;
+import com.google.common.net.UrlEscapers;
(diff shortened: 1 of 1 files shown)


paulmillar:
retest this please

paulmillar:
retest this please

PR #2173, opened 2016-02-11
pool: send connection header when closing a connection.

paulmillar:
HTTP/1.0 allows the client to request a connection is kept open. With
HTTP/1.1 this became the default, but a client could request the
connection is closed once the server has completed sending the
response.

The pool honours this and will close the connection if it detects
this is appropriate: HTTP/1.0 request without Connection: Keep-Alive
header, or HTTP/1.1 request with Connection: Close.

RFC 2616 (8.1.2.1) states "If the server chooses to close the
connection immediately after sending the response, it SHOULD send a
Connection header including the connection-token close."

Currently dCache sends no Connection header in the response when
shutting down the connection.  Although not required by RFC 2616, we
have no good reason to deviate from recommended behaviour.
Therefore, this patch updates the pool to include the connection
response when it will close the connection.

Target: master
Patch: https://rb.dcache.org/r/8754
Acked-by: Gerd Behrmann
Requires-notes: yes
Requires-book: no
Request: 2.14
Request: 2.13
Request: 2.12
Request: 2.11
Request: 2.10


paulmillar:
retest this please

paulmillar:
retest this please

PR #2175, opened 2016-02-12
webdav: Decorate the PathMapper in order to provide the desired WebDAV door mapping for specific(OwnCloud Sync client) User-agents.

IvanUgrin:
Motivation:

The OwnCloud Sync Client requers expects the WebDAV door to be at serverContext/remote.php/webdav so it requests files with this prefix. Milton has to be fooled that the files exist at this paths and basiclly relocate the door at remote.php/webdav without milton and owncloud understanting about it. Redirects and rewrites won't work.

Modification:

Creation of AbstractPathMapper and decorator in order to modify the original PathMapper if certain configuration is present.

Result:

OwnCloud Sync client synchronises as expected.

Target: trunk
Require-notes: no
Require-book: no


paulmillar:
ok to test

PR #2237, opened 2016-03-02
poolmanager: Acquire read lock when serializing pool selection unit

gbehrmann:
Motivation:

Java serialization is not thread safe, yet pool selection unit
currently does not protect against modifications while it is
serialized.

Modification:

Acquire a read lock during serialization.

Result:

Fixes a race condition in pool manager that could feed erroneous
data to pin manager, space manager, srm, xrootd and webdav.

Target: trunk
Require-notes: yes
Require-book: no
Request: 2.15
Request: 2.14
Request: 2.13
Patch: https://rb.dcache.org/r/9075/
(cherry picked from commit be1d495da5195ab3bf3bce083b272c4989e40a78)


mksahakyan:
retest this please

PR #2273, opened 2016-03-11
authorisation: reintroduce ReadOnly class

paulmillar:
Motivation:

The SRM records which user is associated with long-lived requests by
serialising information about the user.  Currently, master and 2.15 do
not have the ReadOnly class, which prevents deserialisation that
information; and, in turn, prevents the srm service from starting.

Modification:

Reintroduce the ReadOnly attribute, with support in SRM for mapping
this to a Restriction if the user has read-only authorisation.

Result:

srm service will start when dCache has existing user requests; the
ReadOnly attribute is still enforced.

Target: master
Patch: https://rb.dcache.org/r/9114/
Acked-by: Gerd Behrmann
Requres-notes: yes
Requires-book: no
Request: 2.15

### DIFF
--- a/modules/common/src/main/java/org/dcache/auth/attributes/ReadOnly.java
+++ b/modules/common/src/main/java/org/dcache/auth/attributes/ReadOnly.java
@@ -1,0 +1,67 @@
+package org.dcache.auth.attributes;
+
+import java.io.Serializable;
+
+/**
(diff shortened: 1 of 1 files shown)


mksahakyan:
retest this please

mksahakyan:
retest this please

PR #2277, opened 2016-03-14
chimera: Fix regression in inheriting ACLs on directory creation

gbehrmann:
Motivation:

2.13 introduced a regression which caused directories to inherit
ACLs as if they were files rather than directories.

Modification:

Partially revert f88ecf06676b064df24ada38ad57a64c02d1e199.

Result:

Fixes a regression in directory creation in which the newly created
directory would inherit ACLs as if it were a file rather than a
directory.

The bug was a regression in 2.13 and is not present in previous or
later releases. That is, 2.14 and newer are fine. I request backport
for these branches only to support correct schema downgrade.

Target: trunk
Request: 2.15
Request: 2.14
Request: 2.13
Require-notes: yes
Require-book: no
Ticket: http://rt.dcache.org/Ticket/Display.html?id=8915
Patch: https://rb.dcache.org/r/9129
(cherry picked from commit e2d611349a9abddcc4b9729036f614b74d662d5b)


mksahakyan:
retest this please

mksahakyan:
retest this please

PR #2323, opened 2016-04-03
fix regression from 859431218e3d7cdddcf906144c1b8928bb625fed

calestyo:
Commit 859431218e3d7cdddcf906144c1b8928bb625fed introduced a regression and
breaks dcache database showUpdateSQL because at printf(1) doesn’t allow “--”.

Target: master
Target: trunk
Require-notes: yes
Require-book: no
Request: 2.15

### DIFF
--- a/skel/bin/dcache
+++ b/skel/bin/dcache
@@ -837,7 +837,7 @@ case "$1" in
                     for cell in $(getProperty dcache.domain.cells "$domain"); do
                         if [ $# -eq 0 ] || matchesAny "$cell@$domain" "$@"; then
                             if hasManagedDatabase "$domain" "$cell"; then
-                                printf "-- %s: \n" "$cell@$domain"
+                                printf -- "-- %s: \n" "$cell@$domain"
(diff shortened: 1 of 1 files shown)


kofemann:
the idea behind original commit it to by showUpdateSQL generate an output, which can be applied directly to db:
```
 $ psql < output.of.showUpdateSQL
```
this requires that we include '--' when commenting out non SQL information provided by output.

kofemann:
how about fool the print command:

 printf "\055\055 %s: \n" "$cell@$domain"

where \055 is the ascii code for '-'

calestyo:
Ah I see... then a more readabl way would perhaps be:
printf -- "-- %s: \n" "$cell@$domain
with the first "--" ending option parsing and the second being literal "--" from the format string.

kofemann:
sounds good. Could you please update the pull requests?

calestyo:
updated..

kofemann:
thanks

PR #2381, opened 2016-04-27
    Motivation:

DmitryLitvintsev:
A significant slowdown of inserts into billinginfo has been
    observed on the new CMS disk instance running 2.13. Comparing
    previous 2.2 and new 2.13 it was discovered that in the absence
    of p2p (in particular) transfers a trigger that populates aggeregate
    data for p2p transfers would execute a rather heavy query on each
    insert into billinginfo. That is, because underlying select of p2p
    transfer summary returns no records, no records are made in aggregation
    table, and therefore a natural sentinel (time of last insert) is not
    present and select scans the whole billinginfo table for p2p. This
    repeats until a p2p transfer is made. If such transfer never made,
    the querty keeps spinning on each insert into billinginfo.

    Modification:

    Fixed all triggers that generate aggregate data by
    checking result of select and storing a record with
    0 values and current timestamp.

    Result:

    Significant improvement of insert performance into billinginfo table.
    (tested on CMS dCache instance)

    Target: trunk
    Request: 2.15
    Request: 2.14
    Request: 2.13
    Request: 2.12
    Require-notes: no
    Require-book: no
(cherry picked from commit 641aca8a6987db3eb643f49db73ff69369ff3dc7)

### DIFF
--- a/modules/dcache/src/main/resources/org/dcache/services/billing/db/sql/billing.changelog-2.6.xml
+++ b/modules/dcache/src/main/resources/org/dcache/services/billing/db/sql/billing.changelog-2.6.xml
@@ -402,7 +402,7 @@
                       oldColumnName="totaltime"
                       tableName="billinginfo_tm_daily"/>
     </changeSet>
-    <changeSet id="6.4.1" author="arossi" context="billing">
+    <changeSet id="6.4.1.1" author="litvinse">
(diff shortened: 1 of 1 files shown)


paulmillar:
retest this please

paulmillar:
retest this please

PR #2417, opened 2016-05-18
pool: Don't log illegal state exception on migration job cancellation as a bug

gbehrmann:
Motivation:

When cancelling a job in a state that doesn't allow cancellation, an illegal
state exception is thrown. This is currently logged as a bug.

Modification:

Declare the exception such that it is not logged as a bug.

Result:

Avoids logging an exception on migration module cancellation as a bug.

Target: trunk
Require-notes: yes
Require-book: no
Request: 2.15
Request: 2.14
Request: 2.13

Reviewed at https://rb.dcache.org/r/9274/

(cherry picked from commit 0ca62f6de1b6f33ed91ed786a1357f3957e1c81d)


mksahakyan:
retest this please

PR #2427, opened 2016-05-18
pool: Fix p2p cancellation

gbehrmann:
Motivation:

Pool to pool transfers are supposed to be cancellable, but this currently
doesn't work as the HttpURLConnection does not appear to react to thread
interrupt. The consequence is that one may observe pool to pool transfers
on slow transfers hang around for a long time even though the requesting
node already asked for it to be cancelled. E.g. migration job cancellation
may appear to hang due to this as well as rebalance job cancellation.

Modification:

Port the P2P companion code to Apache HTTP Client. This client supports
explicit cancellation. Also made use of the checksum channel support to
avoid the custom checksum calculation in the p2p code.

As a side effect we may hope that this fixes the problem of incomplete
transfers sometimes observed with p2p transfers, but since I cannot reproduce
that problem, it is hard to say if this patch would fix it.

Result:

Fixes a problem in the pool to pool transfer code that prevented cancellation
of transfers. This could lead to migration job and rebalance job cancellation
appearing to hang.

Target: trunk
Request: 2.15
Request: 2.14
Request: 2.13
Require-notes: yes
Require-book: no
Acked-by: Albert Rossi <arossi@fnal.gov>

Reviewed at https://rb.dcache.org/r/9295/

(cherry picked from commit 8a8687caaf9e1493b269fed1f33e8ece18cbf233)

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/pool/p2p/Companion.java
+++ b/modules/dcache/src/main/java/org/dcache/pool/p2p/Companion.java
@@ -1,24 +1,31 @@
 package org.dcache.pool.p2p;
 
+import com.google.common.base.Optional;
+import com.google.common.io.ByteStreams;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
+import org.apache.http.HttpEntity;
+import org.apache.http.HttpHeaders;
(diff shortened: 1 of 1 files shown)


mksahakyan:
retest this please

PR #2435, opened 2016-05-19
pnfsmanager: Move flush notification out of chimera transaction

gbehrmann:
Motivation:

Upon receiving a flush message, pnfsmanager adds the tape location to
chimera and then generates asynchronous flush notifications for space
manager. These flush notifications are submitted asynchronously so
the transaction doesn't block, but there is no reason to send them
from within the transaction. Worse, the transaction commit may fail
and in that case we should not send the flush notifications to space
manager.

Modifiction:

Adds a post processing step for pnfs messages. The flush notification
is generated during post processing.

Result:

Reduces transaction length of flush processing in pnfs manager.

Target: trunk
Request: 2.15
Require-notes: yes
Require-book: no
Acked-by: Albert Rossi <arossi@fnal.gov>

Reviewed at https://rb.dcache.org/r/9296/

(cherry picked from commit f4dcb5bade1039e90731b70b6c60c1c29a6e82a9)

### DIFF
--- a/modules/dcache/src/main/java/diskCacheV111/namespace/PnfsManagerV3.java
+++ b/modules/dcache/src/main/java/diskCacheV111/namespace/PnfsManagerV3.java
@@ -1,5 +1,6 @@
 package diskCacheV111.namespace;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Splitter;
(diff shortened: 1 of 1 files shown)


mksahakyan:
retest this please

PR #2455, opened 2016-05-25
httpd: restore original fields and format to the TransferObserverV1 a…

alrossi:
…scii output

Motivation:

A previous patch 9258 introduced several changes to the ascii (text) output
for active transfer information.  As this is consumed by scripts, it is
better to leave it as it was for backward compatibility.

Modification:

Restore the original number of fields and the formatting of several of the
strings to what it was before that patch.

Result:

Backward compatibility is maintained.

Target: master
Request: 2.16
Require-notes: no
Require-book: no
Acked-by: Gerd


mksahakyan:
retest this please

mksahakyan:
retest this please

PR #2456, opened 2016-05-26
srmmanager: Make scheduler ID configurable and compatible with older versions

gbehrmann:
Motivation:

SrmManager tags all records with a scheduler ID and only processes those with its
own tag. In replicated setups, each instance sharing a database must have its
own scheduler ID. Furthermore, when upgrading from an earlier version, one should
try to maintain the scheduler ID to ensure that old transfers get cleaned up.

Modification:

Introduced the srmmanager.scheduler-id property and let it default to
SRM-${host.name} for compatibility with older versions. This also likely makes
it unique in replicated deployments.

Result:

Introduced the srmmanager.scheduler-id property.

Target: trunk
Request: 2.16
Require-notes: yes
Reqiure-book: yes

Reviewed at https://rb.dcache.org/r/9333/

(cherry picked from commit f7511cbf0280dc840cc2923bd9de779635c66070)

### DIFF
--- a/modules/dcache-srm/src/main/resources/diskCacheV111/srm/srmmanager.xml
+++ b/modules/dcache-srm/src/main/resources/diskCacheV111/srm/srmmanager.xml
@@ -450,7 +450,7 @@
           init-method="start" destroy-method="stop">
         <description>Scheduler for GET operations</description>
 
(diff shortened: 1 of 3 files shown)


mksahakyan:
retest this please

mksahakyan:
retest this please

Issue #2476, opened 2016-05-30
httpd throws NPE on config error


gbehrmann:
When upgrading to dCache 2.16 with a custom `httpd.conf`, if one forgets to adjust the configuration to include `-cell=${httpd.service.info}` in the info handler, the `httpd` service throws an NPE and the info handler fails to load.

PR #2486, opened 2016-05-31
httpd: Fail gracefully in case of missing options

gbehrmann:
Motivation:

When upgrading to dCache 2.16 with a custom httpd.conf, if one forgets to
adjust the configuration to include -cell=${httpd.service.info} in the info
handler, the httpd service throws an NPE and the info handler fails to load.

Modification:

Add a precondition check to ensure that the option was provided. Refactored
the httpd service to propagate instantiation failures and distinguish between
exceptions indicating bugs and exceptions indicating other errors.

Result:

31 maj 2016 10:09:22 (httpd) [] context:httpdSetup: line 21: Command failed
((3) java.lang.IllegalArgumentException: -cell option is required for
InfoHttpEngine handler. from ac_set_alias_$_3_16; caused by: -cell option is
required for InfoHttpEngine handler.)

Target: trunk
Request: 2.16
Require-notes: no
Require-book: no
Fixes: #2476

Reviewed at https://rb.dcache.org/r/9355/

(cherry picked from commit 5b92ef434694a86b2a69f89185f5ae326d60d0a8)

### DIFF
--- a/modules/dcache-info/src/main/java/org/dcache/services/info/InfoHttpEngine.java
+++ b/modules/dcache-info/src/main/java/org/dcache/services/info/InfoHttpEngine.java
@@ -35,6 +35,7 @@ import org.dcache.services.info.serialisation.XmlSerialiser;
 import org.dcache.util.Args;
 import org.dcache.vehicles.InfoGetSerialisedDataMessage;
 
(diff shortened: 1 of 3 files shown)


mksahakyan:
retest this please

PR #2532, opened 2016-06-08
Fix/2.16/location selection failure

alrossi:


alrossi:
This patch depends on https://github.com/dCache/dcache/pull/2531

PR #2573, opened 2016-06-21
webdav: avoid NPE if client fails to send a User-Agent header

paulmillar:
Motivation:

Not all user-agents send a User-Agent header in their requests, yet
dCache webdav door fails with a NPE if not.  Duplicity is apparently
one such client.

Modification:

Update code to check if user-agent is the OwnCloud sync-client to be
robust against a client not sending the header.

Result:

No more NPEs.

Target: master
Request: 2.16
Request: 2.15
Patch: https://rb.dcache.org/r/9444/
Acked-by: Gerd Behrmann
Ticket: http://rt.dcache.org/Ticket/Display.html?id=8998


paulmillar:
retest this please

paulmillar:
retest this please

PR #2593, opened 2016-06-27
resilience: eliminate unnecessary update calls and fix synchronizatio…

alrossi:
…n of file op registration

Motivation:

Patch 9451 made adjustments to the way a duplicated
file operation updates the extant one.  It did so
in the FileOperationMap's 'add' method.

However, the two entry points for a new file operation
were also pre-checking the existence of the operation
before actually processing the update.  This was
being done in order to avoid the database call to
get file attributes.

Now that the logic of the update has changed slightly,
this simple counter increment is not correct.

Moreover, the lack of synchronization (the underlying
call to the index fetches from a ConcurrentMap) was
based on the idea that we would not get into serious
trouble if simultaneous calls were sliced such
that both actually end up adding an operation (the
second overwriting the first).

However, it is possible for this to confuse the
operation counts on simultaneous pool scans (it
happens rarely, but recent testing revealed a few
instances).

Modification:

First, the extra call (and its updateCount method)
have been eliminated.  Testing shows that there
is no significant performance loss in doing so.
In fact, the collision is most likely to occur
between two handleScanned calls, and the ingest
rates there are already masked by the slower
select query.

Second, the add method has now made the check-and-add
sequence into an atomic critical section, as it
normally should have been.  Once again, testing
reveals no noticeable drop in performance.

Result:

Correct update code, and race condition eliminated.

Target: master
Request: 2.16
Acked-by: Gerd

### DIFF
--- a/modules/dcache-resilience/src/main/java/org/dcache/resilience/data/FileOperationMap.java
+++ b/modules/dcache-resilience/src/main/java/org/dcache/resilience/data/FileOperationMap.java
@@ -76,6 +76,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
+import javax.ws.rs.HEAD;
 
 import diskCacheV111.util.CacheException;
 import diskCacheV111.util.PnfsId;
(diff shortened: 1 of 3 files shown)


alrossi:
Depends on #2590

alrossi:
The failure is due to the dependency of this patch on a previous one.

jstarek:
Retest this please

alrossi:
Jürgen, this suggests to me that this patch has not been applied on top of 2590.  There is a missing overloaded method and the call at that line cannot match the existing method signatures.

I don't know what to do about pull requests that depend on each other.  I once tried posting one patch built on another, but that seemed not to work ...

Let me know what to do about this (i.e., how to resolve the dependency).  

Thanks, Al

PR #2628, opened 2016-06-30
SrmShell.java: changed 'rw--' to 'rw-' in case TPermissionMode._RW

onnozweers:
I was searching for directory right code to investigate why we have a directory that spontaneously changes from 755 to 770 every now and then. Then I stumbled upon what appears to be a coding typo. I have no idea if it is related to our bug, but I'd like to bring it to your attention anyway.


paulmillar:
Good catch!  Thanks for the patch.

It affects the new SRM shell, which is unrelated code, but it's good to have it fixed.

PR #2661, opened 2016-07-26
alarms: add ndc info to alarm info

alrossi:
Motivation:

Sometimes information, like the pnfsid, is missing from a logging
message which we promote to alarm, but the information is present
in the MDC (NDC):

rw-stkendca13a-5Domain.log:20 Jul 2016 09:30:40 (rw-stkendca13a-5) [b55b5e9b-baa0-4613-b56c-e10c32d71d19 000013A9AA57B8684A4AA5E407F9A0F5E795] HSM script failed (script reported: 1: ...

This information can be usefully displayed in the
webadmin table for diagnostic purposes.

Modification:

Instead of appending the information on the client end (to the
event message), we allow the server-end to do so.

Result:

The missing diagnostic info is displayed in the table.

While not a bug, the improvement is minimally invasive, affects
only the displayed element, and would be useful to have.
Hence I am requesting a backport.
I leave it to the reviewers' discretion.

Target: master
Acked-by: Paul

### DIFF
--- a/modules/common/src/main/java/org/dcache/commons/util/NDC.java
+++ b/modules/common/src/main/java/org/dcache/commons/util/NDC.java
@@ -2,6 +2,8 @@ package org.dcache.commons.util;
 
 import org.slf4j.MDC;
 
+import java.util.Map;
+
 /**
  * The class emulates the Nested Diagnostic Context of Log4j.
  *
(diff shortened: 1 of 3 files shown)


paulmillar:
Al, looks like your pull request is broken; see:

    https://ci.dcache.org/job/dCache-v2.16/157/console

alrossi:
I fixed the package for the NDC on the other pull requests; evidently this one got neglected.

PR #2821, opened 2016-10-03
Active Transfers: substitute ? for <unknown> on html pages

alrossi:
Motivation:

DoorInfo and TransferInfo make the default values for owner, pool, protocol and process
"<unknown>".  While the Wicket engine automagically converts the angle brackets to
"&lt;" and "&gt;", the classic engine leaves the browser to interpret this marker
(erroneously) as an html tag.

See ticket 9056.

Modification:

Since the pool name is already handled by converting it to "N.N", and the process
by converting it to "?", we have added a similar conversion for protocol parts.

Further, we have made the webadmin pages conform to this representation as well.

Note that plain text and JSON output is not affected (only HTML).

Result:

No confusing "<unknown>" elements in the active transfer page markup.

Target: master
Request: 2.16
Request: 2.15
Request: 2.14
Request: 2.13
Ticket: http://rt.dcache.org/Ticket/Display.html?id=9056
Patch: https://rb.dcache.org/r/9785
Acked-by: Paul

### DIFF
--- a/modules/dcache-webadmin/src/main/java/org/dcache/webadmin/controller/util/BeanDataMapper.java
+++ b/modules/dcache-webadmin/src/main/java/org/dcache/webadmin/controller/util/BeanDataMapper.java
@@ -1,5 +1,7 @@
 package org.dcache.webadmin.controller.util;
 
+import com.google.common.base.Strings;
+
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


paulmillar:
Patch doesn't match RB entry; in particular, it introduces ETWS in TransferObserverV1.java on line 563.

alrossi:
The patch can't match RB because the Transfer Observer code is different from 2.16 (patches were not backported).   Aside from ETWS, what else are you objecting to precisely?

alrossi:
Oh, wait ... I see now.  This is 2.16.  This should be identical.

PR #2837, opened 2016-10-11
common: do not markup ellipses as a user-value

paulmillar:
Motivation:

Stylistically, the text help printer supports output that distinguishes
between user-input and structual elements in the output.  This currently
results in user-input text being underlined.

The ellipsis in an argument's representation (that denote accepting
multiple arguments) are marked up as user-input.  This developer finds
it easier to parse if ellipses are not marked up as a user-value; that
it is, instead, an non-marked-up structural element.

Modification:

No longer markup ellipses.

Result:

Easier to parse help text.

Target: master
Request: 2.16
Requires-notes: yes
Requires-srm-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/9825/
Acked-by: Olufemi Adeyemi

### DIFF
--- a/modules/common-cli/src/main/java/dmg/util/command/TextHelpPrinter.java
+++ b/modules/common-cli/src/main/java/dmg/util/command/TextHelpPrinter.java
@@ -198,7 +198,7 @@ public abstract class TextHelpPrinter implements AnnotatedCommandHelpPrinter
                 signature.append("[").append(metaVar).append("]");
             }
             if (field.getType().isArray()) {
-                signature.append(value("..."));
+                signature.append("...");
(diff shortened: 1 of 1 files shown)


gbehrmann:
Wow, you dislike it so much you actually have to back port a stylistic change?

paulmillar:
I see it more that I'm fixing a bug in the srmfs: I'm not happy with the help output and want to fix it.  This means back-porting patches like this.

gbehrmann:
Wait a couple of weeks and srmclient will be 3.0 (we always build from the latest stable branch, right?).

PR #2845, opened 2016-10-13
common: update ColumnWriter to remove any trailing whitespace

paulmillar:
Motivation:

Trailing whitespace can lead to ugly output as ColumnWriter ensures all
columns output the same width so if one column is too long for the
console (and is therefore wrapped), all columns are too long.

Modification:

Create a specialisation of PrintWriter that separates trailing
whitespace from other output.  The trailing whitespace (if any) is
cached.  Any text preceeding trailing whitespace (if any) triggers any
previously cached whitespace to be written before writing the
non-whitespace output.  Cached whitespace is discarded when a newline is
created.

Result:

ColumnWriter is updated so it never prints non-whitespace characters
before a newline.

Target: master
Requires-notes: yes
Requires-srm-notes: yes
Request: 2.16
Patch: https://rb.dcache.org/r/9826/
Acked-by: Gerd Behrmann

### DIFF
--- a/modules/common/src/main/java/org/dcache/util/ColumnWriter.java
+++ b/modules/common/src/main/java/org/dcache/util/ColumnWriter.java
@@ -102,7 +102,7 @@ public class ColumnWriter
         List<Integer> spaces = new ArrayList<>(this.spaces);
 
         StringWriter result = new StringWriter();
-        try (PrintWriter out = new PrintWriter(result)) {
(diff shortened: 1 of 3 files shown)


paulmillar:
retest this please

paulmillar:
retest this please

paulmillar:
The first "retest this please" build failed due to OpenJDK RPM being updated whilst building.

mksahakyan:
retest this please

PR #2926, opened 2016-11-09
srmclient: fix compatibility with Bestman

paulmillar:
Motivation:

A bestman instance can reject srmfs' initial stat (in reality, an srmLs)
request with a response that does not include any details
(ArrayOfTMetaDataPathDetail).  Such behaviour currently triggers a
NullPointerException in the client.

Modification:

Check the request-level ReturnStatus.StatusCode is successful before
attempting to extract the details and checking whether the individual
surl was successfully stat-ed.

Result:

The client correctly reports when a bestman endpoint does not accept a
user's SRM request.

Target: master
Requires-notes: no
Requires-srmclient-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/9890/
Acked-by: Gerd Behrmann
Request: 3.0
Request: 2.16


paulmillar:
retest this please

PR #2930, opened 2016-11-11
srmclient: fix srmfs 'get permission' command

paulmillar:
Motivation:

The srmfs 'get permission' command has several problems:
  1. the output includes eronous space in 'group' output,
  2. it fails to show 'owner' permissions when owner is unknown,
  3. it has potential to throw NullPointerException,
  4. it fails to show the 'other' permissions.

Modification:

Update code to fix these issues.

Result:

More detailed and accurate output that is more robust.

Target: master
Request: 3.0
Requires-notes: no
Requires-srmclient-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/9896/
Acked-by: Gerd Behrmann


jstarek:
retest this please

PR #2931, opened 2016-11-11
srmclient: update command names in srmfs

paulmillar:
Motivation:

The 'get' command clashes with commands that start 'get' (e.g., "get
space tokens") that can result in an ambiguity whether the user is
requesting a file be downloaded or whether the shell should run the
named command.

Modification:

Update commands that start 'get'.  Some descriptions are also adjusted.

Result:

The 'get' command becomes unambigious.

Target: master
Request: 3.0
Requires-notes: no
Requires-book: no
Requires-srmclient-notes: yes
Patch: https://rb.dcache.org/r/9895/
Acked-by: Gerd Behrmann


jstarek:
retest this please

PR #2935, opened 2016-11-14
srmclient: by default show only CN of owner in 'ls -l' output

paulmillar:
Motivation:

Some SRM storage systems provide a Distinguished Name (DN) in OpenSSL
format (e.g., "/C=DE/O=GermanGrid/OU=...").  This is cumbersome since in
the vast majority of cases, only the person's name is of interest which
(with high likelihood) identifies the person.  This is encoded in the
Common Name (CN) RDN, with the remaining text both taking up horizontal
space and being largely irrelevant.

Modification:

Add support for detecting file owners that are expressed as an
OpenSSL-formatted DN string.  By default, such strings are parsed and
only the CN value is printed.  Strings that do not match as an
OpenSSL-formatted DN string with a CN RDN are shown as-is.  An option is
provided to bypass this behaviour, always showing the owner as returned
by the server.

Result:

Output from 'ls -l' is more likely to fit within a typical terminal
dimensions.

Target: master
Request: 3.0
Requires-notes: no
Requires-srmclient-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/9902/
Acked-by: Albert Rossi

Conflicts:
	modules/srm-client/src/main/java/org/dcache/srm/shell/SrmShell.java

### DIFF
--- a/modules/srm-client/src/main/java/org/dcache/srm/shell/SrmShell.java
+++ b/modules/srm-client/src/main/java/org/dcache/srm/shell/SrmShell.java
@@ -17,6 +17,7 @@
  */
 package org.dcache.srm.shell;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
 import com.google.common.base.Predicate;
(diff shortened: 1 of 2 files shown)


jstarek:
retest this please

PR #2971, opened 2016-11-29
pool: Delete files concurrently

gbehrmann:
Motivation:

cleaner submits batches of files to delete to a pool, but a pool processed this
batch sequentially. On a pool with slow meta data operations (for whatever
reason), this can result if fairly low deletion rates and this can in turn slow
down deletion in the entire system.

Modification:

Submit delete to the pool's worker thread pool. Not only does this free one of
the message threads, it allows the delete tasks to be run concurrently. The
increase in concurrency can reduce the overhead of syncing the meta data to
disk after every deletion, thus increasing the deletion rate.

Result:

Improved the deletion rate of files being cleaned on a pool by running deletion
of multiple files on a pool concurrently from the pool's worker thread pool.

I am asking for backport as sites have complained about deletion throughput,
but will only ask for backport to 2.16 as it isn't really fixing a bug. Sites
should move ahead to 2.16 if they want to benefit from this improvement.

Target: master
Request: 3.0
Request: 2.16
Require-notes: yes
Require-book: no

Reviewed at https://rb.dcache.org/r/9924/

(cherry picked from commit ece9982a985d5b4eeb86cbea48896eb8ae45a157)

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/pool/classic/PoolV4.java
+++ b/modules/dcache/src/main/java/org/dcache/pool/classic/PoolV4.java
@@ -5,6 +5,9 @@ package org.dcache.pool.classic;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.net.InetAddresses;
+import com.google.common.util.concurrent.FutureCallback;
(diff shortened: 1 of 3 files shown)


paulmillar:
retest this please

PR #2976, opened 2016-12-05
srmclient: fix stat/ls caching for glob expansion for DPM

paulmillar:
Motivation:

Glob expansion is a potentially expensive operation and the current
implementation makes use of caching to reduce the number of srmLs
operations.

The work-around for DPM returning non-canonical absolute paths wasn't
included in this caching, resulting in needless cache-misses and
corresponding (unnecessary) additional srmLs operations.

This made glob-based directory listing slow.

Modification:

Ensure that glob based directory listing makes use of the DPM
work-around.

Result:

Much faster namespace operations on DPM when globbing is in use.

Target: master
Request: 3.0
Requires-notes: no
Requires-book: no
Requires-srmclient-notes: yes
Patch: https://rb.dcache.org/r/9954/
Acked-by: Gerd Behrmann

### DIFF
--- a/modules/srm-client/src/main/java/org/dcache/srm/shell/SrmShell.java
+++ b/modules/srm-client/src/main/java/org/dcache/srm/shell/SrmShell.java
@@ -216,7 +216,11 @@ public class SrmShell extends ShellApplication
     {
         File absPath = new File(metadata.getPath());
         try {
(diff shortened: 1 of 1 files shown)


jstarek:
retest this please

PR #2977, opened 2016-12-05
srmclient: percent-encode paths that require it

paulmillar:
Motivation:

srmfs does not work with paths that require percent encoding. This
includes any file containing a space or a hash symbol.

Modification:

The problem stems from Axis' URI class failing to do any escaping.
Therefore, this patch ensures paths are encoded before delivering them
to Axis.

A similar problem exists in the FTP transfer agent: the path was not
decoded, resulting in files being uploaded as their encoded filename
("test%20file" rather than "test file") subsequently failing the upload.

Use existing support in dCache for percent-encoding a path and switch
FTP handler to store the remote URI using Java's URI, which handles
encoding correctly.

Result:

Namespace operations and file transfers succeed with files containing
escapable characters.

Target: master
Request: 3.0
Requires-notes: no
Requires-book: no
Requires-srmclient-notes: yes
Patch: https://rb.dcache.org/r/9957/
Acked-by: Gerd Behrmann


jstarek:
retest this please

jstarek:
retest this please

PR #2990, opened 2016-12-12
srm-common: don't retry when JVM runs out of memory

paulmillar:
Motivation:

The generic SRMClient code invokes SRM stub methods using reflection.
This process will always retry the request if the exception isn't
ConnectionException.  This means that the client will retry an
operation, even if there is a critical Errors and or a bug
(RuntimeExceptions).

Modification:

Ensure bugs and JVM-critical errors do not trigger a retry of the
operation, but favour a fail-fast approach.

Result:

Slightly better user experience.

Target: master
Request: 3.0
Requires-notes: no
Requires-srmclient-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/9976/
Acked-by: Tigran Mkrtchyan


paulmillar:
retest this please

PR #3023, opened 2017-01-31
lm: Fix lost thread interrupt that prevents shutdown

gbehrmann:
Motivation:

When a socket connect is interrupted it throws ClosedByInterruptException.

Modification:

Allow this exception to terminate the connector thread such that it shuts
down if interrupted.

Result:

Fixed a problem in location manager in which it would fail to shut down
and could leave two overlapping connectors running.

Target: trunk
Request: 3.0
Request: 2.16
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/10013/
(cherry picked from commit 0c13fa2fa67708c9aef8418816e3539c4d330c2c)


gbehrmann:
retest this please

PR #3024, opened 2017-01-31
pool: Fix transactional behaviour with Berkeley DB

gbehrmann:
Motivation:

In version 3.0 we enabled transactional updates of the Berkeley DB meta data
database used in pools, however the code was flawed: The low level transactions
exposed by Berkeley DB are not bound to threads and a simple beginTransaction
call is not enough to make subsequent operations by this part of that transaction.

Modification:

Use the TransactionRunner helper class to encapsulate the update in a transaction.
This helper class associates the transaction with the calling thread.

Result:

Fixed a flaw in the pool in which transactional updates were not transactional.

Target: trunk
Request: 3.0
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/10028/
(cherry picked from commit 2469be06e6a2795a7c7a8ad64bfc52026c946094)


paulmillar:
retest this please

PR #3025, opened 2017-01-31
pool: Fix race and nested transaction problem during deletion

gbehrmann:
Motivation:

There are several problems that may occur when deleting a file: First
of all, the call to destroy happens from inside a transaction but
destroy itself issues an update call and thus triggers another
transaction. Second, actually deleting the physical file from
within the transaction is bad for two reason: The transaction may
fail and doing the part we cannot undo (deleting the actual file) from
within the transaction is obviously flawd; moreover, the conditions
we test before calling destroy are uncommitted ie another thread
may concurrently increase the link count as it doesn't 'see' the
entry as being REMOVED yet.

Modification:

Commit the transaction before checking the state of the entry and
calling destroy.

Result:

Fixed a race condition in pools that could cause repository corruption
while trying to delete a file.

Target: trunk
Require-notes: yes
Require-book: no
Request: 3.0
Acked-by: Jürgen Starek <juergen.starek@desy.de>
Patch: https://rb.dcache.org/r/10029/
(cherry picked from commit ecf5d2607faf2f1368bf1e04abcd057fca3532cf)

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/pool/repository/ReplicaStoreCache.java
+++ b/modules/dcache/src/main/java/org/dcache/pool/repository/ReplicaStoreCache.java
@@ -344,7 +344,7 @@ public class ReplicaStoreCache
         public synchronized <T> T update(Update<T> update) throws CacheException
         {
             try {
-                return _record.update(
+                T result = _record.update(
                         r -> update.apply(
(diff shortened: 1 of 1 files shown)


paulmillar:
retest this please

PR #3039, opened 2017-02-14
ftp: add support for paths relative to home directory

paulmillar:
Motivation:

Globus client makes requests using paths like '~/', with the tilde
symbol is a common place-holder for the users home directory.

Modification:

When resolving a client-supplied path, interpret paths that start '~/'
as the same path without this prefix and resolved relative to the users
home directory.  If the user has no home directory, this is resolved
relative to the root directory.

Note that '~NAME' is a common place-holder for the home directory of
user NAME.  Implementing support for this is deferred until a client
requires it.

Result:

Globus transfer service shows the user's home directory when first
logged it, rather than failing.

Target: master
Request: 3.0
Request: 2.16
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/10064/
Acked-by: Dmitry Litvintsev


jstarek:
retest this please

jstarek:
retest this please

PR #3047, opened 2017-02-15
restful: improve error handling

paulmillar:
Motivation:

When there is some kind of failure, the response is a JSON object
contains a list of error objects.  Each error object contains a status
code and a message.  However, the message is predetermined and contains
no contextual information about what went wrong.  This makes it hard for
someone using the dCache REST API to discover what went wrong when the
request is a 400 or 500 response.

If the Resource handling the request contains a bug, nothing is logged.
This is against dCache behaviour elsewhere where all bugs trigger a
stack-trace.

Modification:

Add a ExceptionMapper to control how exceptions are mapped to Response
objects.  Any non JAX-WS exception is treated as a bug: the Resource is
responsible for mapping all other exceptions (e.g., CacheException) to a
corresponding JAX-WS exception.

Result:

The error JSON object contains potentially useful information in the
message, as supplied by the exception.  Any bugs are logged with
stack-trace to facilitate debugging.

Target: master
Patch: https://rb.dcache.org/r/10054/
Acked-by: Olufemi Adeyemi
Requires-notes: yes
Requires-book: no
Request: 3.0

### DIFF
--- a/modules/dcache-restful-api/src/main/java/org/dcache/restful/DcacheRestApplication.java
+++ b/modules/dcache-restful-api/src/main/java/org/dcache/restful/DcacheRestApplication.java
@@ -8,6 +8,7 @@ import org.glassfish.jersey.server.ResourceConfig;
 import org.glassfish.jersey.server.filter.EncodingFilter;
 
 import org.dcache.restful.filters.ResponseHeaderFilter;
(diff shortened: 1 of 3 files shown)


paulmillar:
retest this please

paulmillar:
retest this please

PR #3076, opened 2017-02-17
srm: include remaining request lifetime in various responses

paulmillar:
Motivation:

During an ATLAS stress-test of tape recalls, it was discovered that
various sites had relatively short request lifetimes.  However, the SRM
spec provides the opportunity for the server to inform the client (FTS,
in this case) of what lifetime a request actually has.

This gives the client an opportunity to discover a discrepency between
desired lifetime and the lifetime of the job.

Modification:

Include the requests remaining lifetime in the response from the server.

Target: master
Requires-notes: yes
Requires-book: no
Request: 3.0
Request: 2.16
Request: 2.15
Request: 2.14
Request: 2.13
Patch: https://rb.dcache.org/r/10077/
Acked-by: Tigran Mkrtchyan


mksahakyan:
retest this please

PR #3109, opened 2017-03-07
ftp: do not fail proxy transfer if read returns zero bytes

paulmillar:
Motivation:

The Socket#read method may return zero to indicate that no bytes were
read.  Although this is not an error, such occurances will result
in a transfer failing.

Modification:

Continue reading from a socket if a read operation does not read any
bytes.

Result:

FTP proxied transfers are more reliable.

Target: master
Requires-notes: yes
Requires-book: no
Request: 3.0
Request: 2.16
Request: 2.15
Request: 2.14
Request: 2.13
Patch: https://rb.dcache.org/r/10098/
Acked-by: Dmitry Litvintsev


mksahakyan:
retest this please

PR #3134, opened 2017-03-23
nfs: recall file layout on pool disable

kofemann:
Motivation:
nfs door keeps a mapping between file's open state, unique id associated
with an open, and a transfer, associated with the mover on pool. Every time
when a client requests a mover for a given open-stateid, the same transfer
will be used.

Nevertheless, if pool gets disabled, restarted or unavailable, the mover
associated with the transfer is not valid any more. To create a new mover
on the same on a different pool, door have to 'forget' existing transfer
and notify nfs client that data server does not exists any more.

Modification:
NFS door reacts on pool-disable message and recall layouts. The read
transfers will become new movers. Writes are poisoned with IO error.

Result:
pool restart on read will resume the transfer when pool is back.
As a side effect, pool disable till redirect transfer to an other
pool, if replica exists. Pool disable in pool manager will trigger
p2p and transfer will continue.

All writed will fail with IO error.

Acked-by: Paul Millar
Target: master
Require-book: no
Require-notes: yes
(cherry picked from commit ee816235190616bf157f7a9359b3bd4233c22518)


mksahakyan:
retest this please

PR #3146, opened 2017-04-14
parent/dcache: add mail jar to deployment

alrossi:
Motivation:

The alarms service needs a javax.mail implementation
jar to be present in order for it to send email alerts.

This jar was present in 2.13, but somehow disappeared
from the pom of 2.16+.

Modification:

Add the jar to the parent pom.xml and to the dcache pom.xml.

Result:

Alarms is happy when you configure it to send emails.

Target: master
Request: 3.1
Request: 3.0
Request: 2.16
Bug: 3135
Acked-by: Tigran


mksahakyan:
retest this please

PR #3159, opened 2017-04-28
xrootd: writing to an existing file is kXR_NotAuthorized, not kXR_Unsupported

fizmat:
Motivation:
kXR_Unsupported is interpreted by xrootd as ENOSYS, which gets ignored by FUSE and leads to errors later that are hard to diagnose.
Attempting to copy a file to dcache mounted over xrootd leads to as segfault in xrootd filesystem code because the (second) open operation fails, but the error is ENOSYS so FUSE does not stop and tries to use the null file descriptor for writing. 

Modification:
Respond kXR_NotAuthorized instead

Result:
Attempting to copy a file to dcache mounted over xrootd by fuse leads to a "permission denied" error and a file in a 'new' state in dcache. This situation needs fixing, but it is better, as the exact point of failure is easier to locate.
`cp: cannot create regular file 'mnt/public/123': Permission denied`


paulmillar:
ok to test

kofemann:
https://github.com/xrootd/xrootd/blob/52032c50f9d7dc77354a836741d3ae46dbf1fb14/src/XProtocol/XProtocol.hh#L852

Patch looks  correct

PR #3165, opened 2017-05-03
xrootd: allow xrootd to overwrite new empty files.

fizmat:
Motivation:
Xrootdfs FUSE driver immediately closes a file on creation, then reopens it to write.
The file cannot be opened as it already exists. This patch allows the dcache xrootd door to tolerate this behavior.

Modification:
If a file is newly created, allow overwriting it even if the request flags don't ask for it.

Specifically, unless XrdCl::OpenFlags::Delete is set overwriting is disallowed.
But XrootdDoor now checks if the file is new by querying transfer.isCreatedOnly(), and in that case overwrites the file anyway.

Result:
xrootdfs FUSE driver can be used to copy a file onto a mounted dCache filesystem.
This process is inefficient, because the file is created twice, but getting rid of the double open requires changes in the xrootd driver.

### DIFF
--- a/modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/StorageInfo.java
+++ b/modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/StorageInfo.java
@@ -110,8 +110,7 @@ public interface StorageInfo
     void setHsm(String newHsm);
     /**
       * Determines whether the file exists somewhere (cache or HSM)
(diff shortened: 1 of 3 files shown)


paulmillar:
ok to test

jstarek:
test this please

jstarek:
retest this please

kofemann:
ok to test

jstarek:
retest this please

PR #3184, opened 2017-05-12
nfs: bind vfs cache invalidation with file's layout

kofemann:
Motivation:
On write we have to invalidate vfs cache, as file's attributes are
changed. For now, we do this only for files that had a transfer. However
this have to be done for IO on dot files as well.

Modification:
Bind vfs cache invalidation with layout state id disposal. This will
happen on close for regular and dot files.

Result:
no stale information for dot files after update.

Acked-by: Paul Millar
Target: master, 3.1
Require-book: no
Require-notes: no
(cherry picked from commit 3961e557c61347f7d3d31b54521200b7322dc433)


paulmillar:
retest this please

PR #3185, opened 2017-05-15
xrootd: better error responses to invalid open requests

fizmat:
Motivation:
When opening file with no write flag but a delete or create operation, the response text was generic, while the response codes are different.
kXR_Unsuported does not fit, as it is not an operation that is expected to work on any server.

Change:
Respond to both with kXR_ArgInvalid with specific error text.

Result:
No change shoud be noticeable with any sane client, but debugging errors a bad xrootd client code should become easier.


jstarek:
OK to test

PR #3186, opened 2017-05-16
[WIP] xrootd: refactor read-write logic for pool open requests

fizmat:
Motivation:
The reason why pool responds to similar requests with different error code is not obvious.

Change:
Rewrite the if tree. Separate the file state and the request flags to see the possible cases more clearly.
Add comments, especially for non-obvious cases.
Maybe some of the non-obvious cases are wrong and should be changed within the same pull request?

Result:
The logic is equivalent so the result should be exactly the same in all cases.

### DIFF
--- a/modules/dcache-xrootd/src/main/java/org/dcache/xrootd/pool/XrootdPoolRequestHandler.java
+++ b/modules/dcache-xrootd/src/main/java/org/dcache/xrootd/pool/XrootdPoolRequestHandler.java
@@ -250,15 +250,41 @@ public class XrootdPoolRequestHandler extends AbstractXrootdRequestHandler
             try {
                 FileDescriptor descriptor;
                 IoMode mode = file.getIoMode();
-                if (msg.isNew() && mode != IoMode.WRITE) {
-                    throw new XrootdException(kXR_ArgInvalid, "File exists.");
-                } else if (msg.isDelete() && mode != IoMode.WRITE) {
-                    throw new XrootdException(kXR_Unsupported, "File exists.");
-                } else if ((msg.isNew() || msg.isReadWrite()) && mode == IoMode.WRITE) {
-                    descriptor = new WriteDescriptor(file, (msg.getOptions() & kXR_posc) == kXR_posc ||
-                            file.getProtocolInfo().getFlags().contains(XrootdProtocolInfo.Flags.POSC));
+                if (mode == IoMode.WRITE) {
+                    // File is new, can write.
+                    if (msg.isNew() || msg.isReadWrite()) {
+                        // Client wants to write, can write, threfore write.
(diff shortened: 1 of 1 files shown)


modules/dcache-xrootd/src/main/java/org/dcache/xrootd/pool/XrootdPoolRequestHandler.java
```diff
@@ -250,15 +250,41 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable t)
...
+                    if (msg.isNew() || msg.isReadWrite()) {
+                        // Client wants to write, can write, threfore write.
+                        descriptor = new WriteDescriptor(file, (msg.getOptions() & kXR_posc) == kXR_posc ||
+                                file.getProtocolInfo().getFlags().contains(XrootdProtocolInfo.Flags.POSC));
+                    } else if (msg.isDelete()) {
+                        // Client wants to "open a new file, deleting any existing file".
+                        // Is it possible to have isDelete without isReadWrite?
+                        // If it is possible, is this response correct?
+                        descriptor = new ReadDescriptor(file);
+                    } else {
```
paulmillar: Looks wrong: both sides of if-else-statement are the same

modules/dcache-xrootd/src/main/java/org/dcache/xrootd/pool/XrootdPoolRequestHandler.java
```diff
@@ -250,15 +250,41 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable t)
...
+                    // File exists, can only read.
+                    if (msg.isNew()) {
+                        // Client wants to create file only if it does not exist.
+                        throw new XrootdException(kXR_ArgInvalid, "File exists.");
+                    } else if (msg.isDelete()) {
+                        // Client wants to "open a new file, deleting any existing file".
+                        // We don't support that on the pool level.
+                        // Should we redirect to the door instead?
+                        throw new XrootdException(kXR_Unsupported, "File exists.");
+                    } else if (msg.isReadWrite()){
```
paulmillar: As above, suggest removing the `if (msg.isReadWrite)` statement since the code does the same thing in either branch.  The comments should be updated accordingly.

jstarek:
Moin, are there any updates on this?

jstarek:
@fizmat , any news on this?

fizmat:
Sorry for not responding sooner.

Different branches with the same action are intentional, to make non-obvious cases easier to see and comment, while keeping the logic the same as the original code. I intended to change what gets done in different branches if we come to the conclusion that it's incorrect.
However, Tigran and I could not settle which xrootd flag combinations are possible/reasonable so this idea stalled.

This PR can be closed for now, it's unlikely to be finished soon. If I find out more and come up with useful changes, I will create a new request.

jstarek:
Closing due to inactivity.

PR #3198, opened 2017-05-24
httpd: Fix incomplete restore list in HA setup

gbehrmann:
Motivation:

In an HA setup with multiple pool managers, the httpd (old) service failed to
fetch the list of restore requests from all instances. Instead the service
would query one of the instances, possibly alternating depending on the
grouping of services in domains.

Modification:

The gist of the fix is to let the responsible collector in the httpd service
use a PoolManagerHandler to query the list of restores and then extend the
implementations of PoolManagerHandler to recognize that query and fetch the
list from all PoolManager instances.

Since PoolManagerHandler is limited to PoolManagerMessage subclasses, a new
request object is introduced. To maintain backwards compatibility, the
PoolManagerHandler implementations translate this request to the classic `xrc
ls` command, but that translation can be reduced once backwards compatibility
is no longer required.

To handle the startup and shutdown of the PoolManagerHandlerSubscriber
correctly, the httpd engine interface had to be adjusted to remove the custom
startup and shutdown methods and instead rely on regular cell lifecycle
callbacks.

Target:

Fixed a problem in which the restore list in the old httpd interface did not
show all restored when used in an HA setup.

Target: trunk
Require-notes: yes
Require-book: no
Request: 3.1
Request: 3.0
Acked-by: Tigran Mkrtchya <tigran.mkrtchyan@desy.de>

Reviewed at https://rb.dcache.org/r/10192/

(cherry picked from commit d7db2599227895891d61ba7044cb7e8298e4593d)

### DIFF
--- a/modules/cells/src/main/java/dmg/cells/nucleus/CellNucleus.java
+++ b/modules/cells/src/main/java/dmg/cells/nucleus/CellNucleus.java
@@ -516,7 +516,7 @@ public class CellNucleus implements ThreadFactory
                             long timeout)
         throws SerializationException
     {
-        checkState(_state.isSendWithCallbackAllowed);
+        checkState(_state.isSendWithCallbackAllowed, "Cannot send message with callback in state {}", _state);
         checkArgument(!msg.isFinalDestination(), "Message has no next destination: %s", msg.getDestinationPath());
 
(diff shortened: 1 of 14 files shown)


jstarek:
retest this please

PR #3200, opened 2017-05-31
webdav: Fix restriction check when downloading a file

paulmillar:
Motivation:

The Restriction allows for namespace limitations on what a user can do.
By omitting to send the file's path to PnfsManager, the WebDAV door
effectively by-passed the restrictions check when downloading a file.

Modification:

Fix Transfer class to include the path when querying file attributes,
even if the PnfsId is already known.

Update PnfsManager to log if a door has supplied insufficient
information for the Restriction check.  This should hopefully make it
easier to find any similar bugs.

Result:

Restrictions are enforced when downloading files using the WebDAV door.

Target: master
Request: 3.1
Request: 3.0
Request: 2.16
Request: 2.15
Request: 2.14
Request: 2.13
Patch: https://rb.dcache.org/r/10211/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache/src/main/java/diskCacheV111/namespace/PnfsManagerV3.java
+++ b/modules/dcache/src/main/java/diskCacheV111/namespace/PnfsManagerV3.java
@@ -2338,6 +2338,8 @@ public class PnfsManagerV3
             if (path != null) {
                 checkRestriction(message.getRestriction(), message.getAccessMask(),
                         activity, path);
+            } else {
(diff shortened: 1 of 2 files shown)


jstarek:
You have Request: statements for 2.13 and 2.14, but did not submit the pull requests yet -- do you still want to send them?

paulmillar:
It turns out that Restriction support was introduced with dCache v2.15, so there's nothing to fix in v2.14 or earlier.

PR #3209, opened 2017-06-09
webdav: make Milton work-around more robust

paulmillar:
Motivation:

Commit 5abc0e1c introduced a work-around against Milton's bad behaviour
if there is an IOException during an upload.  Unfortunately, that
work-around wasn't broad enough to cover all failure modes.  The result
is that bad client behaviour can trigger a stack-trace in dCache.

Additionally, the work-around returned a 400 error code.  Conforming
clients will not retry if they see this error, which is undesireable
behaviour since an IOException may be a transitory problem.

Modification:

Update work-around to cover any IOException that is thrown.

Return 500 Internal Error instead of 400 Bad Request.

Result:

Bad client or network behaviour should no longer trigger a stack-trace.

Clients should be free to retry the failed upload.

Target: master
Patch: https://rb.dcache.org/r/10223
Acked-by: Tigran Mkrtchyan
Requires-notes: yes
Requires-book: no
Request: 3.1
Request: 3.0
Request: 2.16
Request: 2.15
Request: 2.14
Request: 2.13

### DIFF
--- a/modules/dcache-webdav/src/main/java/org/dcache/webdav/DcacheDirectoryResource.java
+++ b/modules/dcache-webdav/src/main/java/org/dcache/webdav/DcacheDirectoryResource.java
@@ -88,8 +88,7 @@ public class DcacheDirectoryResource
     @Override
     public Resource createNew(String newName, InputStream inputStream,
                               Long length, String contentType)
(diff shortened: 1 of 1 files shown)


paulmillar:
retest this please

Issue #3231, opened 2017-06-15
A frontend.dcache-view.org-name value with quote marks fail


paulmillar:
From messages sent on user-forum, on 2017-06-14:

From Femi:
> Although, I think the cause of the problem is in the layout file, try this:
>       change
>            frontend.dcache-view.org-name="SURFsara dCache-View"
>    to
>      frontend.dcache-view.org-name=SURFsara dCache-View
> that is, set the value without the quotation marks.

From Onno:
> Thanks, removing the quotes helped!

paulmillar:
Patch: https://rb.dcache.org/r/10282/

PR #3235, opened 2017-06-15
Comment frontend.dcache-view.org-name: no quotation marks

onnozweers:


paulmillar:
ok to test

femiadeyemi:
Thanks @onnozweers! 
@paulmillar I think this should fix https://github.com/dCache/dcache/issues/3231
If @paulmillar agree, @jstarek may you please merge and backport up to 2.16, thanks.

paulmillar:
@femiadeyemi this patch should be back-ported to all branches back to (and including) 2.16, since it accurately describes the current situation: the value cannot contain quote marks.

#3231 is a bug, something that simply adding documentation doesn't fix 😉 Instead, to fix this bug, dCache needs to be updated so that including double-quotes in the value is perfectly fine.

Ideally, what should have happened is that the admin included the double-quotes, sees a working dCache-View that has a organisation name that also has these double-quotes and realises that this isn't what is wanted, so removes them.

paulmillar:
Hi @onnozweers,

Thanks again for making this pull-request.

As it happens, the problem was relatively easy to fix.  The limitation is gone with dCache releases v2.16.40, v3.0.21 and v3.1.9.  With these versions of dCache, quote-marks are perfectly acceptable in the org name.

Therefore, I'm closing this pull-request.

Issue #3310, opened 2017-07-09
resilience: improve documentaion of options


calestyo:
Hi.

I was having a look at the .properties documentation of 2.16, and some things with respect to the resilience service remained pretty unclear to me, even after having again a look at the slides from Umeå.

First, in pnfsmanager.properties, there is this snipped for pnfsmanager.db.connections.max:
```
#  NOTE:  when running resilience embedded here, this number should
#         be increased. The recommended minimum setting would be
#
#               pnfsmanager.resilience.submit-threads
#             + pnfsmanager.resilience.pnfs-op-threads
#             + pnfsmanager.resilience.db.connections.max
#             + whatever maximum allowed for normal namespace settings
#               (default = 30)
#
#       Submit and pnfs threads require 1 database connection, and scan
#       threads need 2.
#
#       Be sure to adjust postgresql.conf max connections to allow
#       for the larger value here, plus the added pool scan
#       connections specified by pnfsmanager.resilience.db.connections.max.
#
```
What is actually meant here? What is "resilience embedded"? Does it mean the resilience *service*? And in which sense embedded? Running resilience in some special embedded mode (couldn't find anything about that) or simply having that service within the cluster?

Also the mentioned pnfsmanager.resilience.* do not seem to exist anywhere. And "whatever maximum allowed for normal namespace settings" should probably be clarified as well... e.g. is it just pnfsmanager.db.connections.max or also what NFS doors open to chimera? etc. pp...
Also, how does it relate to resilience.db.connections.max ?

Then there are these:
```
# ---- File-system-related properties.  These mirror the normal
#      namespace service setup.
#
resilience.plugins.storage-info-extractor=org.dcache.chimera.namespace.ChimeraOsmStorageInfoExtractor
(one-of?ONLINE|NEARLINE)resilience.default-access-latency=NEARLINE
(one-of?CUSTODIAL|REPLICA|OUTPUT)resilience.default-retention-policy=CUSTODIAL
resilience.enable.inherit-file-ownership = false
resilience.enable.full-path-permission-check=true
resilience.enable.acl = false
```
And to me it's really unclear what they do.
First, if they *mirror* the normal settings, why not taking them, i.e.:
`resilience.enable.acl = ${pnfsmanager.enable.acl}`
and so on.

Is it strictly necessary to keep them in sync with their pnfsmanager.* counterparts? What happens if one doesn't?

E.g. if resilience.enable.acl = false, does that mean that files copied by resilience won't get ACLs?
[...]

PR #3321, opened 2017-07-11
resilience: make namespace provider properties immutable

alrossi:
See https://github.com/dCache/dcache/issues/3310

Motivation:

Resilience uses its own namespace provider, but its interactions
with it are read-only.  While the provider needs to be fully
configured, some properties need not be exposed to the user.

Modifiation:

Remove properties that do not need to be exposed.
Inject defaults directly.
Mark removed properties as obsolete.

Result:  properties not needing to be exposed to configuration
are not.

Target: master
Request: 3.2
Request: 3.1
Request: 2.16
Acked-by: Paul
Closes: #3310

### DIFF
--- a/modules/dcache-resilience/src/main/resources/org/dcache/resilience/resilience.xml
+++ b/modules/dcache-resilience/src/main/resources/org/dcache/resilience/resilience.xml
@@ -54,8 +54,10 @@
 
     <bean id="extractor" class="${resilience.plugins.storage-info-extractor}">
       <description>Storage info extractor</description>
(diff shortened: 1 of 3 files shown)


jstarek:
Hi Al, could you take a look at the conflict in resilience.properties? Looks like it's just a single duplicated line. Thanks!

paulmillar:
Really?  Have a look at `resilience.properties`

alrossi:
I don't know what the heck is going on here.  I just fixed this damned thing.


________________________________________________
Albert L. Rossi
Application Developer & Systems Analyst III
Scientific Computing Division, Data Movement Development
FCC 229A
Mail Station 369 (FCC 2W)
Fermi National Accelerator Laboratory
Batavia, IL 60510
(630) 840-3023


________________________________

PR #3330, opened 2017-07-13
Fix/3.0/move to dir property

alrossi:
This patch depends on https://github.com/dCache/dcache/pull/3322 being merged first.


alrossi:
Will repost after 3322 is committed

PR #3363, opened 2017-07-27
webdav: fix more regressions with CredentialSource.NONE

paulmillar:
Motivation:

A previous patch (committed as 7e5e81e1) attempted to fix the regression
introduced by commit d132d860 where the webdav service no longer
accepted the "Credential: none" header for HTTPS-based third-party
transfers.  Unfortunately that fix was incomplete and additional changes
proved necessary, which this patch provides.

Modification:

Relax conditions that prevented HTTPS-based third-party transfers
without delegation.

Result:

A regression is fixed that prevented third-party transfers without a
delegated credential (e.g., using HTTP basic authentication or a bearer
token like macaroons).

Target: master
Request: 3.1
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/10386/
Acked-by: Tigran Mkrtchyan


paulmillar:
Merge request was originally with the wrong patch.  This is now fixed.

PR #3375, opened 2017-07-31
chimera: use jsr-107 jcache interface for caching

kofemann:
TEST pull request

Motivation:
to reduce load on DB, chimera caches inumber <-> pnfsid mapping.
This there is an assumption that this mapping will never change.
As some functionality in dCache may break this rule, we need a
way to invalidate cached values.

In case of multiple direct chimera instances, like PnfsManager in
HA setup or nfs door, the cache invalidation must be propagated to
all instances.

JSR-107 defines interface only API for caching. Depending on
implementation, cache can be local or distributed.

Modification:
Use JCache based distributed cache to load data into Guava's cache. This
construct still uses local cache, but loads data from distributed cache
if available. additionally, cache invalidation events handled by all cache
clients.

The JdbcFs updated to join or create a distributed Hazelcast cache.
A zookeeper based autodiscovery is added to cluster members to discover
each other.

The Guava's cache size is reduced to ensure, that values are always in the
shared cache, when they are in the local caches.

Result:
No visible user changes. However all direct chimera clients build a distributed
cache with autodiscovery.

Acked-by:
Target: master
Require-book: no
Require-notes: yes

### DIFF
--- a/modules/chimera/pom.xml
+++ b/modules/chimera/pom.xml
@@ -86,5 +86,10 @@
         <artifactId>h2</artifactId>
         <scope>test</scope>
     </dependency>
+    <dependency>
+        <groupId>javax.cache</groupId>
+        <artifactId>cache-api</artifactId>
+    </dependency>
+
   </dependencies>
 </project>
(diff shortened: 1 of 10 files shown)


mksahakyan:
retest this please

mksahakyan:
retest this please

PR #3384, opened 2017-08-02
billing: use cellName.cell instead of cellName

calestyo:
Motivation:
In 2.16 the field “cellName” for billing format strings got extended by
sub-fields, amongst others “cell”.
The actual output of “cellName.cell” seems to be identical to just specifying
“cellName” and in some but not all of the default billing format strings this
has been adapted already.

Modification:

Consistently use “cellName.cell” instead of just “cellName” in all billing
format strings.

Result:

Should have no effect on acutal output, but result in a more consistent
configuration.

Target: master
Require-notes: no
Require-book: no

### DIFF
--- a/skel/share/defaults/billing.properties
+++ b/skel/share/defaults/billing.properties
@@ -298,19 +298,19 @@ billing.text.format.mover-info-message = $date$ [$cellType$:$cellName.cell$:$typ
 #
 #    Submitted by PnfsManager on file removal.
 #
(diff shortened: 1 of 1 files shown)


paulmillar:
First, `$cellName$` and `$cellName.cell$` are not identical if the address is qualified.  This is described in the billing.properties file.

Second, the patch changes `billing.parser.format` prefix values.  This is clearly wrong as these properties describe the format of billing entries used before billing files became self-documenting.  The default values capture the defaults dCache used; i.e., if the admin didn't change how billing files are logged then everything works out-of-the-box.  These changes would break that compatibility.

calestyo:
ah I see... sorry...

PR #3390, opened 2017-08-03
system-test: explicit specify nfs domain

hamletson:
Motivation:
on some platforms (osx) the revers lookup of the IP may
produce results that will force dcache-startup to fail

Modification:
define nfs.domain property in system-test.layout

Result:
dcache start on osx with weird network config

Target: master
Require-book: no
Require-notes: no


kofemann:
ok to test

PR #3404, opened 2017-08-04
systemd: start service after ZooKeeper

calestyo:
Motivation:

Every dCache domain connects to ZooKeeper.
If a ZooKeeper service is running on a host where dCache domains are running, it
should in practise be pretty likely that this is a ZooKeeper server from the
ensmble used by dCache.
To prevent log messages it would be beneficial if dCache starts after a local
ZooKeeper service (if any).

Modification:

Added zookeeper.service to systemd’s “After” option.

Result:

dCache domains will start after a local ZooKeeper service (if any).

Target: master
Require-notes: no
Require-book: no


kofemann:
makes sense. Thanks!

calestyo:
:-)

btw: Is it planned to get the systemd support backported to 3.1? :-)

kofemann:
No.

PR #3410, opened 2017-08-04
FTPServerFacade: refactoring slf4j logging messages

lalottacontinua:

Result: gained efficiency

### DIFF
--- a/modules/ftp-client/readme.md
+++ b/modules/ftp-client/readme.md
@@ -1,0 +1,3 @@
+Modul ftp-client
+
+author: Lotta Rüger
--- a/modules/ftp-client/readme.md
+++ b/modules/ftp-client/readme.md
@@ -1,3 +1,3 @@
-Modul ftp-client
+Module: ftp-client
 
 author: Lotta Rüger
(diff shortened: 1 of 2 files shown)


lalottacontinua:
Would I do a new pull request for that (changing the language of "Modul" to english) or do I just make a new commit? And is the pull request already merged with the master?

alrossi:
You can just modify and commit.  The pull request should be automatically updated.  It hasn't been merged yet.

alrossi:
Lotta, did you modify the same branch used to make the original pull request?  It seems this commit created a conflict.

lalottacontinua:
sry. I'm a little confused. I thought, I had already committed the readme.md in an earlier pull-request. and I thought, I had been doing the change on the same branch, but maybe I did the checkout after I edited the file. I'll try to further fix this...

alrossi:
The rebase fixed it.  Thanks!

lalottacontinua:
Thank you! 
I'm going to do more of this kind of refactoring. I'll use different branches for this, as it'll be in different classes/modules. Is that correct/ should I do so? And do you think I should test the refactoring somehow or are the automatically tests after merge sufficient in this case?

alrossi:
OK, I see what happened here, I think.   We need to return to base one.

Could you possibly do the following:

1.  update your current branch of master first, to make sure that your ReadMe commit is there.
2. make a new pull request with all the slf4j changes you intend to do.  If that is all the patch does, it can embrace different classes.  You don't need a different request for each one.  But make sure you don't do anything else except change the log statements in the patch (the purpose of the patch).
3. when you are done, create a new pull request.

We will close this one (i.e., discard it).

Thanks, Al

alrossi:
P.S. Don't worry about the German in the ReadMe.

PR #3418, opened 2017-08-04
chimerashell: use chimera.db.* options as defaults

calestyo:
Motivation:

chimerashell accesses the Chimera DB.
With respect to the DB config settings, there are basically two scenarios:
The Chimera DB is on the DB server as specified by the global defaults
(dcache.db.*) or on another one and then the chimera “service” specific options
(chimera.db.*) are set accordingly.

Currently, in the later case, chimerashell will not find the right DB server, as
the dcache.db.* options are used (which may still point to the DB server, where
all non-Chimera-DBs run).

Modification:

For the chimerashell.db.* options, use their respective chimera.db.*
counterparts and no longer a static name for chimerashell.db.name.

Result:

In any case, the chimerashell should now get the right connection settings.

Target: master
Request: 3.1
Require-notes: yes

### DIFF
--- a/skel/share/defaults/chimerashell.properties
+++ b/skel/share/defaults/chimerashell.properties
@@ -6,27 +6,27 @@
 
 #  ---- Chimera database name
 #
-chimerashell.db.name = chimera
+chimerashell.db.name = ${chimera.db.name}
 
 #  ---- Chimera database host name
 #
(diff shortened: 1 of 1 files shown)


kofemann:
I am a victim of this as well, but was lazy to fix that. Thanks.

PR #3419, opened 2017-08-06
acl-vehicles: refactoring slf4j logging messages

lalottacontinua:

Result: gained efficiency


kofemann:
Thanks!

kofemann:
ok to test

PR #3420, opened 2017-08-06
acl: refactoring slf4j logging messages

marisanest:

Result: gained efficiency

### DIFF
--- a/modules/acl/src/main/java/org/dcache/acl/mapper/AclMapper.java
+++ b/modules/acl/src/main/java/org/dcache/acl/mapper/AclMapper.java
@@ -31,10 +31,12 @@ public class AclMapper {
     }
 
     public static Permission getPermission(Subject subject, Origin origin, Owner owner, ACL acl) {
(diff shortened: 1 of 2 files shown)


kofemann:
ok to test

modules/acl/src/main/java/org/dcache/acl/mapper/AclMapper.java
```diff
@@ -78,10 +75,9 @@ public static Permission getPermission(Subject subject, Origin origin, Owner own
         } catch (ACLException e) {
             logger.error(e.getMessage());
         } finally {
-            if ( logger.isDebugEnabled() ) {
```
kofemann: I guess, the is statement was there to prevent 'expensive' **permACL.asString(rsType)**

kofemann:
Thanks!

alrossi:
ok to test

kofemann:
Thanks!

PR #3421, opened 2017-08-06
chimera: refactoring slf4j logging messages and logger variable name

marisanest:

### DIFF
--- a/modules/chimera/src/main/java/org/dcache/chimera/DirectoryStreamHelper.java
+++ b/modules/chimera/src/main/java/org/dcache/chimera/DirectoryStreamHelper.java
@@ -25,7 +25,7 @@ import java.util.List;
 
 public class DirectoryStreamHelper {
 
-    private static final Logger _log = LoggerFactory.getLogger(DirectoryStreamHelper.class);
+    private static final Logger logger = LoggerFactory.getLogger(DirectoryStreamHelper.class);
 
     /**
      * Convert directory stream into a {@link List}.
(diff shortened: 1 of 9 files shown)


kofemann:
ok to test

modules/chimera/src/main/java/org/dcache/chimera/posix/UnixPermissionHandler.java
```diff
@@ -184,7 +184,7 @@ public boolean isAllowed(Acl acl, User user, int requsetedAcl) {
 
         } // switch( userUid )
 
-        _log.debug("IsAllowed: " + isAllowed);
+        logger.debug("IsAllowed: {}", isAllowed);
```
kofemann: @marisanest I suggest you close this pull request and create a new one, which includes this change only.

modules/chimera/src/main/java/org/dcache/chimera/DirectoryStreamHelper.java
```diff
@@ -25,7 +25,7 @@
 
 public class DirectoryStreamHelper {
 
-    private static final Logger _log = LoggerFactory.getLogger(DirectoryStreamHelper.class);
+    private static final Logger logger = LoggerFactory.getLogger(DirectoryStreamHelper.class);
```
marisanest: If we fix the variable naming should we use "LOG" or "LOGGER"?

kofemann: If you change it, then yes. Please use LOGGER. But please make variable name only commit.

PR #3423, opened 2017-08-06
cells: refactoring slf4j logging messages and logger variable name

lalottacontinua:


lalottacontinua:
Dear person who is going to review this pull request. There has been an accident with the last commit, where files were added, which belong to the git.ignore part, I think (when trying to change the name of the branch with the last push). At least those are targets file and it'd be great, if you could ignore them, if you'd merge this pull request. Furthermore this pull request has become quite big, as it wasn't clear to us, that changes on the same branch couldn't get split into different pull requests - or we at least didn't manage to do this. Sry, for the workload thus and thanks for your support. :)

kofemann:
ok to test

kofemann:
Hi Lotta,

Thanks for your pool request. You can fix your commit and send a new pool request.
Alternatively, we can do it manually and then close this pool request without merging 
in github

lalottacontinua:
ok. Marissa and me can try to do this. But never used revert or any other going-back git command. if you can suggest a procedure? Elsewise, we'll try to find out. And we should just make the last commit and push once more, the other two are fine in this pull request, or? Thanks for the feedback!

kofemann:
This the safes way to do it:
  
- checkout you branch with changes
- checkout a new branch, like git checkout -b my-branch-v2
- do required changes
- commit
- push
- create a new pull request from the new branch and close this one

if you more brave, then you can work on your original branch.
let say you have two commits which you want to re-work:
``` git reset HEAD~2```
This will rollback last two commits, but keep changes in the workspace.
Then you can do **git add** on desired changes and commit them.
push your branch with --force flag.

If you not sure, use first method. I can squash changes by prior merging.

**Comment by lalottacontinua** [comment]
I think originally there were some files which weren't supposed to become synchronized with your repository.. or they were just untracked from the beginning. and all of the sudden, they were in the commit.. 
it doesn't really feel right to delete: dcache/archetypes/dcache-nearline-plugin-archetype/target/classes/archetype-resources/pom.xml , as it is a pom.xml. but maybe it is ok, because it's inside the target folder? and I can just delete anything inside the target folder before doing a new commit?

**Comment by kofemann** [comment]
I think it's IdeaJ have added them. I am sure it's safe to delete them.

File: modules/cells/src/main/java/dmg/cells/network/LocationMgrTunnel.java [hunk 1]
```diff
@@ -66,7 +66,7 @@
      */
     private static final Tunnels _tunnels = new Tunnels();
 
-    private static final Logger _log =
+    private static final Logger logger =
```
kofemann: Please do not combine naming and logic changes into a single commit

modules/cells/src/main/java/dmg/cells/nucleus/CellAdapter.java
```diff
@@ -75,7 +75,7 @@
 public class CellAdapter
     implements Cell, CellEventListener, CellEndpoint
 {
-    private static final Logger _log =
+    private static final Logger logger =
```
kofemann: If you want to fix naming issues, then please use UPPER case variable names for constants, like **static final**

lalottacontinua:
ok. and I should not just do another commit in this branch, because this pull request is already getting tested, yes? :)

modules/cells/src/main/java/dmg/cells/nucleus/CellAdapter.java
```diff
@@ -758,15 +758,15 @@ public void prepareRemoval(KillEvent killEvent)
     @Override
     public void postRemoval(KillEvent ce)
     {
-        _log.info("CellAdapter : prepareRemoval : waiting for gate to open");
+        logger.info("CellAdapter : prepareRemoval : waiting for gate to open");
```
kofemann: you can see here: you change is unnecessary big just because of naming changes.

kofemann:
I hope your build system will survive this :-D

modules/cells/src/main/java/dmg/cells/nucleus/CellAdapter.java
lalottacontinua: oh, but we have been researching. Loggers are not constants according to a certain java styleguide. I'll research the link..

lalottacontinua: https://stackoverflow.com/questions/1417190/should-a-static-final-logger-be-declared-in-upper-case (3rd comment) shall we change it to uppercase anyway?

modules/cells/src/main/java/dmg/cells/nucleus/CellAdapter.java
lalottacontinua: true. We just thought it might be more efficient to do it both at once. shall we do it again/ seperate the two changes?

modules/cells/src/main/java/dmg/cells/network/LocationMgrTunnel.java
lalottacontinua: ok. thanks for the feedback!

modules/cells/src/main/java/dmg/cells/nucleus/CellAdapter.java
kofemann: That my be the case. As you see in our code, we have all  kind of  weird formatting. However, if you want to convince us, then this should be a different commit (in an ideal world, of corse).

lalottacontinua: ok. we'll see what we can do! :)

modules/cells/src/main/java/dmg/cells/nucleus/CellAdapter.java
kofemann: "Jeder Jeck ist anders". The main goal of the style is what we use as a team. Currently, we prefer LOGGER vs logger. I am ok with other places, where you change **_log** or **log** to logger.

kofemann:
You can use this pull request. We will re-test it.

archetypes/dcache-nearline-plugin-archetype/target/classes/META-INF/maven/archetype-metadata.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kofemann: this file must be gone as well

lalottacontinua: ok.

modules/cells/src/main/java/dmg/cells/nucleus/CellAdapter.java
lalottacontinua: and as we not live in an ideal word, should we continue changing the renaming together with the changing of the string-concatenations or should we rather focus on re-re-naming the "logger" to "LOGGER" or "LOG"? or should we rather just change the string-concatenations? tomorrow Marissa and me are going to focus on the other exercise, but we wanted to continue with this one, as far as we got time for it.

kofemann: I would suggest you continue with  string-concatenations and forget about naming issue. Leave it to the next generation :). It will keep your changes small and easy to review.

lalottacontinua: ok. :) we'll do so...

modules/cells/src/main/java/dmg/cells/nucleus/SystemCell.java
```diff
@@ -356,15 +356,15 @@ public void uncaughtException(Thread t, Throwable e)
         Throwable root = Throwables.getRootCause(e);
         if (root instanceof FileNotFoundException) {
             if (root.getMessage().contains("Too many open files")) {
-                _log.error(AlarmMarkerFactory.getMarker(PredefinedAlarm.OUT_OF_FILE_DESCRIPTORS,
+                logger.error(AlarmMarkerFactory.getMarker(PredefinedAlarm.OUT_OF_FILE_DESCRIPTORS,
                                                         getCellDomainName(),
                                                         getCellName()),
-                           "Uncaught exception in thread " + t.getName(),
+                           "Uncaught exception in thread {}", t.getName(),
```
kofemann: this one is an unfortunate change, as you change from log.error(String s, Throwable t) to  log.error(String format, Object o1, Object o1). The first method will log the stack trace as wee, while the second option will show only message from the exception.

kofemann: The following code shows that FAQ from slf4j is wrong:

```java
package net.kofemann.coffeesnack;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public class LoggingTest {

    private final static Logger LOGGER = LoggerFactory.getLogger(LoggingTest.class);

    public static void main (String[] args) {

        LOGGER.error("This is multi args statement [{}, {}, {}] with throwable {}",
                "one", "two", "three", new Exception("fail with args"));


        LOGGER.error("This statement without args, but with a throwable {}", new Exception("fail alone...."));
    }

}
```

The output:
```
--- exec-maven-plugin:1.2.1:exec (default-cli) @ coffeesnack ---
[main] ERROR net.kofemann.coffeesnack.LoggingTest - This is multi args statement [one, two, three] with throwable java.lang.Exception: fail with args
[main] ERROR net.kofemann.coffeesnack.LoggingTest - This statement without args, but with a throwable {}
java.lang.Exception: fail alone....
	at net.kofemann.coffeesnack.LoggingTest.main(LoggingTest.java:16)
------------------------------------------------------------------------
```

marisanest: We checked again this case and got the following which we think show that the FAQ from slf4j is right:

`package diskCacheV111.admin;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public class TestClass {

    private static final Logger _log =
            LoggerFactory.getLogger(TestClass.class);
    public static void main(String [] args)
    {
        String name = "NAME";
        String name2 = "NAME2";

        _log.error("InterruptedException while waiting for a reply from pools " + new Exception("TESTTESTETST"));

        _log.error("InterruptedException while waiting for a reply from pools", new Exception("TESTTESTETST"));

        _log.error("InterruptedException while waiting for a reply from pools {}", new Exception("TESTTESTETST").toString());

        _log.error("InterruptedException while waiting for a reply from pools {}", name ,new Exception("TESTTESTETST"));

        _log.error("InterruptedException while waiting for a reply from pools {}, {}", name , name2, new Exception("TESTTESTETST"));
    }
}
`
The output:

`14:30:58.828 [main] ERROR diskCacheV111.admin.TestClass - InterruptedException while waiting for a reply from pools java.lang.Exception: TESTTESTETST
14:30:58.846 [main] ERROR diskCacheV111.admin.TestClass - InterruptedException while waiting for a reply from pools
java.lang.Exception: TESTTESTETST
	at diskCacheV111.admin.TestClass.main(TestClass.java:17) [classes/:na]
14:30:58.847 [main] ERROR diskCacheV111.admin.TestClass - InterruptedException while waiting for a reply from pools java.lang.Exception: TESTTESTETST
14:30:58.847 [main] ERROR diskCacheV111.admin.TestClass - InterruptedException while waiting for a reply from pools NAME
java.lang.Exception: TESTTESTETST
	at diskCacheV111.admin.TestClass.main(TestClass.java:21) [classes/:na]
14:30:58.847 [main] ERROR diskCacheV111.admin.TestClass - InterruptedException while waiting for a reply from pools NAME, NAME2
java.lang.Exception: TESTTESTETST
	at diskCacheV111.admin.TestClass.main(TestClass.java:23) [classes/:na]
`

kofemann: Interesting....may be it depends on appender...

kofemann: Ok. in did, depending on backend logger implementation I get a different result:
```
14:46:02.149 [main] ERROR net.kofemann.coffeesnack.LoggingTest - This is multi args statement [one, two, three] with throwable {}
java.lang.Exception: fail with args
	at net.kofemann.coffeesnack.LoggingTest.main(LoggingTest.java:24)
14:46:02.153 [main] ERROR net.kofemann.coffeesnack.LoggingTest - This statement without args, but with a throwable {}
java.lang.Exception: fail alone....
	at net.kofemann.coffeesnack.LoggingTest.main(LoggingTest.java:28)
```
which is for me an indication, that we can't relay on it. We stick to javadoc, sorry.

marisanest: That's really interesting... so we will stick to the javadoc :)

PR #3424, opened 2017-08-06
dcache-common: refactoring slf4j logging messages and logger variable name

marisanest:

### DIFF
--- a/modules/common/src/main/java/org/dcache/auth/OpenIdCredentialRefreshable.java
+++ b/modules/common/src/main/java/org/dcache/auth/OpenIdCredentialRefreshable.java
@@ -26,7 +26,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 public class OpenIdCredentialRefreshable extends WrappingOpenIdCredential
 {
-    private static final Logger LOG =
+    private static final Logger logger =
             LoggerFactory.getLogger(OpenIdCredentialRefreshable.class);
     private final HttpClient client;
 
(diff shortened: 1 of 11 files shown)


modules/common/src/main/java/org/dcache/auth/OpenIdCredentialRefreshable.java
```diff
@@ -26,7 +26,7 @@
 
 public class OpenIdCredentialRefreshable extends WrappingOpenIdCredential
 {
-    private static final Logger LOG =
+    private static final Logger logger =
```
marisanest: Hi, thanks for your feedback! We (lotta and I) saw different naming styles within the code. One was "LOG", one "_log" and the last one "logger". We thought that it might be good to unify these names. After some researches we thought that "logger" would be the best name because a logger reference is not a constant value and therefore uppercase is maybe not the best case :) But if you already have a certain naming convention it might be superfluous to change that. Can I make the changes directly in this pull request or should I make a new one without renaming?

modules/common/src/main/java/org/dcache/auth/OpenIdCredentialRefreshable.java
```diff
@@ -42,7 +42,7 @@ public String getBearerToken()
             try {
                 refreshOpenIdCredentials();
             } catch (IOException | AuthenticationException e) {
-                LOG.warn("Error Refreshing OpenId Bearer Token with {}: {}",
+                logger.warn("Error Refreshing OpenId Bearer Token with {}: {}",
```
marisanest: Hi, thanks for the tip! We will consider it in our future requests :)

kofemann: Ok :)

modules/common/src/main/java/org/dcache/auth/OpenIdCredentialRefreshable.java
kofemann: That's the point. You see, that we are already confused, so don't make it worse :). If you want to enforce a single style, then you can do it with a dedicated commit and describe there why.

kofemann:
ok to test

kofemann:
Thanks! Manually merged as eb9b04f981

PR #3426, opened 2017-08-07
Task8

dcachess17:
Motivation: Prior to Java 1.5 enums did not exist, thus integer constants were used. These constants can now be replaced by enums.
    
Modification: diskCacheV111/poolManager/RequestContainerV5  return code constants were replaced by an enum called RequestStatusCode
for instance :    RT_OK  to  RequestStatusCode.OK
    
Result: slightly higher memory consumption for the benefits of type safety and more maintainable code
    
Signed-off-by: Vuong Luu Minh s0548844@htw-berlin.de

### DIFF
--- a/modules/dcache/src/main/java/diskCacheV111/poolManager/RequestContainerV5.java
+++ b/modules/dcache/src/main/java/diskCacheV111/poolManager/RequestContainerV5.java
@@ -716,6 +716,20 @@ public class RequestContainerV5
         return commandReply;
     }
 
+    /**
+     * Return Codes used in PoolRequestHandler
+     */
+    public enum RequestStatusCode {
+        OK,
+        FOUND,
+        NOT_FOUND,
+        ERROR,
+        OUT_OF_RESOURCES,
+        COST_EXCEEDED,
+        NOT_PERMITTED,
+        S_COST_EXCEEDED,
+        DELAY
+    }
     ///////////////////////////////////////////////////////////////
     //
     // the read io request handler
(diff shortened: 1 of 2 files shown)


modules/dcache/src/main/java/diskCacheV111/poolManager/RequestContainerV5.java
```diff
@@ -1396,7 +1407,7 @@ private void stateEngine( Object inputObject ) {
                         return ;
                     }
 
-                    if( ( rc = askIfAvailable() ) == RT_FOUND ){
+                    if( ( rc = askIfAvailable() ) == ReturnCode.FOUND.value ){
```
kofemann: you will need to update askIfAvailable to return enum value instead of int

dcachess17: are you sure you want to do that? ReturnCode.FOUND.value  is an integer, so the comparison works

kofemann: Sure, but then you use enums just as a decorators around integers. If **askIfAvailable** will return **11**,  your code will compile and probably work. Nevertheless, **11** is not a legal value. On the other hand, if it returns enum ReturnCode (which is btw not a great name) then it can return only values defined by the enum. This makes your code more robust.

dcachess17: Thanks for your explaination, i'll change the code. Which name do you suggest for the enum ? : D

kofemann: This is the most difficult question :). RequestStatusCode or so....

dcachess17: I've reworked the code. I hope i didn't miss anything.

kofemann:
ok to test

kofemann:
I will prefer to squash the commits. Can you do that?

dcachess17:
I squashed them and opened a new pullrequest. https://github.com/dCache/dcache/pull/3429

kofemann:
Ok. Thanks.

PR #3429, opened 2017-08-08
poolManager.RequestContainerV5: Integer constants replaced by enum called RequestStatusCode

dcachess17:
Motivation: Prior to Java 1.5 enums did not exist, thus integer constants were used. These constants can now be replaced by enums.

Modification: diskCacheV111/poolManager/RequestContainerV5  return code constants were replaced by an enum called RequestStatusCode
for instance :    RT_OK  to  RequestStatusCode.OK

Result: slightly higher memory consumption for the benefits of type safety and more maintainable code

Signed-off-by: Vuong Luu Minh s0548844@htw-berlin.de

### DIFF
--- a/modules/dcache/src/main/java/diskCacheV111/poolManager/RequestContainerV5.java
+++ b/modules/dcache/src/main/java/diskCacheV111/poolManager/RequestContainerV5.java
@@ -716,6 +716,20 @@ public class RequestContainerV5
         return commandReply;
     }
 
+    /**
+     * Return Codes used in PoolRequestHandler
+     */
+    public enum RequestStatusCode {
+        OK,
+        FOUND,
+        NOT_FOUND,
+        ERROR,
+        OUT_OF_RESOURCES,
+        COST_EXCEEDED,
+        NOT_PERMITTED,
+        S_COST_EXCEEDED,
+        DELAY
+    }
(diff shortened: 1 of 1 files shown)


modules/dcache/src/main/java/diskCacheV111/poolManager/RequestContainerV5.java
```diff
@@ -1080,15 +1094,7 @@ private boolean answerRequest(int count) {
...
-        private static final int RT_NOT_FOUND  = 3 ;
-        private static final int RT_ERROR      = 4 ;
-        private static final int RT_OUT_OF_RESOURCES = 5 ;
-        private static final int RT_COST_EXCEEDED    = 7 ;
-        private static final int RT_NOT_PERMITTED    = 8 ;
-        private static final int RT_S_COST_EXCEEDED  = 9 ;
-        private static final int RT_DELAY  = 10 ;
+
 
         private static final int CONTINUE        = 0 ;
```
alrossi: The way the state machine is implemented here of course could be rewritten, which is not the task at hand, but I am wondering whether a package-visible enum for these two states might not also be desirable.

kofemann:
ok to test

alrossi:
It's fine.

kofemann:
Thanks!

PR #3430, opened 2017-08-08
dcache: added a decorator for RepositoryChannels to get IO statistics

marisanest:
Motivation: 

Getting a possibility to generate statistics and spot misbehaveour in IO processes of RepositoryChannel.

Modification: 

Added package statistics, created the class IoStatisticsChannel as decorator for a RepositoryChannel and created the class IoStatistics which provides statistics on number of read and write requests, read and write IO block sizes and different measures about the read and write IO speed.

Result: 

Possibility to add a decorator to a RepositoryChannel which provides statistics about IO processes.

Signed-off-by: Lotta Rüger s0551814@htw-berlin.de
Signed-off-by: marisanest marisanest@mailbox.org

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
+++ b/modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
@@ -3,10 +3,6 @@ package org.dcache.pool.statistics;
 import org.apache.commons.math3.stat.descriptive.rank.Percentile;
 import com.google.common.base.MoreObjects;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-
 /**
  * This class stores statistics about read an write processes
  */
@@ -46,11 +42,11 @@ public class IoStatistics {
      *         The duration of the read request in nanoseconds; must be non-negative
      */
 
-    public void updateRead(long readBytes, long readTime) throws ArrayIndexOutOfBoundsException{
+    public void updateRead(long readBytes, long readTime) {
 
         double readSpeed = calculateSpeed(readBytes, readTime);
         _readSpeeds[_readSpeedsArrayIndex] = readSpeed;
-        _readSpeedsArrayIndex++;
+        _readSpeedsArrayIndex = (_readSpeedsArrayIndex + 1) % _readSpeeds.length;
 
         _minReadSpeed = (_readRequestNum == 0) ? readSpeed : Math.min(readSpeed, _minReadSpeed);
         _maxReadSpeed = (_readRequestNum == 0) ? readSpeed : Math.max(readSpeed, _maxReadSpeed);
@@ -73,11 +69,11 @@ public class IoStatistics {
      *         The duration of the write request in nanoseconds; must be non-negative
      */
 
-    public void updateWrite(long writeBytes, long writeTime) throws ArrayIndexOutOfBoundsException{
+    public void updateWrite(long writeBytes, long writeTime) {
 
         double writeSpeed = calculateSpeed(writeBytes, writeTime);
         _writeSpeeds[_writeSpeedsArrayIndex] = writeSpeed;
-        _writeSpeedsArrayIndex++;
+        _writeSpeedsArrayIndex = (_writeSpeedsArrayIndex + 1) % _writeSpeeds.length;
 
 
         _minWriteSpeed = (_writeRequestNum == 0) ? writeSpeed : Math.min(writeSpeed, _minWriteSpeed);
@@ -100,11 +96,11 @@ public class IoStatistics {
      * @param  time
      *         The duration of the request in milliseconds; must be non-negative
      *
-     * @return The speed of the process in bytes per nanoseconds.
+     * @return The speed of the process in bytes per seconds.
      */
 
     private double calculateSpeed(long bytes, long time){
-        return (double) bytes / (double) time;
+        return (double) bytes / ((double) time / 1000000000);
     }
 
     /**
--- a/modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
+++ b/modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
@@ -3,6 +3,8 @@ package org.dcache.pool.statistics;
 import org.apache.commons.math3.stat.descriptive.rank.Percentile;
 import com.google.common.base.MoreObjects;
 
+import java.util.concurrent.TimeUnit;
+
 /**
  * This class stores statistics about read an write processes
  */
@@ -10,7 +12,6 @@ import com.google.common.base.MoreObjects;
 public class IoStatistics {
 
     private long _readRequestNum = 0;
-    private long _readBytes;
     private double[] _readSpeeds = new double[8192];
     private int _readSpeedsArrayIndex = 0;
     private double _minReadSpeed;
@@ -18,9 +19,11 @@ public class IoStatistics {
     private double _avgReadSpeed;
     private double _95ReadSpeed;
     private long _totalReadTime;
+    private long _totalReadBytes;
+    private long _totalRequestedBytes;
+    private long _avgRequestedBytes;
 
     private long _writeRequestNum = 0;
-    private long _writtenBytes;
     private double[] _writeSpeeds = new double[8192];
     private int _writeSpeedsArrayIndex = 0;
     private double _minWriteSpeed;
@@ -28,6 +31,7 @@ public class IoStatistics {
     private double _avgWriteSpeed;
     private double _95WriteSpeed;
     private long _totalWriteTime;
+    private long _totalWrittenBytes;
 
     private final Percentile percentile = new Percentile(0.95);
 
@@ -42,20 +46,25 @@ public class IoStatistics {
      *         The duration of the read request in nanoseconds; must be non-negative
      */
 
-    public void updateRead(long readBytes, long readTime) {
+    public void updateRead(long readBytes, long readTime, long requestedBytes) {
 
-        double readSpeed = calculateSpeed(readBytes, readTime);
-        _readSpeeds[_readSpeedsArrayIndex] = readSpeed;
-        _readSpeedsArrayIndex = (_readSpeedsArrayIndex + 1) % _readSpeeds.length;
+        if(readBytes >= 0) {
+            double readSpeed = calculateSpeed(readBytes, readTime);
+            _readSpeeds[_readSpeedsArrayIndex] = readSpeed;
+            _readSpeedsArrayIndex = (_readSpeedsArrayIndex + 1) % _readSpeeds.length;
 
-        _minReadSpeed = (_readRequestNum == 0) ? readSpeed : Math.min(readSpeed, _minReadSpeed);
-        _maxReadSpeed = (_readRequestNum == 0) ? readSpeed : Math.max(readSpeed, _maxReadSpeed);
-        _avgReadSpeed = (_readRequestNum == 0) ? readSpeed : calculateNewAvg(_readRequestNum,_avgReadSpeed, readSpeed);
-        _95ReadSpeed = (_readRequestNum == 0) ? readSpeed : percentile.evaluate(_readSpeeds, 0, (_readSpeedsArrayIndex - 1));
+            _minReadSpeed = (_readRequestNum == 0) ? readSpeed : Math.min(readSpeed, _minReadSpeed);
+            _maxReadSpeed = (_readRequestNum == 0) ? readSpeed : Math.max(readSpeed, _maxReadSpeed);
+            _avgReadSpeed = (_readRequestNum == 0) ? readSpeed : calculateNewAvg(_readRequestNum,_avgReadSpeed, readSpeed);
+            _95ReadSpeed = (_readRequestNum == 0) ? readSpeed : percentile.evaluate(_readSpeeds, 0, (_readSpeedsArrayIndex - 1));
 
+            _totalReadBytes += readBytes;
+        }
         _totalReadTime += readTime;
         _readRequestNum ++;
-        _readBytes += readBytes;
+
+        _totalRequestedBytes += requestedBytes;
+        _avgRequestedBytes += (_readRequestNum == 0) ? _avgRequestedBytes : calculateNewAvg(_readRequestNum,_avgRequestedBytes, requestedBytes);
     }
 
     /**
@@ -83,7 +92,7 @@ public class IoStatistics {
 
         _totalWriteTime += writeTime;
         _writeRequestNum ++;
-        _writtenBytes += writeBytes;
+        _totalWrittenBytes += writeBytes;
     }
 
     /**
@@ -100,7 +109,7 @@ public class IoStatistics {
      */
 
     private double calculateSpeed(long bytes, long time){
-        return (double) bytes / ((double) time / 1000000000);
+        return (double) bytes / ((double) time / TimeUnit.SECONDS.toNanos(1));
     }
 
     /**
@@ -127,8 +136,8 @@ public class IoStatistics {
         return _readRequestNum;
     }
 
-    public long getReadBytes() {
-        return _readBytes;
+    public long getTotalReadBytes() {
+        return _totalReadBytes;
     }
 
     public double getMinReadSpeed() {
@@ -155,8 +164,8 @@ public class IoStatistics {
         return _writeRequestNum;
     }
 
-    public long getWrittenBytes() {
-        return _writtenBytes;
+    public long getTotalWrittenBytes() {
+        return _totalWrittenBytes;
     }
 
     public double getMinWriteSpeed() {
@@ -179,18 +188,28 @@ public class IoStatistics {
         return _totalWriteTime;
     }
 
+    public long getTotalRequestedBytes() {
+        return _totalRequestedBytes;
+    }
+
+    public long getAvgRequestedBytes() {
+        return _avgRequestedBytes;
+    }
+
     public String toString(){
 
        return MoreObjects.toStringHelper(this)
                 .add("number of read requests:", _readRequestNum)
-                .add("total read bytes:", _readBytes)
                 .add("min readSpeed:", _minReadSpeed)
                 .add("max readSpeed:", _maxReadSpeed)
                 .add("avg readSpeed:", _avgReadSpeed)
                 .add("95percentile readSpeed:", _95ReadSpeed)
                 .add("total readTime:", _totalReadTime)
+                .add("total read bytes:", _totalReadBytes)
+                .add("total requested bytes:", _totalRequestedBytes)
+                .add("avg requested bytes:", _totalRequestedBytes)
                 .add("number of write requests:", _writeRequestNum)
-                .add("total written bytes:", _writtenBytes)
+                .add("total written bytes:", _totalWrittenBytes)
                 .add("min writeSpeed:", _minWriteSpeed)
                 .add("max writeSpeed:", _maxWriteSpeed)
                 .add("avg writeSpeed:", _avgWriteSpeed)
--- a/modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
+++ b/modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
@@ -62,7 +62,7 @@ public class IoStatistics {
      *         The number of written bytes; must be non-negative
      *
      * @param  writeTime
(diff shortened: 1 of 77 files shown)


modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
```diff
@@ -0,0 +1,168 @@
...
+ */
+public class IoStatisticsChannel implements RepositoryChannel {
+
+    private static final Logger LOGGER =
+            LoggerFactory.getLogger(IoStatisticsChannel.class);
+
+    /**
+     * Inner channel to which all operations are delegated.
+     */
+    @VisibleForTesting
```
alrossi: I don't believe this is going to work.  You need to make the field package visible and put the test class in the same package.  This annotation is for informational purposes only.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
```diff
@@ -0,0 +1,168 @@
...
+        _channel.sync();
+    }
+
+    @Override
+    public long transferTo(long position, long count, WritableByteChannel target) throws IOException {
+        long startTime = System.currentTimeMillis();
+        long readBytes =  _channel.transferTo(position, count, target);
+        long duration = System.currentTimeMillis() - startTime;
+        _statistics.updateRead(readBytes, duration);
+        // long discrepance = count - readBytes;
```
alrossi: discrepancy?

if this is unused, please remove.  We do not commit code with commented-out statements.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
```diff
@@ -0,0 +1,168 @@
...
+        return readBytes;
+    }
+
+    @Override
+    public long transferFrom(ReadableByteChannel src, long position, long count) throws IOException {
+        long startTime = System.currentTimeMillis();
+        long writtenBytes =  _channel.transferFrom(src, position, count);
+        long duration = System.currentTimeMillis() - startTime;
+        _statistics.updateWrite(writtenBytes, duration);
+        // long discrepance = count - writtenBytes;
```
alrossi: idem

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
```diff
@@ -0,0 +1,168 @@
...
+        return writtenBytes;
+    }
+
+    @Override
+    public long write(ByteBuffer[] srcs, int offset, int length) throws IOException {
```
alrossi: It seems you are not actually implementing the delegated method here.

long writtenBytes = _channel.write(srcs, offset, length) ?

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
```diff
@@ -0,0 +1,168 @@
...
+        long startTime = System.currentTimeMillis();
+        int readBytes = _channel.read(dst);
+        long duration = System.currentTimeMillis() - startTime;
+        _statistics.updateRead(readBytes, duration);
+        return readBytes;
+    }
+
+    @Override
+    public int write(ByteBuffer src) throws IOException {
+        // long supposedBytes = src.limit() - src.position();
```
alrossi: please remove unused code

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
```diff
@@ -0,0 +1,205 @@
...
+     *         The number of read or write requests so far; must be non-negative
+     *
+     * @param  speeds
+     *         All speeds of all read or write requests so far.
+     *
+     * @return The calculated 95% percentile of the speed of all read or write requests.
+     */
+
+    private float calculate95Percentile(long requestNum, ArrayList<Float> speeds){
+        Collections.sort(speeds);
```
alrossi: This is not essential, but you do realize it is not efficient to sort the list each time you recalculate.

It would be better if you could maintain the sort order on insert/add.  But  this is just a suggestion.

alrossi: Actually, now that I think about this, is it such a good idea to maintain this list?  We could eventually hit a memory limit.  Technically this is a leak.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
lalottacontinua: ok. thanks. I'll make it package visible. but should I leave the annotation?

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * This class stores statistics about read an write processes
+ */
+
+public class IoStatistics {
+
+    private long _readRequestNum = 0;
+    private long _readBytes;
+    private ArrayList<Float> _readSpeeds = new ArrayList<Float>();
```
marisanest: Hi, no there is no reason for using float so we will change all of them to double. Thanks for the hint :)

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
lalottacontinua: idem.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
lalottacontinua: sure. I will remove it.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
lalottacontinua: oh, copy paste failure. :/ thanks!

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
lalottacontinua: yes. thanks.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
marisanest: Thats a good point, we don't considered the memory limit. We would change that immediately but we don't know how to calculate the 95%  percentile without a list of all speed values. Do you have an idea?

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
alrossi: It makes sense only if you actually write the unit test that uses it.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
alrossi: Is the 95% percentile a necessary statistic?  

(I did not read the assignment)

marisanest: @kofemann asked us to do so.

alrossi: @kofemann, do you have a suggestion here, or is the memory limit not an issue in your opinion?

lalottacontinua:
sry. there happened a failure. wrong commits have been pushed from module cell. will try to revert.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
alrossi: We have decided to accept this for the moment, as the more efficient implementation would require too much work.   We can do that ourselves later.

kofemann: yes, I agree with @alrossi. You need to use some kind moving median calculation, lke described here https://arxiv.org/pdf/1406.1717.pdf. But I think it will be too much for Thursday :)

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
```diff
@@ -0,0 +1,168 @@
...
+     * Returns the object most central of this decorator
+     * @return object with collected and evaluated statistics data
+     */
+    public IoStatistics getStatistics() {
+        return _statistics;
+    }
+
+    @Override
+    public int write(ByteBuffer buffer, long position) throws IOException {
+        long startTime = System.currentTimeMillis();
```
kofemann: The IO request can be quite small and with SSD ~200 microseconds. Please use System.nanoTime() to calculate time delta. See http://download.java.net/java/jdk9/docs/api/java/lang/System.html#nanoTime-- for details.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
```diff
@@ -0,0 +1,205 @@
...
+
+    public long getReadRequestNum() {
+        return _readRequestNum;
+    }
+
+    public long getReadBytes() {
+        return _readBytes;
+    }
+
+    public ArrayList<Double> getReadSpeeds() {
```
alrossi: Sorry I missed this one the first time.

You should return the abstract type rather than the concrete type, i.e., List<Double>.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
```diff
@@ -0,0 +1,205 @@
...
+
+    public long getWriteRequestNum() {
+        return _writeRequestNum;
+    }
+
+    public long getWrittenBytes() {
+        return _writtenBytes;
+    }
+
+    public ArrayList<Double> getWriteSpeeds() {
```
alrossi: Here too.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * This class stores statistics about read an write processes
+ */
+
+public class IoStatistics {
+
+    private long _readRequestNum = 0;
+    private long _readBytes;
+    private ArrayList<Double> _readSpeeds = new ArrayList<>();
```
alrossi: the fields should also be defined using the abstract interface List<Double>

My oversight.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
```diff
@@ -0,0 +1,205 @@
...
+    private double _minReadSpeed;
+    private double _maxReadSpeed;
+    private double _avgReadSpeed;
+    private double _95ReadSpeed;
+    private long _totalReadTime;
+
+    private long _writeRequestNum = 0;
+    private long _writtenBytes;
+    private ArrayList<Double> _writeSpeeds = new ArrayList<>();
```
kofemann: you can have here a fixed size array of doubles, let say  8192 values. BTW, we already use commons-math library, which already provides percentile computation.

marisanest: Shall we change this?

alrossi: I agree, maybe this should be substituted for 
http://commons.apache.org/proper/commons-math/javadocs/api-3.5/org/apache/commons/math3/stat/descriptive/rank/Percentile.html

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatisticsChannel.java
lalottacontinua: do you know why this doesn't mark as outdated?

kofemann:
To test your code, you can update AbstractMover#openChannel() to decorate returned channel. Add toString method to of your IoStatistics class and print if with System.out.println(). The statistics should be available in the log file. Please use toString helper like in org.dcache.vehicles.FileAttributes.toString() method.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
alrossi: There's also probably little reason to provide a public getter for that list/array.

lalottacontinua:
Why should we do update AbstractMover#openChannel()? Is that class tested, and we'll be able to thereby test our decorator? 
I've been starting to set up a mock-channel as done in ChecksumChannelTest. I just find it difficult to use the methods, like write(ByteBuffer buffer, long position) where I'm supposed to know what position I should hand in to the method. I could make a pull request on another branch, where the test class is in, if you want to have a look?

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
```diff
@@ -39,15 +41,16 @@
      *         The duration of the read request in nanoseconds; must be non-negative
      */
 
-    public void updateRead(long readBytes, long readTime){
+    public void updateRead(long readBytes, long readTime) throws ArrayIndexOutOfBoundsException{
 
         double readSpeed = calculateSpeed(readBytes, readTime);
-        _readSpeeds.add(readSpeed);
+        _readSpeeds[_readSpeedsArrayIndex] = readSpeed;
```
alrossi: what happens when the array index >= 8192?

I think you need to make this a circular buffer.  The percentile computation then is valid only over the last 8192 values.

You would need to do something likeL

_readSpeedsArrayIndex = (_readSpeedsArrayIndex + 1) % _readSpeeds.len;

to find the index of the next value to be inserted.

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
```diff
@@ -65,14 +68,17 @@ public void updateRead(long readBytes, long readTime){
      *         The duration of the write request in nanoseconds; must be non-negative
      */
 
-    public void updateWrite(long writeBytes, long writeTime){
+    public void updateWrite(long writeBytes, long writeTime) throws ArrayIndexOutOfBoundsException{
+
         double writeSpeed = calculateSpeed(writeBytes, writeTime);
-        _writeSpeeds.add(writeSpeed);
+        _writeSpeeds[_writeSpeedsArrayIndex] = writeSpeed;
+        _writeSpeedsArrayIndex++;
```
alrossi: same here

alrossi: Sorry if we are making your life difficult :-)

But I think we're almost there  ...

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
```diff
@@ -42,20 +46,25 @@
...
+        if(readBytes >= 0) {
+            double readSpeed = calculateSpeed(readBytes, readTime);
+            _readSpeeds[_readSpeedsArrayIndex] = readSpeed;
+            _readSpeedsArrayIndex = (_readSpeedsArrayIndex + 1) % _readSpeeds.length;
 
-        _minReadSpeed = (_readRequestNum == 0) ? readSpeed : Math.min(readSpeed, _minReadSpeed);
-        _maxReadSpeed = (_readRequestNum == 0) ? readSpeed : Math.max(readSpeed, _maxReadSpeed);
-        _avgReadSpeed = (_readRequestNum == 0) ? readSpeed : calculateNewAvg(_readRequestNum,_avgReadSpeed, readSpeed);
-        _95ReadSpeed = (_readRequestNum == 0) ? readSpeed : percentile.evaluate(_readSpeeds, 0, (_readSpeedsArrayIndex - 1));
+            _minReadSpeed = (_readRequestNum == 0) ? readSpeed : Math.min(readSpeed, _minReadSpeed);
```
alrossi: One last observation.

Instead of checking the counter each time, you could simply preset/initialize _minReadSpeed to Double.MAX_VALUE and _maxReadSpeed to Double.MIN_VALUE, and then just call Math.min and Math.max.

(I would imagine the apache common code for Percentile would also handle a 0, -1 boundary OK, but that would need checking)

For average, if you examine your method, you should be able to call it without checking (on the first try it would be (0*oldAvg + newSpeed)/1, which equals newSpeed, no matter what oldAvg is, but you could just initialize the average value to 0 as well.

The code does a bit less work this way.

kofemann:
@lalottacontinua the update of was just to let you to try your code. If you get it tested as unit test it will be great!

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
alrossi: Actually, I misspoke.  the percentile.evaluate should work on the first call because (a) you already added a value to the array and (b) you already incremented your index.  The second value should be the effective length of the array, not the last index, BTW.

percentile.evaluate(_readSpeeds, 0, _readSpeedsArrayIndex) is correct, I think.  In the case of the first call, you would have an array of length 1, and that is the third parameter:

--> evaluate(double[] values, int start, int length)

The problem comes when your index reaches length and turns over.  Your computation then will get messed up because you are calling the modulo index length.  

So we need to do something else there.  Perhaps using the _request counter and doing

percentile.evaluate(_readSpeeds, 0, Math.min(_readSpeeds.length, _readRequestNum) ?

kofemann:
ok to test

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
marisanest: These are all good points and I fixed it! Thank for your patience :)

The only problem I see in initializing _minReadSpeed with Double.MAX_VALUE and _maxReadSpeed with Double.MIN_VALUE is that when the methods updateRead and updateWrite are never called but the statistics are however evaluated the values of these to parameters could be confusing or are they not?

alrossi:
I'm not sure whether it makes more sense to include or exclude the reads which return -1 bytes from the average or not, but this looks good now.

alrossi:
ok to test

modules/dcache/src/main/java/org/dcache/pool/statistics/IoStatistics.java
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    public long getTotalRequestedWriteBytes() {
+        return _totalRequestedWriteBytes;
+    }
+
+    public double getAvgRequestedWriteBytes() {
+        return _avgRequestedWriteBytes;
+    }
+
```
kofemann: Please add @Override annotation. By this we ensure that mistyped methods names will be catch at compile time.

kofemann:
@lalottacontinua , @marisanest I think it's good to go. Are you working on unit test? The other thing, it's quite some number of commits. Do you want to squash them? Or shall we squash them? Or you want it get merged as-is?

lalottacontinua:
@kofemann Right now I'm working on the presentation for tomorrow. I would say, squash it and merge it, if you want to, and then I can see tomorrow, if I'll have time to finish the Test? I would in that case just start a new pull request. Is that an idea?

kofemann:
@lalottacontinua sounds like a plan.

marisanest:
@alrossi In my opinion it makes more sense to exclude the new calculation of the average read speed if the read request return -1 because we don't have a useful speed value at that point (we could assign that value to 0 but that would be not true because there are no bytes transferred at all) and for the calculation of the new average requested read bytes I think it makes also more sense to include that because it only take the requested bytes in calculation and not the effective read bytes. So I would let it as it is. Would that be fine with you?

marisanest:
@kofemann is there an easy way we can do the squashing?

alrossi:
@marisanest yes, it's fine.  I think we can merge.

kofemann:
@marisanest If you are uncomfortable with it, we will do it. I will show you tomorrow how you can do it in yourself.

kofemann:
@marisanest, @lalottacontinua  thanks!

lalottacontinua:
Thank you! :)

PR #3431, opened 2017-08-08
core: Add Remote-Host-Restriction capability to Ssh2Admin PublicKeyAu…

solen0id:
…thenticator

Motivation:
To increase security when authentication users to the AdminInterface,
one can now optionally specify restrictions on which remote hosts are
allowed to authenticate with a given public key in the authorized_key
file. This makes using stolen keys more difficult.

Modification:
Add RemoteHostValidator subclass to Ssh2Admins PublicKeyAuthenticator.
Update the authenticate method to evaluate restrictions from authorized_keys.

Result:
Public key authentication for the Admin-Interface will only succeed if
the remote host can satisfy all restrictions or no restrictions are set
for the associated key

Target: master
Require-notes: yes
Require-book: yes

Signed-off-by: Max Patzelt <max.patzelt@gmail.com>

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
+++ b/modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
@@ -364,7 +364,7 @@ public class Ssh2Admin implements CellCommandListener, CellLifeCycleAware
                 } catch (IllegalArgumentException e) {
                     String remoteName = ((InetSocketAddress) remote).getHostName();
                     if (Glob.isGlob(pattern)) {
-                        Pattern p = convertPatternToRegex(pattern);
+                        Pattern p = Glob.parseGlobToPattern(pattern);
                         Matcher addressMatcher = p.matcher(remoteAddress);
                         Matcher nameMatcher = p.matcher(remoteName);
                         patternMatches = (addressMatcher.matches() || nameMatcher.matches());
@@ -390,11 +390,6 @@ public class Ssh2Admin implements CellCommandListener, CellLifeCycleAware
                     return false;
                 }
             }
-
-        private Pattern convertPatternToRegex(String pattern) {
-            pattern = pattern.replace(".", "\\.");
-            return Glob.parseGlobToPattern(pattern);
-        }
 
         private String strip(String linePart) {
             return linePart.substring(6, linePart.length()-1);
--- a/modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
+++ b/modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
@@ -339,9 +339,10 @@ public class Ssh2Admin implements CellCommandListener, CellLifeCycleAware
 
         private boolean isValidHost(String line, ServerSession session) {
             for (String linePart : Splitter.on(" ").trimResults().omitEmptyStrings().split(line)) {
-                if (linePart.startsWith("from=") && linePart.endsWith("\"")) {
+                if (linePart.startsWith("from=\"") && linePart.endsWith("\"")) {
+                    String from = linePart.substring(6, linePart.length()-1);
                     Set<Outcome> outcomes = EnumSet.noneOf(Outcome.class);
-                    for (String pattern : Splitter.on(",").trimResults().omitEmptyStrings().split(strip(linePart))) {
+                    for (String pattern : Splitter.on(",").trimResults().omitEmptyStrings().split(from)) {
                         outcomes.add(patternMatchesHost(pattern, session));
                     }
                     return outcomes.contains(Outcome.ALLOW) && !outcomes.contains(Outcome.DENY);
@@ -351,8 +352,8 @@ public class Ssh2Admin implements CellCommandListener, CellLifeCycleAware
         }
 
         private Outcome patternMatchesHost(String pattern, ServerSession session) {
-            boolean patternMatches = false;
             boolean patternIsNegated = pattern.startsWith("!");
+            Outcome matchOutcome = patternIsNegated ? Outcome.DENY : Outcome.ALLOW;
             if (patternIsNegated) {
                 pattern = pattern.substring(1);
             }
@@ -360,26 +361,18 @@ public class Ssh2Admin implements CellCommandListener, CellLifeCycleAware
             if (remote instanceof InetSocketAddress) {
                 String remoteAddress = ((InetSocketAddress) remote).getAddress().getHostAddress();
                 try {
-                    patternMatches = addressInCidrRange(pattern, remoteAddress);
+                    if (addressInCidrRange(pattern, remoteAddress)) {
+                        return matchOutcome;
+                    }
                 } catch (IllegalArgumentException e) {
                     String remoteName = ((InetSocketAddress) remote).getHostName();
-                    if (Glob.isGlob(pattern)) {
-                        Pattern p = Glob.parseGlobToPattern(pattern);
-                        Matcher addressMatcher = p.matcher(remoteAddress);
-                        Matcher nameMatcher = p.matcher(remoteName);
-                        patternMatches = (addressMatcher.matches() || nameMatcher.matches());
-                    } else {
-                        patternMatches = pattern.equals(remoteName);
+                    Glob glob = new Glob(pattern);
+                    if (glob.matches(remoteName) || glob.matches(remoteAddress)) {
+                        return matchOutcome;
                     }
                 }
             }
-            if (patternIsNegated && patternMatches) {
-                return Outcome.DENY;
-            } else if (!patternIsNegated && patternMatches) {
-                return Outcome.ALLOW;
-            } else {
-                return Outcome.DEFER;
-            }
+            return Outcome.DEFER;
         }
 
         private boolean addressInCidrRange(String cidrAddress, String remoteAddress) throws IllegalArgumentException {
@@ -390,10 +383,6 @@ public class Ssh2Admin implements CellCommandListener, CellLifeCycleAware
                     return false;
                 }
             }
-
-        private String strip(String linePart) {
-            return linePart.substring(6, linePart.length()-1);
-        }
     }
 
     private class AdminConnectionLogger implements SessionListener {
--- a/modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
+++ b/modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
@@ -82,6 +82,7 @@ public class Ssh2Admin implements CellCommandListener, CellLifeCycleAware
     private LoginStrategy _loginStrategy;
     private TimeUnit _idleTimeoutUnit;
     private long _idleTimeout;
+    private enum Outcome { ALLOW, DENY, DEFER }
 
     public Ssh2Admin() {
         _server = SshServer.setUpDefaultServer();
@@ -291,12 +292,7 @@ public class Ssh2Admin implements CellCommandListener, CellLifeCycleAware
             }
             return null;
         }
-
-        private boolean remoteHostIsPermitted(String keyLine, ServerSession session) {
-            RemoteHostValidator validator = new RemoteHostValidator();
-            return validator.isValidHost(keyLine, session);
-        }
-
+        
         @Override
         public boolean authenticate(String userName, PublicKey key,
                 ServerSession session) {
@@ -308,7 +304,7 @@ public class Ssh2Admin implements CellCommandListener, CellLifeCycleAware
                     successful =
                         fileStream
                         .filter(l -> !l.isEmpty() && !l.matches(" *#.*"))
-                        .filter(l -> remoteHostIsPermitted(l, session))
+                        .filter(l -> isValidHost(l, session))
                         .map(this::toPublicKey)
                         .filter(Objects::nonNull)
                         .anyMatch(key::equals);
@@ -328,13 +324,6 @@ public class Ssh2Admin implements CellCommandListener, CellLifeCycleAware
                 logLoginTry(userName, session, method, successful, null);
             }
             return successful;
-        }
-    }
-
-    private static class RemoteHostValidator {
-
-        private enum Outcome {
-            ALLOW, DENY, DEFER
         }
 
         private boolean isValidHost(String line, ServerSession session) {
--- a/modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
+++ b/modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
@@ -1,5 +1,6 @@
 package org.dcache.services.ssh2;
 
+import com.google.common.base.Splitter;
 import org.apache.sshd.common.Factory;
 import org.apache.sshd.common.FactoryManager;
 import org.apache.sshd.common.NamedFactory;
@@ -14,6 +15,7 @@ import org.apache.sshd.server.SshServer;
 import org.apache.sshd.server.auth.password.PasswordAuthenticator;
 import org.apache.sshd.server.auth.pubkey.PublickeyAuthenticator;
 import org.apache.sshd.server.session.ServerSession;
+import org.dcache.util.Glob;
 import org.dcache.util.Subnet;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -31,8 +33,10 @@ import java.net.UnknownHostException;
 import java.security.NoSuchAlgorithmException;
 import java.security.PublicKey;
 import java.security.spec.InvalidKeySpecException;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -288,7 +292,7 @@ public class Ssh2Admin implements CellCommandListener, CellLifeCycleAware
             return null;
         }
 
-        private boolean remoteHostIsPermitted(String keyLine, ServerSession session){
+        private boolean remoteHostIsPermitted(String keyLine, ServerSession session) {
             RemoteHostValidator validator = new RemoteHostValidator();
             return validator.isValidHost(keyLine, session);
         }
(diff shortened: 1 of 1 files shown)


kofemann:
ok to test

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
...
+            for (String linePart : line.split(" ")) {
+                if (linePart.startsWith("from=")) {
+                    for (String pattern : strip(linePart).split(",")){
+                        boolean patternIsNegated = pattern.startsWith("!");
+                        boolean patternMatchesHost = patternMatchesHost(pattern, session);
+                        if (patternIsNegated==patternMatchesHost){
+                            return false;
+                        }
+                    }
+                }
```
paulmillar: Should return `false` if line contains a `from=""` stanza and no patterns match.

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
...
 
+    private class RemoteHostValidator {
+
+        private boolean isValidHost(String line, ServerSession session) {
+            for (String linePart : line.split(" ")) {
+                if (linePart.startsWith("from=")) {
+                    for (String pattern : strip(linePart).split(",")){
+                        boolean patternIsNegated = pattern.startsWith("!");
+                        boolean patternMatchesHost = patternMatchesHost(pattern, session);
+                        if (patternIsNegated==patternMatchesHost){
```
paulmillar: The result can short-cut only if the pattern is a deny-pattern.  If an allow-pattern matches then subsequent deny-patterns should be considered.

An allow-pattern match should be remembered, as it should affect the return value.

Also coding style: space either side of `==` and  between parenthesis and brace: `) {`

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
         }
     }
 
+    private class RemoteHostValidator {
+
+        private boolean isValidHost(String line, ServerSession session) {
```
paulmillar: I may have missed it, but the ssh documentation doesn't really say how multiple entries should be combined to come up with an overall authorisation decision.  I had to go to the C source to get a definitive answer.  Basically any deny statements (i.e., starting with `!`) matches the the request is denied, otherwise the request is accepted only if it matches one or more allow statements (i.e., not starting with `!`).

The current semantics of this method are wrong: it short-cuts the tests if an allow pattern matches, it returns false if an allow pattern matches, and it returns true of no pattern matches.

Finally, given the rather complicated nature of parsing (and poor documentation from OpenSSH), it would be good to include some unit tests, so we can capture what "correct behaviour" means.

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
...
+
+        private String convertPatternToRegex(String pattern){
+            pattern = pattern.replace("!","");
+            pattern = pattern.replace(".", "\\.");
+            pattern = pattern.replace("?", ".{1}");
+            pattern = pattern.replace("*", ".*");
+            return pattern;
+        }
+
+        private String strip(String linePart){
```
paulmillar: code-style: space between `)` and `{`; i.e., `) {`

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
...
+                Matcher nameMatcher = p.matcher(remoteName);
+                return (addressMatcher.matches()
+                        || nameMatcher.matches()
+                        || addressInCidrRange(pattern, remoteAddress));
+            }
+            return false;
+        }
+
+        private boolean addressInCidrRange(String cidrAddress, String remoteAddress) {
+            cidrAddress = cidrAddress.replace("!","");
```
paulmillar: This code, to remove of the `!`, appears both here and convertPatternToRegex, suggesting it should be handled in a common location (e.g., `patternMatchesHost`).  Also, since you know it must be the first character, removing it can be done much simpler (and faster): just create a substring starting from the second character.

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
         }
     }
 
+    private class RemoteHostValidator {
```
solen0id: I reckoned it might make sense to encapsulate all the host validation related functionalities in this inner class, but if this goes against convention or is unnecessary we can get rid of it!

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
solen0id: You're absolutely right, I've updated the behavior to combine multiple entries according to this specification. I will also add the unit tests in a seperate PR

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -283,6 +288,11 @@ private PublicKey toPublicKey(String s) {
             return null;
         }
 
+        private boolean remoteHostIsPermitted(String keyLine, ServerSession session){
```
solen0id: roger that!

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
solen0id: I have updated the method to make a decision based on a combination of all entries, as you suggested below. Testing with multiple combinations on my home network I was able to confirm that all the cases you mentioned are now handled correctly. I will write some unit tests and open a new PR for them

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
         }
     }
 
+    private class RemoteHostValidator {
+
+        private boolean isValidHost(String line, ServerSession session) {
+            for (String linePart : line.split(" ")) {
```
solen0id: great package, thanks for the tip! :+1:

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
...
+                if (linePart.startsWith("from=")) {
```
solen0id: done!

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
solen0id: updated to reflect this behavior

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
...
+                        if (patternIsNegated==patternMatchesHost){
+                            return false;
+                        }
+                    }
+                }
+            }
+            return true;
+        }
+
+        private boolean patternMatchesHost(String pattern, ServerSession session){
```
solen0id: Great suggestion, thanks!

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
...
+            SocketAddress remote = session.getClientAddress();
+            if (remote instanceof InetSocketAddress) {
+                String remoteAddress = ((InetSocketAddress) remote).getAddress().getHostAddress();
+                String remoteName = ((InetSocketAddress) remote).getHostName();
```
solen0id: good point!

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
...
+                Pattern p = Pattern.compile(convertPatternToRegex(pattern));
```
solen0id: done

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
...
+                Matcher addressMatcher = p.matcher(remoteAddress);
```
solen0id: updated `addressInCidrRange` as suggested to also handle valid IP Addresses.

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
solen0id: done!

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +327,66 @@ public boolean authenticate(String userName, PublicKey key,
...
+                try {
+                    return subnet.containsHost(remoteAddress);
+                } catch (UnknownHostException e) {
+                    return false;
+                }
+            }
+            return false;
+        }
+
+        private String convertPatternToRegex(String pattern){
```
solen0id: :+1:

solen0id:
implemented the requested changes, let me know what you think

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +331,71 @@ public boolean authenticate(String userName, PublicKey key,
...
+                        Pattern p = Glob.parseGlobToPattern(pattern);
+                        Matcher addressMatcher = p.matcher(remoteAddress);
+                        Matcher nameMatcher = p.matcher(remoteName);
+                        patternMatches = (addressMatcher.matches() || nameMatcher.matches());
+                    } else {
+                        patternMatches = pattern.equals(remoteName);
+                    }
+                }
+            }
+            if (patternIsNegated && patternMatches) {
```
paulmillar: Although the code is OK, as-is, I think it could be simplified by removing the `patternMatches` variable and adding a variable that holds the response to match; something like:

    Outcome matchOutcome = patternisNegated ? Outcome.DENY : Outcome.ALLOW;

Then, the code can simply `return matchOutcome;` when there's a match.

Technically, it would also be possible to update the patch to eliminate the patternIsNegated variable, too.  I'm not sure if that improves or detracts from the code's readability.

In any case, have a play with these ideas and see if you like them.

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
paulmillar: I don't think there's any hard-and-fast rules here -- certainly, having the class doesn't prevent the patch from being accepted.

Just to describe my thinking: normally, a class is present for a reason -- it makes working with the code easier by provides encapsulation (hiding state), abstraction (through polymorphism), or promotes code reuse.  None of these apply here.

Classes also exist to split off functionality -- a class that tries to do too much is then unwieldy: difficult to work with.  So, one could argue that grouping the methods together into class is about trying to keep Ssh2Admin class small.  However, that argument kinda falls flat, because `RemoteHostValidator` is an inner class, so Ssh2Admin is actually slightly larger that if there were no RemoteHostValidator class.

On the flip side, there is a cost to having the class.  Including `RemoteHostValidator` actually makes the code a little harder to read.  Rather than directly invoking a method to test whether the login is acceptable (as a filter in the stream processing), a "helper" method is called (`remoteHostIsPermitted`) that instantiates a class (without any state) and calls the method.  Without this class, the `isValidHost` method could be called directly from the stream.

So, it's ultimately about balance: does the benefits of having the class outweigh the cost?

In the end, it's your call.  I'd say currently, the costs outweigh the benefits.  ... but this is just code: we can always change it later on; for example, should we need to "configure" how host validation works; e.g., with decorator and/or strategy design patterns.

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +331,71 @@ public boolean authenticate(String userName, PublicKey key,
...
+                pattern = pattern.substring(1);
+            }
+            SocketAddress remote = session.getClientAddress();
+            if (remote instanceof InetSocketAddress) {
+                String remoteAddress = ((InetSocketAddress) remote).getAddress().getHostAddress();
+                try {
+                    patternMatches = addressInCidrRange(pattern, remoteAddress);
+                } catch (IllegalArgumentException e) {
+                    String remoteName = ((InetSocketAddress) remote).getHostName();
+                    if (Glob.isGlob(pattern)) {
```
paulmillar: I think the if-else clause is unnecessary: you can write this as just:

    Glob glob = new Glob(pattern);
    patternMatches = glob.matches(remoteName) || glob.matches(remoteAddress);

As is, there's a slight performance penalty if the pattern doesn't contain any wildcards, since the address would be tested twice, and there's (probably) some overhead in computing a regex state-machine instead of simply testing string equality.  I suspect that both cases don't add too much just now, and the latter should be fixed in Glob, anyway.

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +331,71 @@ public boolean authenticate(String userName, PublicKey key,
...
 
+    private static class RemoteHostValidator {
+
+        private enum Outcome {
+            ALLOW, DENY, DEFER
+        }
+
+        private boolean isValidHost(String line, ServerSession session) {
+            for (String linePart : Splitter.on(" ").trimResults().omitEmptyStrings().split(line)) {
+                if (linePart.startsWith("from=") && linePart.endsWith("\"")) {
```
paulmillar: you probably want `startsWith("from=\"")` to ensure there's a quote character after the equals

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +331,71 @@ public boolean authenticate(String userName, PublicKey key,
...
+                    Set<Outcome> outcomes = EnumSet.noneOf(Outcome.class);
```
paulmillar: The code is fine as-is, but you could also get by with just a boolean to record whether Outcome.ALLOW has been returned, provided a Outcome.DENY short-circuits the tests and immediate returns false.

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -318,6 +331,71 @@ public boolean authenticate(String userName, PublicKey key,
...
+                try {
+                    Subnet subnet = Subnet.create(cidrAddress);
+                    return subnet.containsHost(remoteAddress);
+                } catch (UnknownHostException e) {
+                    return false;
+                }
+            }
+
+        private String strip(String linePart) {
+            return linePart.substring(6, linePart.length()-1);
```
paulmillar: Not blocking the patch, but given this method is so small, why not include it the one place it gets called?
Something like:

    if (linePart.startsWith("from=") && linePart.endsWith("\"")) {
        String from = linePart.substring(6, linePart.length()-1);

solen0id:
Thanks for the Feedback @paulmillar , much appreciated :)
I took your advice to heart and implemented the changes, ready for round 3!

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -316,6 +325,53 @@ public boolean authenticate(String userName, PublicKey key,
...
+                    }
+                }
+            }
+            return Outcome.DEFER;
+        }
+
+        private boolean addressInCidrRange(String cidrAddress, String remoteAddress) throws IllegalArgumentException {
+                try {
+                    Subnet subnet = Subnet.create(cidrAddress);
+                    return subnet.containsHost(remoteAddress);
```
paulmillar: Actually, I think you want the `Subnet#contains` method.  This takes an InetAddress (rather than a String).

This has a few benefits: using this method we avoid converting the address into a String form, only to parse it back again.
The method also never throws UnknownHostException :-)

The two lines could be combined into just:

    return Subnet.create(cidrAddress).contains(remoteAddress);

Also, given this method is now just a single line and used just once, you can probably refactor the code to include this line in patternMatchesHost directly:

    if (Subnet.create(cidrAddress).contains(remote.getAddress())) {
        return matchOutcome;
    }

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -316,6 +325,53 @@ public boolean authenticate(String userName, PublicKey key,
...
+
+        private Outcome patternMatchesHost(String pattern, ServerSession session) {
+            boolean patternIsNegated = pattern.startsWith("!");
+            Outcome matchOutcome = patternIsNegated ? Outcome.DENY : Outcome.ALLOW;
+            if (patternIsNegated) {
+                pattern = pattern.substring(1);
+            }
+            SocketAddress remote = session.getClientAddress();
+            if (remote instanceof InetSocketAddress) {
+                String remoteAddress = ((InetSocketAddress) remote).getAddress().getHostAddress();
```
paulmillar: Since it is used in two places (and those lines are kinda long), you should probably factor out this cast; e.g.,

    SocketAddress rawRemote = session.getClientAddress();
    if (rawRemote instanceof InetSocketAddress) {
        InetSocketAddress remote = (InetSocketAddress) rawRemote;
        // ...
    }

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -282,7 +292,7 @@ private PublicKey toPublicKey(String s) {
             }
             return null;
         }
-
+
```
paulmillar: Eek: our coding style doesn't allow trailing white-space in our code ;-)

solen0id: whoopsie

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
solen0id: Done!

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
solen0id: nice catch!

solen0id:
All done!
If we keep going at this pace there won't be any more code to review left soon :D

solen0id:
Yeah I aggree!
Thanks for your good advice :)

PR #3433, opened 2017-08-09
acl: refactoring slf4j logging messages

marisanest:
Motivation: 
With normal string concatenations in log-messages strings are always build, regardless if log-level is activated or not. with parameterized log-messages the strings only become build, when the log-level is activated;

Modification: 

Result: 
Gained efficiency

Signed-off-by: 
marisanest <marisanest@mailbox.org>


kofemann:
thanks!

PR #3435, opened 2017-08-10
vehicles: Updated JobInfo

alenaschemmert:
Motivation:

The new time API from Java8 is more modern and threadsafe then in further versions.

Modification:

Updated SimpleDateFormat to DateTimeFormatter and used it with LocalDateTime in toString(). Also removed the synchronized block, because DateTimeFormatter is thread safe.

Added tests, to show threadsafety and equality of the old and new toString().

Result:

The class is now more up-to-date

Target: master
Require-notes: no
Require-book: no

### DIFF
--- a/modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
+++ b/modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
@@ -5,6 +5,8 @@ package diskCacheV111.vehicles;
 import java.io.IOException;
 import java.io.Serializable;
 import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -53,8 +55,8 @@ public class JobInfo implements Serializable {
         StringBuilder sb = new StringBuilder();
         sb.append(_jobId).append(';');
         sb.append(_client).append(':').append(_clientId);
-        sb.append(';').append(__formatter.format(Instant.ofEpochMilli(_submitTime)));
-        sb.append(';').append(__formatter.format(Instant.ofEpochMilli(_startTime)));
+        sb.append(';').append(LocalDateTime.ofInstant(Instant.ofEpochMilli(_submitTime), ZoneId.systemDefault()).format(__formatter));
+        sb.append(';').append(LocalDateTime.ofInstant(Instant.ofEpochMilli(_startTime), ZoneId.systemDefault()).format(__formatter));
         sb.append(';').append(_status).append(';');
         return sb.toString();
     }
--- a/modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
+++ b/modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
@@ -55,8 +55,8 @@ public class JobInfo implements Serializable {
         StringBuilder sb = new StringBuilder();
         sb.append(_jobId).append(';');
         sb.append(_client).append(':').append(_clientId);
-        sb.append(';').append(LocalDateTime.ofInstant(Instant.ofEpochMilli(_submitTime), ZoneId.systemDefault()).format(__formatter));
         sb.append(';').append(LocalDateTime.ofInstant(Instant.ofEpochMilli(_startTime), ZoneId.systemDefault()).format(__formatter));
+        sb.append(';').append(LocalDateTime.ofInstant(Instant.ofEpochMilli(_submitTime), ZoneId.systemDefault()).format(__formatter));
         sb.append(';').append(_status).append(';');
         return sb.toString();
     }
--- a/modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
+++ b/modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
@@ -13,9 +13,9 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 public class JobInfo implements Serializable {
 
-     private static final DateTimeFormatter __formatter =
+    private static final DateTimeFormatter __formatter =
             DateTimeFormatter.ofPattern("MM/dd-HH:mm:ss");
-     private static final long serialVersionUID = 5209798222006083955L;
+    private static final long serialVersionUID = 5209798222006083955L;
 
     private final String _client;
     private final long _clientId;
@@ -24,7 +24,7 @@ public class JobInfo implements Serializable {
     private String _status;
     private final long _jobId;
 
-     public JobInfo(long submitTime, long startTime, String status, int id, String clientName, long clientId) {
+    public JobInfo(long submitTime, long startTime, String status, int id, String clientName, long clientId) {
         _submitTime = submitTime;
         _startTime = startTime;
         _status = checkNotNull(status);
--- a/modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
+++ b/modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
@@ -4,55 +4,75 @@ package diskCacheV111.vehicles;
 
 import java.io.IOException;
 import java.io.Serializable;
-import java.text.SimpleDateFormat;
-import java.util.Date;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
 public class JobInfo implements Serializable {
 
-    private static final SimpleDateFormat __format =
-            new SimpleDateFormat( "MM/dd-HH:mm:ss" ) ;
+    private final DateTimeFormatter __formatter = DateTimeFormatter.ofPattern("MM/dd-HH:mm:ss");
+
     private static final long serialVersionUID = 5209798222006083955L;
 
-   private final String _client;
-   private final long   _clientId;
-   private final long   _submitTime;
-   private final long   _startTime;
-   private String _status;
-   private final long   _jobId;
+    private final String _client;
+    private final long _clientId;
+    private final long _submitTime;
+    private final long _startTime;
+    private String _status;
+    private final long _jobId;
 
-    public JobInfo(long submitTime, long startTime, String status, int id, String clientName , long clientId ){
-      _submitTime = submitTime ;
-      _startTime  = startTime;
-      _status     = checkNotNull(status);
-      _jobId      = id;
-      _client   = clientName ;
-      _clientId = clientId ;
-   }
-   public String getClientName(){ return _client ; }
-   public long   getClientId(){ return _clientId ; }
-   public long   getStartTime(){ return _startTime ; }
-   public long   getSubmitTime(){ return _submitTime ; }
-   public String getStatus(){ return _status  ;}
-   public long   getJobId(){ return _jobId ; }
 
-   public String toString(){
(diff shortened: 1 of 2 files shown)


modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
```diff
@@ -4,55 +4,75 @@
...
-      sb.append(_jobId).append(';');
-      sb.append(_client).append(':').append(_clientId) ;
-      synchronized (__format) {
-          sb.append(';').append(__format.format(new Date(_startTime))).
-                  append(';').append(__format.format(new Date(_submitTime))).
-                  append(';').append(_status).append(';') ;
-      }
-      return sb.toString();
-   }
+    private final String _client;
```
paulmillar: While the reformatting is reasonable, please try to avoid doing it in this patch. Reformatting the code while also changing what the code does makes it harder to see what has actually changed.

modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
```diff
@@ -4,55 +4,75 @@
...
+
+    public long getJobId() {
+        return _jobId;
+    }
+
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append(_jobId).append(';');
+        sb.append(_client).append(':').append(_clientId);
+        sb.append(';').append(LocalDateTime.ofInstant(Instant.ofEpochMilli(_submitTime), ZoneId.systemDefault()).format(__formatter));
```
paulmillar: Would something like `__formatter.format(Instant.ofEpochMilli(_submitTime))` also work?  That seems less convoluted.

modules/dcache-vehicles/src/test/java/diskCacheV111/vehicles/JobInfoTests.java
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.Test;
+
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.concurrent.*;
+
+/**
+ * Created by alenaschemmert on 03.08.17.
```
paulmillar: Don't worry, git remembers when this class was created and by whom, so you don't need a comment like this :-)

modules/dcache-vehicles/src/test/java/diskCacheV111/vehicles/JobInfoTests.java
```diff
@@ -0,0 +1,74 @@
...
+
+public class JobInfoTests {
+
+    private JobInfo jobInfo;
+    private OldJobInfo oldJobInfo;
+    private String ref;
+
+    @Before
+    public void setUp() {
+        jobInfo = new JobInfo(System.currentTimeMillis(), System.currentTimeMillis() + 1000, "working", 1, "waldo", 2407);
```
paulmillar: Ideally, unit-tests should be repeatable and always give the same output.  This is not always possible, but is certainly something to strive for.

By using `System.currentTimeMillis`, the patch introduces an element of randomness.  For example, suppose there is some subtle bug that only shows up with the last digit is a 0 the test would then fail (roughly) 1 time in 10.

Wherever possible, it's better to use constant values.  When a bug is discovered (e.g., the last digit is a 0 bug) an additional test is adding that targets exactly the conditions that trigger the bug.  That way testing is precise and predictable.

modules/dcache-vehicles/src/test/java/diskCacheV111/vehicles/JobInfoTests.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Test
+    public void toStringShouldBeThreadSafe() {
+        Callable<String> task = jobInfo::toString;
+        ExecutorService exec = Executors.newFixedThreadPool(8);
+        List<Future<String>> results = new ArrayList<>();
+
+        for (int i = 0; i < 8; i++) {
+            results.add(exec.submit(task));
```
paulmillar: minor comment: since `task` is only used here, you could reasonably replace it with the actual method reference -- `jobInfo::toString`

modules/dcache-vehicles/src/test/java/diskCacheV111/vehicles/JobInfoTests.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+        exec.shutdown();
+
+        for (Future<String> result : results) {
+            try {
+                Assert.assertEquals(ref, result.get());
```
paulmillar: It's common to statically import Assert methods; e.g.,

    import static org.junit.Assert.*;

modules/dcache-vehicles/src/test/java/diskCacheV111/vehicles/JobInfoTests.java
```diff
@@ -0,0 +1,74 @@
...
+            } catch (InterruptedException | ExecutionException e) {
+                Assert.assertFalse(true);
```
paulmillar: ... or perhaps less obtuse as `Assert.fail`

modules/dcache-vehicles/src/test/java/diskCacheV111/vehicles/JobInfoTests.java
```diff
@@ -0,0 +1,74 @@
...
+            }
+        }
+    }
+
+    @Test
+    public void OldAndNewShouldBeEqual() {
+        Assert.assertEquals(jobInfo.toString(), oldJobInfo.toString());
```
paulmillar: You've actually got this the wrong way around: the expected value is the first argument.

This is one of the motivations for the (slightly) newer `assertThat` syntax, which makes these kind of mistakes a little less likely; e.g.,

    assertThat(jobInfo.toString(), is(equalTo(oldJobInfo.toString())));

Also, you have `ref` as a field member -- even better would be to replace this with a String literal.

modules/dcache-vehicles/src/test/java/diskCacheV111/vehicles/JobInfoTests.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private static class OldJobInfo extends JobInfo {
```
paulmillar: Hmmm, this is kinda fragile -- if the arguments changes then this could cause problems.  For example, imagine the `submitTime` argument in JobInfo changes from `long` to `Instant`.  We would need to update the JobInfoTests.

Ideally, unit tests shouldn't need updating like this.

An alternative approach is for `@Test` methods to check against literal (or literal fragments).  After all, the output from `oldJobInfo.toString()` is just a String, so could be replaced by a String literal.

paulmillar:
ok to test

modules/dcache-vehicles/src/test/java/diskCacheV111/vehicles/JobInfoTests.java
paulmillar: As if to underline this point: the unit test just failed when I trigged our CI system:

https://ci.dcache.org/job/dCache-master/7613/testReport/junit/diskCacheV111.vehicles/JobInfoTests/OldAndNewShouldBeEqual/

Seemingly, the seconds "rolled over" between creating JobInfo and OldJobInfo, so the output was different (failing the test) although the code is basically sound.

modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
alenaschemmert: 👍 
tried to undo it

modules/dcache-vehicles/src/test/java/diskCacheV111/vehicles/JobInfoTests.java
alenaschemmert: I can't think of an way, to test the `toString` method without having an object of `JobInfo`. So, the test will fail if JobInfo changes.
But I deleted the class `OldJobInfo`

paulmillar: `toString()` is often designed for human consumption (e.g., interactive admin commands, log files, ...).  Writing unit-tests that are specific enough to catch problems, but general enough to allow "reasonable" changes is difficult.  There's also the very real risk that someone has written some script that parses the output.

Ultimately, we don't know what expectations our users have here, so we have to "use our best judgement".

However, to give some concrete advice here, your patch is about the output of the formatted string, so it's reasonable that your unit-test focuses on that part and ignores everything else.

Lets say you update your `@Before` method to use a fixed time.  Any value is fine, for example suppose the submit time is 1502369009000L (which is about now).  We expect that the output from `toString();` will contain this time in the correct format; i.e.,  "08/10-14:43:29".

So, the simplest test would be (using the newer `assertThat` syntax) to check whether that string appears in the output:

    assertThat(toString, containsString("08/10-14:43:29"));

If the startTime is sufficiently different (currently it is one second after submitTime), then an additional assertion can be added, looking for that string:

    assertThat(toString, containsString("08/10-14:43:30"));

Certainly, this is a very limited test, but sufficient to check that the code works as anticipated.

The assertion could be made stronger, by insisting that the output is a `;`-separated list of items, two of which must match each of the dates exactly, something like:

    List<String> toStringItems = Splitter.on(';').splitToList(toString)
    assertThat(toStringItems, contains("08/10-14:43:29"));
    assertThat(toStringItems, contains("08/10-14:43:30"));

The tests could be made stronger still by insisting in the order of these items, but I think this is roughly the right balance between being too restrictive in the unit-test and catching potential problems.

modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
paulmillar: Thanks

modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
```diff
@@ -24,36 +22,29 @@
     private String _status;
     private final long _jobId;
 
-
-    public JobInfo(long submitTime, long startTime, String status, int id, String clientName, long clientId) {
+     public JobInfo(long submitTime, long startTime, String status, int id, String clientName, long clientId) {
```
paulmillar: Did you nudge the constructor?  It looks like the patch has added a space to its indentation 😕

modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/JobInfo.java
```diff
@@ -4,54 +4,65 @@
...
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
 public class JobInfo implements Serializable {
 
-    private static final SimpleDateFormat __format =
```
paulmillar: Could you update the patch so it doesn't reformat all these lines?

modules/dcache-vehicles/src/test/java/diskCacheV111/vehicles/JobInfoTests.java
```diff
@@ -0,0 +1,50 @@
...
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.*;
+
+public class JobInfoTests {
+
+    private JobInfo jobInfo;
+
+    @Before
+    public void setUp() {
+        long time = 1502366668987L;
```
anupash: make *time* a 
```java
private static final long time = 1502366668987L;
```

jstarek:
Although this looks like the remaining issues could be solved rather easily, I'll close the PR now due to inactivity. Feel free to re-open once the issues are resolved.

PR #3436, opened 2017-08-10
nfsv41door: add filter method for showtransfers command

Yannick90:
Motivation:
When inspecting active transfers a "show transfers" command in the dCache's admin interface is used. On a busy system, the output of the command may return more lines than can be visually inspected.

Modification:
added another "showtransfers" command with possibility to filter by client, pnfsid and pool.

Result:
Possibility to filter the inspection of "show transfers"

Signed-off-by: Yannick Vahldieck <yannick.vahldieck@googlemail.com>


modules/dcache-nfs/src/main/java/org/dcache/chimera/nfsv41/door/NFSv41Door.java
```diff
@@ -858,6 +858,34 @@ public String call() throws IOException {
         }
     }
 
+
+    @Command(name = "show transfers filtered", hint = "show active transfers filtered",
+            description = "Show active transfers excluding proxy-io filtered by pool, client or pnfsid.")
+    public class ShowTransfersFilteredCmd implements Callable<String> {
+
+        @Argument(required = false, metaVar = "pool")
```
paulmillar: Although arguments can be made optional, there's an inherent problem -- arguments are ordered, which means that an admin cannot filter on pnfsid without also filtering on client and pool.

One way around this (what we do elsewhere) is to use `@Option` annotation instead.  This gives the admin the opportunity to "switch on" the desired behaviour; for example:

    show transfers -pool=my_interesting_pool -pnfsid=....

modules/dcache-nfs/src/main/java/org/dcache/chimera/nfsv41/door/NFSv41Door.java
```diff
@@ -858,6 +858,34 @@ public String call() throws IOException {
         }
     }
 
+
+    @Command(name = "show transfers filtered", hint = "show active transfers filtered",
```
paulmillar: I'm wondering whether it would be better to update the existing command rather than adding a new command: the new command, when run without any filters, provides the same output as the `show transfers` command currently.

modules/dcache-nfs/src/main/java/org/dcache/chimera/nfsv41/door/NFSv41Door.java
```diff
@@ -858,6 +858,34 @@ public String call() throws IOException {
...
+            description = "Show active transfers excluding proxy-io filtered by pool, client or pnfsid.")
+    public class ShowTransfersFilteredCmd implements Callable<String> {
+
+        @Argument(required = false, metaVar = "pool")
+        String pool;
```
paulmillar: While exact matching Strings may be OK under some circumstances, client identifiers and often pool names are structured.  Therefore, it could be useful to support selecting a family of pools or a family of clients, rather than one in particular.

This is pretty easy to do: dCache already has a class to support globbing patterns (Glob).  The change would look something like:

    @Option(name="pool")
    Glob pool;

    // ...
        .filter( d -> pool == null ? true : pool.matches(d.getPool()))

paulmillar:
ok to test

modules/dcache-nfs/src/main/java/org/dcache/chimera/nfsv41/door/NFSv41Door.java
kofemann: Jup. I agree with Paul.

modules/dcache-nfs/src/main/java/org/dcache/chimera/nfsv41/door/NFSv41Door.java
```diff
@@ -848,11 +843,23 @@ public String call() throws IOException {
             description = "Show active transfers excluding proxy-io.")
     public class ShowTransfersCmd implements Callable<String> {
 
+        @Option(name = "pool")
```
paulmillar: The `@Option` annotation has a `usage`  field.  Although this is optional, the default value (an empty string) means that these options will not be include in the online documentation, which would be a shame.

Perhaps you could add a short `usage` field for each option?  It doesn't have to be long, one sentence is enough, but it would be nice if the message mentioned that glob pattern matching is supported.

Here's an example on how to see the online help:
```
paul@celebrimbor:~/git/dCache (master)$ ssh -p 22224 admin@localhost
dCache (3.2.0-SNAPSHOT)
Type "\?" for help.

[celebrimbor] (local) admin > \c NFS-celebrimbor 
[celebrimbor] (NFS-celebrimbor@dCacheDomain) admin > \h show transfers
NAME
       show transfers -- show active transfers

SYNOPSIS
       show transfers 

DESCRIPTION
       Show active transfers excluding proxy-io.


[celebrimbor] (NFS-celebrimbor@dCacheDomain) admin > 
```

modules/dcache-nfs/src/main/java/org/dcache/chimera/nfsv41/door/NFSv41Door.java
```diff
@@ -4,6 +4,7 @@
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.ListenableFuture;
+import org.dcache.util.*;
```
paulmillar: Sorry, one last thing: our coding-style doesn't allow for stars in non-static imports.

Having said that, the imports are already non-conforming since they are not sorted *sigh*.

I suggest you update the patch so that it doesn't convert the imports into a star-import.  Of course the patch still need to add the Glob import.  Just put that where the majority of org.dcache.util imports are location; i.e., after `import org.dcache.util.FireAndForgetTask;`

kofemann:
👍

kofemann:
Thanks!

Issue #2529, opened 2016-06-08
OOM exception in ChimeraCleaner


kofemann:
ChimeraCleaner uses JdbcTemplate#queryForList() which builds a list out of result of the query. For a big results, the constructed list may not fit into memory and produce OOM exception, like this:

```
pool-4-thread-1
  at java.lang.OutOfMemoryError.<init>()V (OutOfMemoryError.java:48)
  at org.postgresql.core.PGStream.ReceiveTupleV3()[[B (PGStream.java:375)
  at org.postgresql.core.v3.QueryExecutorImpl.processResults(Lorg/postgresql/core/ResultHandler;I)V (QueryExecutorImpl.java:1879)
  at org.postgresql.core.v3.QueryExecutorImpl.execute(Lorg/postgresql/core/Query;Lorg/postgresql/core/ParameterList;Lorg/postgresql/core/ResultHandler;III)V (QueryExecutorImpl.java:173)
  at org.postgresql.jdbc.PgStatement.execute(Lorg/postgresql/core/Query;Lorg/postgresql/core/ParameterList;I)V (PgStatement.java:622)
  at org.postgresql.jdbc.PgStatement.executeWithFlags(Ljava/lang/String;I)Z (PgStatement.java:458)
  at org.postgresql.jdbc.PgStatement.executeQuery(Ljava/lang/String;)Ljava/sql/ResultSet; (PgStatement.java:374)
  at com.zaxxer.hikari.pool.ProxyStatement.executeQuery(Ljava/lang/String;)Ljava/sql/ResultSet; (ProxyStatement.java:111)
  at com.zaxxer.hikari.pool.HikariProxyStatement.executeQuery(Ljava/lang/String;)Ljava/sql/ResultSet; (Unknown Source)
  at org.springframework.jdbc.core.JdbcTemplate$1QueryStatementCallback.doInStatement(Ljava/sql/Statement;)Ljava/lang/Object; (JdbcTemplate.java:454)
  at org.springframework.jdbc.core.JdbcTemplate.execute(Lorg/springframework/jdbc/core/StatementCallback;)Ljava/lang/Object; (JdbcTemplate.java:404)
  at org.springframework.jdbc.core.JdbcTemplate.query(Ljava/lang/String;Lorg/springframework/jdbc/core/ResultSetExtractor;)Ljava/lang/Object; (JdbcTemplate.java:470)
[...]

PR #3437, opened 2017-08-10
Chimera: fix OOM Exception when too many messages are queued

rwenz3l:
Motivation:
To prevent future Out of Memory Exceptions in Chimera Cleaner when
processing too many messages

Modification:
Set a LIMIT in the SQL Query and loop until all messages are processed.

Result:
Messages are processed in chunks and prevent the method from crashing.

Ticket: #2529

Target: master

Signed-off-by: Robin Wenzel <info@robin-wenzel.de>

### DIFF
--- a/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
+++ b/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
@@ -76,7 +76,7 @@ public class ChimeraCleaner extends AbstractCell implements Runnable
         LoggerFactory.getLogger(ChimeraCleaner.class);
 
     // ToDo: Spring Implementation for queryLimit ( used in runNotification() )
-    private final int queryLimit = 5000;
+    private final static int QUERYLIMIT = 5000;
 
     @Option(
         name="refresh",
@@ -470,9 +470,9 @@ public class ChimeraCleaner extends AbstractCell implements Runnable
 
         List<String> queryList = new ArrayList<String>();
         do {
-            queryList = _db.queryForList(QUERY, String.class, queryLimit);
+            queryList = _db.queryForList(QUERY, String.class, QUERYLIMIT);
 
-            if(queryList.size() == 0) {
+            if( queryList.isEmpty()) {
                 break;
             }
 
@@ -484,7 +484,7 @@ public class ChimeraCleaner extends AbstractCell implements Runnable
                     _log.warn(e.getCause().getMessage());
                 }
             }
-        } while (queryList.size() >= queryLimit);
+        } while (queryList.size() >= QUERYLIMIT);
     }
 
     private ListenableFuture<List<PnfsDeleteEntryNotificationMessage>> sendDeleteNotifications(PnfsId pnfsId)
--- a/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
+++ b/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
@@ -470,7 +470,6 @@ public class ChimeraCleaner extends AbstractCell implements Runnable
 
         List<String> queryFail = new ArrayList<String>();
         do {
-            }
             List<String> queryList = _db.queryForList(QUERY, String.class, QUERYLIMIT);
 
             for (String id : queryList) {
--- a/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
+++ b/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
@@ -468,11 +468,11 @@ public class ChimeraCleaner extends AbstractCell implements Runnable
                 "SELECT ipnfsid FROM t_locationinfo_trash t1 " +
                 "WHERE itype=2 AND NOT EXISTS (SELECT 1 FROM t_locationinfo_trash t2 WHERE t2.ipnfsid=t1.ipnfsid AND t2.itype <> 2) LIMIT=?";
 
-        List<String> queryFail = new ArrayList<String>();
+        HashSet<String> queryFail = new HashSet<String>();
         do {
-            List<String> queryList = _db.queryForList(QUERY, String.class, QUERYLIMIT);
+            List<String> results = _db.queryForList(QUERY, String.class, QUERYLIMIT);
 
-            for (String id : queryList) {
+            for (String id : results) {
                 try {
                     sendDeleteNotifications(new PnfsId(id)).get();
                     _db.update("DELETE FROM t_locationinfo_trash WHERE ipnfsid=? AND itype=2", id);
@@ -484,10 +484,11 @@ public class ChimeraCleaner extends AbstractCell implements Runnable
                 }
             }
             if(queryFail.size() >= QUERYLIMIT){
-                _log.warn("Error: Too many unresolved IDs to finish the loop.");
+                _log.error("Error: There have been too many failures while attempting to collect IDs from the t_locationinfo_trash + " +
+                        "and sending them to the deleteNotifications Service. Further deletion is aborted.");
                 break;
             }
-        } while (queryList.size() >= QUERYLIMIT);
+        } while (results.size() >= QUERYLIMIT);
     }
 
     private ListenableFuture<List<PnfsDeleteEntryNotificationMessage>> sendDeleteNotifications(PnfsId pnfsId)
--- a/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
+++ b/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
@@ -477,9 +477,7 @@ public class ChimeraCleaner extends AbstractCell implements Runnable
                     sendDeleteNotifications(new PnfsId(id)).get();
                     _db.update("DELETE FROM t_locationinfo_trash WHERE ipnfsid=? AND itype=2", id);
                 } catch (ExecutionException e) {
-                    if(!queryFail.contains(id)) {
                         queryFail.add(id);
-                    }
                     _log.warn(e.getCause().getMessage());
                 }
             }
--- a/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
+++ b/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
@@ -75,6 +75,9 @@ public class ChimeraCleaner extends AbstractCell implements Runnable
     private static final Logger _log =
         LoggerFactory.getLogger(ChimeraCleaner.class);
 
+    // ToDo: Spring Implementation for queryLimit ( used in runNotification() )
+    private final int queryLimit = 5000;
+
     @Option(
         name="refresh",
         description="Refresh interval",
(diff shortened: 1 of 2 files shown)


modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
```diff
@@ -75,6 +75,9 @@
     private static final Logger _log =
         LoggerFactory.getLogger(ChimeraCleaner.class);
 
+    // ToDo: Spring Implementation for queryLimit ( used in runNotification() )
+    private final int queryLimit = 5000;
```
kofemann: As this is a constant for now, please make is **private final static**. And constants should be UPPER case.

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
```diff
@@ -454,20 +457,34 @@ public void messageArrived(PoolManagerPoolUpMessage poolUpMessage)
...
-                _db.update("DELETE FROM t_locationinfo_trash WHERE ipnfsid=? AND itype=2", id);
-            } catch (ExecutionException e) {
-                _log.warn(e.getCause().getMessage());
+                "WHERE itype=2 AND NOT EXISTS (SELECT 1 FROM t_locationinfo_trash t2 WHERE t2.ipnfsid=t1.ipnfsid AND t2.itype <> 2) LIMIT=?";
+
+        List<String> queryList = new ArrayList<String>();
+        do {
+            queryList = _db.queryForList(QUERY, String.class, queryLimit);
+
+            if(queryList.size() == 0) {
```
kofemann: conciser to use **isEmpty**

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
```diff
@@ -454,20 +457,34 @@ public void messageArrived(PoolManagerPoolUpMessage poolUpMessage)
...
+
+            for (String id : queryList) {
+                try {
+                    sendDeleteNotifications(new PnfsId(id)).get();
+                    _db.update("DELETE FROM t_locationinfo_trash WHERE ipnfsid=? AND itype=2", id);
+                } catch (ExecutionException e) {
+                    _log.warn(e.getCause().getMessage());
+                }
+            }
+        } while (queryList.size() >= queryLimit);
```
kofemann: technically, you don't need this check as look will break as when query returns an empty list.

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
```diff
@@ -75,6 +75,9 @@
     private static final Logger _log =
         LoggerFactory.getLogger(ChimeraCleaner.class);
 
+    // ToDo: Spring Implementation for queryLimit ( used in runNotification() )
```
alrossi: Is this going to be added with another commit, or is it just an annotation for our future reference?

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
alrossi: I was going to mention same thing.  But it does save one trip to the database.

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
rwenz3l: It's considered an annotation for future reference as I talked with @kofemann about the proper implementation of the limit. Idealy it would be in a properties file, but that seems more difficult to implement for now and we agreed on defining the QueryLimit as a constant variable in the class.

kofemann: As cleaner is not Springified yet, let ignore it for the moment

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
kofemann: ok

alrossi: Actually, there is another problem.   What is the behavior when an ExecutionException is caught?  Should there be a retry?  Could you get yourself into an infinite loop in that case?   The check would stop you from reprocessing the failed entries.

alrossi: Actually, I take that back.  Suppose ... perhaps unlikely, but possible ... you get queryLimit number of failures.  That leaves queryLimit number of entries in the trash table.  If the underlying problem is not transient, you will go into a loop trying to reprocess them and catching the exception.

rwenz3l: @alrossi yeah we thought about that as well.. it's a bit of an issue, yet unlikely as you said.
I initially wanted to put that into a Transaction, but the whole process is inside a transaction anyway. The Query would run indefinitely. The queryLimit Number of results (5000 currently) are processed one by one and throws a log message. If in the next loop they can't be processed, it just goes on and on until it can or it spams the log forever.

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
rwenz3l: changed in new commit.

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
rwenz3l: changed in new commit.

alrossi:
@kofemann if you feel the potential looping behavior need not be addressed here, then I'm OK with the changes.

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
```diff
@@ -454,20 +457,34 @@ public void messageArrived(PoolManagerPoolUpMessage poolUpMessage)
...
-                _db.update("DELETE FROM t_locationinfo_trash WHERE ipnfsid=? AND itype=2", id);
-            } catch (ExecutionException e) {
-                _log.warn(e.getCause().getMessage());
+                "WHERE itype=2 AND NOT EXISTS (SELECT 1 FROM t_locationinfo_trash t2 WHERE t2.ipnfsid=t1.ipnfsid AND t2.itype <> 2) LIMIT=?";
+
+        List<String> queryList = new ArrayList<String>();
+        do {
+            queryList = _db.queryForList(QUERY, String.class, QUERYLIMIT);
+
+            if( queryList.isEmpty()) {
```
paulmillar: This seems unnecessary: if queryList is empty then the for-each loop does nothing and while-loop termination condition is reached.

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
```diff
@@ -454,20 +457,34 @@ public void messageArrived(PoolManagerPoolUpMessage poolUpMessage)
...
-                "WHERE itype=2 AND NOT EXISTS (SELECT 1 FROM t_locationinfo_trash t2 WHERE t2.ipnfsid=t1.ipnfsid AND t2.itype <> 2)";
-        for (String id : _db.queryForList(QUERY, String.class)) {
-            try {
-                sendDeleteNotifications(new PnfsId(id)).get();
-                _db.update("DELETE FROM t_locationinfo_trash WHERE ipnfsid=? AND itype=2", id);
-            } catch (ExecutionException e) {
-                _log.warn(e.getCause().getMessage());
+                "WHERE itype=2 AND NOT EXISTS (SELECT 1 FROM t_locationinfo_trash t2 WHERE t2.ipnfsid=t1.ipnfsid AND t2.itype <> 2) LIMIT=?";
+
+        List<String> queryList = new ArrayList<String>();
```
paulmillar: The patch doesn't need to create this object -- Java should be smart enough to realise that the first line in the do-loop will always assign `queryList` a value ahead of the `while` clause.

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
paulmillar: Perhaps it would be better to remember which PNFSIDs failed and not retry them within the while loop.

The while-loop's exit criteria would then need to be updated  -- requiring (something like) at least one file was removed from the db after successfully notifying the corresponding services; i.e., some akin to "progress was made".

alrossi: @paulmillar that was similar to my thinking;  I know the code probably will not run into such a problem in the practical sense, but technically it is incorrect as it stands ...

paulmillar: I guess the real danger here is that if one of the receiving services goes down, all messages sent to that service will fail.

If the is QUERYLIMIT or more entries in the database, the while-loop will not terminate until the service comes back up again.

This has three bad effects:
  1. cleaner will send messages as fast as it can, hammering both the messaging infrastructure and the targeted services.
  2. cleaner will spam the log file will entries; again, just as fast as it can.
  3. no further cleaning of deleted pool files will take place.

rwenz3l:
Implemented given advice and attempted to circumvent the failed query-ids issue by keeping a list of failed and breaking when said list is greater or equal to QUERYLIMIT.

anupash:
@rwenz3l  can you change the module name in the commit message from **ChimeraCleaner** to **chimera**

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
```diff
@@ -468,22 +468,26 @@ private void runNotification()
...
-                break;
             }
+            List<String> queryList = _db.queryForList(QUERY, String.class, QUERYLIMIT);
 
             for (String id : queryList) {
                 try {
                     sendDeleteNotifications(new PnfsId(id)).get();
                     _db.update("DELETE FROM t_locationinfo_trash WHERE ipnfsid=? AND itype=2", id);
                 } catch (ExecutionException e) {
+                    if(!queryFail.contains(id)) {
```
anupash: Use a HashSet instead of a list. Checking contains on a large list of strings is potentially very slow.

Using List is O ( num-of-strings-in-the-list x size-of-each-string)
Using Set is O ( num-of-strings-in-the-list x  constant-time-for-hashing)

With this you remove the double checking of queryFail.contains

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
```diff
@@ -468,22 +468,26 @@ private void runNotification()
...
 
-        List<String> queryList = new ArrayList<String>();
+        List<String> queryFail = new ArrayList<String>();
         do {
-            queryList = _db.queryForList(QUERY, String.class, QUERYLIMIT);
-
-            if( queryList.isEmpty()) {
-                break;
             }
+            List<String> queryList = _db.queryForList(QUERY, String.class, QUERYLIMIT);
```
anupash: Suggest you to use a better suited variable name like "resultList" or "results"

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
```diff
@@ -468,22 +468,26 @@ private void runNotification()
...
                     _db.update("DELETE FROM t_locationinfo_trash WHERE ipnfsid=? AND itype=2", id);
                 } catch (ExecutionException e) {
+                    if(!queryFail.contains(id)) {
+                        queryFail.add(id);
+                    }
                     _log.warn(e.getCause().getMessage());
                 }
             }
+            if(queryFail.size() >= QUERYLIMIT){
+                _log.warn("Error: Too many unresolved IDs to finish the loop.");
```
anupash: Use of log level warning and error in the start of log message is contradictory. 

Also improve the log message,   "IDs to finish the loop." is not a very useful error message.

rwenz3l:
Made further adjustments according to @anupash recommendations.

modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
```diff
@@ -454,20 +457,38 @@ public void messageArrived(PoolManagerPoolUpMessage poolUpMessage)
...
+        HashSet<String> queryFail = new HashSet<String>();
+        do {
+            List<String> results = _db.queryForList(QUERY, String.class, QUERYLIMIT);
+
+            for (String id : results) {
+                try {
+                    sendDeleteNotifications(new PnfsId(id)).get();
+                    _db.update("DELETE FROM t_locationinfo_trash WHERE ipnfsid=? AND itype=2", id);
+                } catch (ExecutionException e) {
+                    if(!queryFail.contains(id)) {
```
anupash: With the use of HashSet, you do not need to double check for contains

check javadoc, https://docs.oracle.com/javase/7/docs/api/java/util/HashSet.html

"Adds the specified element to this set if it is not already present."

kofemann:
ok to test

jstarek:
Closing due to long inactivity, although most issues seem to have been ironed out. If anyone wants to continue work on this PR, please feel free to resubmit after updating.

PR #3439, opened 2017-08-10
dcache: add client ip filter option to mover ls command

FritzHeiden:
Motivation:
When listing movers only movers regarding a specific client may be listed.

Modification:
Added option for mover ls command providing the client ip or host to be filtered out.
Added filter to the queue stream which is responsible for building the output list.

Acted by: Fritz Heiden
Signed-off-by: Fritz Heiden <fritz.heiden@mailbox.org>

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/pool/classic/IoQueueManager.java
+++ b/modules/dcache/src/main/java/org/dcache/pool/classic/IoQueueManager.java
@@ -1,34 +1,9 @@
 package org.dcache.pool.classic;
 
-import diskCacheV111.util.PnfsId;
-import diskCacheV111.vehicles.IpProtocolInfo;
-import diskCacheV111.vehicles.ProtocolInfo;
-import org.dcache.pool.movers.Mover;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import javax.annotation.Nonnull;
-
-import java.io.PrintWriter;
-import java.io.Serializable;
-import java.net.InetAddress;
-import java.net.UnknownHostException;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Comparator;
-import java.util.Optional;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Function;
-import java.util.function.Predicate;
-
 import diskCacheV111.util.CacheException;
 import diskCacheV111.vehicles.IoJobInfo;
-
+import diskCacheV111.vehicles.IpProtocolInfo;
+import diskCacheV111.vehicles.ProtocolInfo;
 import dmg.cells.nucleus.CellCommandListener;
 import dmg.cells.nucleus.CellSetupProvider;
 import dmg.util.CommandException;
@@ -36,11 +11,26 @@ import dmg.util.command.Argument;
 import dmg.util.command.Command;
 import dmg.util.command.DelayedCommand;
 import dmg.util.command.Option;
-
 import org.dcache.pool.FaultEvent;
 import org.dcache.pool.FaultListener;
 import org.dcache.pool.classic.MoverRequestScheduler.Order;
 import org.dcache.util.IoPriority;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+import java.io.PrintWriter;
+import java.io.Serializable;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.*;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+import java.util.function.Predicate;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static dmg.util.CommandException.checkCommand;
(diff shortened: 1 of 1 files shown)


modules/dcache/src/main/java/org/dcache/pool/classic/IoQueueManager.java
```diff
@@ -406,6 +413,9 @@ public synchronized String call() throws IllegalArgumentException, NoSuchElement
         @Option(name = "r", usage = "Sort output in reverse order.")
         boolean reverseSort;
 
+        @Option(name = "c", usage = "Filter output by client IP or hostname.")
```
paulmillar: `c` is OK, but I think we can be more generous with our name!  perhaps `client` isn't unreasonable, here.  Incidentally, the others (`r`, `S` and `t`) are single letters because this is how the Unix `ls` command works.

modules/dcache/src/main/java/org/dcache/pool/classic/IoQueueManager.java
```diff
@@ -451,16 +461,43 @@ public Serializable call() throws NoSuchElementException
                     comparator = comparator.reversed();
                 }
 
+                // Prepare filters
+                Predicate<MoverRequestScheduler.PrioritizedRequest> clientIpFilter;
+                if (clientHost != null && !clientHost.isEmpty()) {
+                    try {
+                        // check and resolve host
```
paulmillar: This is a non-blocking comment (as it's mostly stylistic), but these comments don't actually help.  The code already describes what it does, so the comments don't add anything.

Instead, they're potential problems in the future -- when someone refactors this code, if they forget to update/move the comments then there's a potential inconsistency between what the code does and what the comment says it should be doing.  This makes debugging harder.

Comments are reasonable to describe *why* a particular behaviour is important, rather than describing what the code is doing -- if you find need a comment to describe what the code is doing, it's better to rewrite the code! :smile:

modules/dcache/src/main/java/org/dcache/pool/classic/IoQueueManager.java
```diff
@@ -451,16 +461,43 @@ public Serializable call() throws NoSuchElementException
                     comparator = comparator.reversed();
                 }
 
+                // Prepare filters
+                Predicate<MoverRequestScheduler.PrioritizedRequest> clientIpFilter;
+                if (clientHost != null && !clientHost.isEmpty()) {
```
paulmillar: clientHost should never be an empty string -- the code should only need to test for `null`.

modules/dcache/src/main/java/org/dcache/pool/classic/IoQueueManager.java
```diff
@@ -451,16 +461,43 @@ public Serializable call() throws NoSuchElementException
...
+                    } catch (UnknownHostException e) {
+                        return String.format("Unknown host %s", clientHost);
+                    }
+
+                    // building the filter
+                    clientIpFilter = prioritizedRequest -> {
+                        ProtocolInfo info = prioritizedRequest.getMover().getProtocolInfo();
+                        // ip protocol only
+                        if (info instanceof IpProtocolInfo) {
+                            return ((IpProtocolInfo) info).getSocketAddress().equals(clientHost);
```
paulmillar: `getSocketAddress` returns an `InetSocketAddress`, whereas `clientHost` is of type `String`.  This means this test will never succeed.

Incidentally, testing IP addresses by doing a String comparison is a somewhat risky strategy -- it probably works with IPv4 address, but an IPv6 address can be "written as a String" in different ways.

modules/dcache/src/main/java/org/dcache/pool/classic/IoQueueManager.java
```diff
@@ -451,16 +461,43 @@ public Serializable call() throws NoSuchElementException
...
                 }
 
+                // Prepare filters
+                Predicate<MoverRequestScheduler.PrioritizedRequest> clientIpFilter;
+                if (clientHost != null && !clientHost.isEmpty()) {
+                    try {
+                        // check and resolve host
+                        clientHost = InetAddress.getByName(clientHost).getHostAddress();
+                    } catch (UnknownHostException e) {
+                        return String.format("Unknown host %s", clientHost);
```
paulmillar: A better way of handling this is to throw a CommandException.  There are a few places where there's a distinction between a command exiting normally and abnormally (e.g., the colour of the output).

modules/dcache/src/main/java/org/dcache/pool/classic/IoQueueManager.java
fritzendgltig: What if I say `mover ls --client ""`? Wouldn't this cause such a scenario? Why is queueName being checked then?

kofemann:
ok to test

modules/dcache/src/main/java/org/dcache/pool/classic/IoQueueManager.java
paulmillar: Oops, sorry, I was thinking about a different part of the code-base.

Your right, it is possible to inject an empty string.

(The syntax is a little out; the correct syntax would be `-client=""`.  The equals is needed and the double-dash actually signifies that everything subsequent to this is arguments, even if they look like options.)

You can use Guava's `com.google.common.base.Strings` class' `isNullOrEmpty` method to make the test a little bit less verbose.

paulmillar:
You need to rebase your patch against a newer version of master.

Unfortunately master broke for a short period and you were unlucky and rebased against master when it was broken.  This is why the build failed.

Rebasing against the current tip of master should fix this problem.

modules/dcache/src/main/java/org/dcache/pool/classic/IoQueueManager.java
```diff
@@ -1,23 +1,29 @@
...
 import java.io.PrintWriter;
 import java.io.Serializable;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Comparator;
-import java.util.Optional;
+import java.util.*;
```
anupash: dCache code style doesn't allow for star non-static imports.

jstarek:
Closing due to long inactivity. Feel free to re-submit once the outstanding issues are resolved.

PR #3441, opened 2017-08-10
dcache-chimera: refactoring slf4j logging messages

marisanest:
Motivation:

Modification:

Result:
Gained efficiency

### DIFF
--- a/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
+++ b/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraCleaner.java
@@ -322,7 +322,7 @@ public class ChimeraCleaner extends AbstractCell implements Runnable
                 runNotification();
             }
         } catch (DataAccessException e) {
-            _log.error("Database failure: " + e.getMessage());
+            _log.error("Database failure: {}", e.getMessage());
         } catch (InterruptedException e) {
             _log.info("Cleaner was interrupted");
         } catch (RuntimeException e) {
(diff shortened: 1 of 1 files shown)


alrossi:
ok to test

marisanest:
Hi,

I just saw that all builds of my pull requests went wrong and I don't know exactly why. I build it local on my desktop and it worked. Do you have an idea?

Thanks

paulmillar:
The problem is because, for a short while, building dCache RPM was broken.

It's fixed now, so if you rebase your patch against the current tip of master, you should be fine.

marisanest:
Hi Paul, 

thanks! By rebasing you mean to make a pull from upstream (dcache master) and then pushing it again to my branch?

paulmillar:
Hiya,

By default, doing `git pull upstream` will do a (non-"fastforward", i.e., non-trivial) merge to bring in the commits that happened after you created your branch.  This is always fine, but it does result in an extra commit.

The alternative is to do a rebase (e.g., `git pull --rebase upstream`).  Rebasing is more intrusive (it "rewrites" your branch's history) and dangerous option if you share your branch with others, but there's no extra commit, so the end result is somewhat simpler.

Either approach should work.

If you're not sure which to do, then use `git pull upstream` if you are sharing this branch with someone else, otherwise use `git pull --rebase upstream`.

marisanest:
Hi, 

I used git pull --rebase upstream but it seems that it didn't work properly... Is it right that there are now 67 changed files?

paulmillar:
It looks like git somehow ignored the `--rebase` and did a regular merge instead.  Perhaps this was when you pushed your new branch into github.

OK, let's try something quite basic as an "easy" way for you to recover.

Create a new branch from `upstream/master` and cherry-pick your patches one by one using their commit IDs.  I believe you have two patches, with commit IDs of a768abfa and 83fba942, right.

(In effect, This is doing the rebase manually.)

You can then rename the branch you pushed into github (`refactor/chimera2`) to something else (`refactor/chimera2-old`, say), and rename the branch you just created to be `refactor/chimera2`.

You should now have a branch with the right name (`refactor/chimera2`) that is up-to-date with master and contains your two patches.

Once that's done you can push your new branch into github, which will automatically update your pull request.  When pushing you will need to use the forced option, since now your local branch has a different history to the one in github.  You can use the `+` prefix to force the change: something like `git push +HEAD origin`.

With that all done, you should have an updated github pull-request with your two patches, which should build correctly.

marisanest:
Thanks for the accurate advice!

paulmillar:
Thanks, looks good.

PR #3446, opened 2017-08-10
dcache-srm: refactoring slf4j logging messages

marisanest:
Motivation:

Modification:

Result:
Gained efficiency

### DIFF
--- a/modules/dcache-srm/src/main/java/diskCacheV111/srm/dcache/SrmReserveSpaceCompanion.java
+++ b/modules/dcache-srm/src/main/java/diskCacheV111/srm/dcache/SrmReserveSpaceCompanion.java
@@ -101,7 +101,7 @@ public final class SrmReserveSpaceCompanion
     @Override
     public void failure(int rc, Object error)
     {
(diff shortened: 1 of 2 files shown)


alrossi:
ok to test

paulmillar:
retest this please

modules/dcache-srm/src/main/java/diskCacheV111/srm/dcache/SrmReserveSpaceCompanion.java
```diff
@@ -101,7 +101,7 @@ private SrmReserveSpaceCompanion(SrmReserveSpaceCallback callback)
     @Override
     public void failure(int rc, Object error)
     {
-        LOGGER.error("Space Reservation Failed rc:" + rc + " error:" + error);
+        LOGGER.error("Space Reservation Failed rc: {} error:{}", rc, error.toString());
```
kofemann: As it not guaranteed, that error is not **null**, you can get a NullPointerException here. String#valueOf() will be  a safer solution here.

kofemann:
@marisanest  Thanks!

PR #3452, opened 2017-08-10
srm-common: refactoring slf4j logging messages

marisanest:
Motivation:

Modification:

Result:
Gained efficiency

### DIFF
--- a/modules/srm-common/src/main/java/org/dcache/srm/util/GridftpClient.java
+++ b/modules/srm-common/src/main/java/org/dcache/srm/util/GridftpClient.java
@@ -114,10 +114,10 @@ public class GridftpClient
     {
         if(bufferSize >0) {
             _bufferSize = bufferSize;
-            logger.debug("memory buffer size is set to "+bufferSize);
+            logger.debug("memory buffer size is set to {}", bufferSize);
         }
         _host = host;
-        logger.debug("connecting to "+_host+" on port "+port);
+        logger.debug("connecting to {} on port {}", _host, port);
 
         ClientGsiEngineDssContextFactory dssContextFactory =
                 new ClientGsiEngineDssContextFactory(
@@ -258,7 +258,7 @@ public class GridftpClient
         if(emode) {
             _client.setMode(GridFTPSession.MODE_EBLOCK);
             // adding parallelism
-            logger.debug("parallelism: " + _streamsNum);
+            logger.debug("parallelism: {}", _streamsNum);
             _client.setOptions(new RetrieveOptions(_streamsNum));
         }
         else {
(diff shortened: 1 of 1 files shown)


modules/srm-common/src/main/java/org/dcache/srm/util/GridftpClient.java
```diff
@@ -397,16 +397,16 @@ else if(size - sink.getTransfered() <0) {
             throw new IOException("we wrote more then file size!!!");
         }
 
-        logger.debug("gridFTPWrite() wrote "+sink.getTransfered()+"bytes");
+        logger.debug("gridFTPWrite() wrote {}bytes", sink.getTransfered());
```
alrossi: let's add a space here ({} bytes)

modules/srm-common/src/main/java/org/dcache/srm/util/GridftpClient.java
```diff
@@ -397,16 +397,16 @@ else if(size - sink.getTransfered() <0) {
...
 
         try {
           if ( _cksmType != null ) {
               verifyCksmValue(_current_source_sink, sourcepath);
           }
         } catch ( ChecksumNotSupported ex){
-          logger.error("Checksum is not supported:"+ex.toString());
+          logger.error("Checksum is not supported:{}", ex.toString());
```
alrossi: same with these (: {})

modules/srm-common/src/main/java/org/dcache/srm/util/GridftpClient.java
```diff
@@ -524,7 +524,7 @@ else if(diskFileLength < source.getTransfered() ) {
             throw new IOException("we read more then file size!!!");
         }
 
-        logger.debug("gridFTPWrite() wrote "+source.getTransfered()+"bytes");
+        logger.debug("gridFTPWrite() wrote {}bytes", source.getTransfered());
         getTransfered();
```
alrossi: here too

modules/srm-common/src/main/java/org/dcache/srm/util/GridftpClient.java
```diff
@@ -549,7 +549,7 @@ private void sendCksmValue(IDiskDataSourceSink source)
             }
         } catch ( Exception ex ){
             // send cksm error is often expected for non dCache sites
-            logger.debug("Was not able to send checksum value:"+ex.toString());
+            logger.debug("Was not able to send checksum value:{}", ex.toString());
```
alrossi: and here

alrossi:
OK to test

kofemann:
Thanks!

PR #3453, opened 2017-08-10
srm-server: refactoring slf4j logging messages

marisanest:
Motivation:

Modification:

Result:
Gained efficiency

### DIFF
--- a/modules/srm-server/src/main/java/org/dcache/srm/scheduler/Scheduler.java
+++ b/modules/srm-server/src/main/java/org/dcache/srm/scheduler/Scheduler.java
@@ -317,7 +317,7 @@ public class Scheduler <T extends Job>
 
                         synchronized (this) {
                             if (!hasBeenNotified) {
-                                //logger.debug("Scheduler(id="+getId()+").run() waiting for events...");
+                                logger.debug("Scheduler(id={}"getId()+").run() waiting for events...");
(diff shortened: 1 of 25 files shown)


modules/srm-server/src/main/java/org/dcache/srm/scheduler/Scheduler.java
```diff
@@ -317,7 +317,7 @@ public void run()
 
                         synchronized (this) {
                             if (!hasBeenNotified) {
-                                //logger.debug("Scheduler(id="+getId()+").run() waiting for events...");
```
alrossi: I'm wondering whether this even should be here, since it was commented out.

marisanest:
Should I remove the commented out logging statements?

alrossi:
Hi Marisa,  

first, thanks so much for going through all these files!

Yes, I would just remove the commented out statement.  Thanks

marisanest:
Hi,

I was just a bit concerned about mixing to different modifications, 
but if it is ok for you, I'll do that!

Thanks for the constructive review :)

alrossi:
No problem.  If you prefer adhering to the letter of the law, then comment out the modified log statement and leave it in the file ... that may be the more correct choice.

alrossi:
OK to test

kofemann:
Thanks !

PR #3455, opened 2017-08-10
cells: refactoring slf4j logging messages

marisanest:
Motivation:

Modification:

Result:
Gained efficiency

### DIFF
--- a/modules/cells/src/main/java/dmg/cells/nucleus/CellAdapter.java
+++ b/modules/cells/src/main/java/dmg/cells/nucleus/CellAdapter.java
@@ -518,9 +518,9 @@ public class CellAdapter
      * @param msg the reference to message arrived.
      */
     public void messageArrived(CellMessage msg) {
-        _log.info(" CellMessage From   : " + msg.getSourcePath());
-        _log.info(" CellMessage To     : " + msg.getDestinationPath());
-        _log.info(" CellMessage Object : " + msg.getMessageObject());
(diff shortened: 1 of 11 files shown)


modules/cells/src/main/java/dmg/cells/nucleus/CellAdapter.java
```diff
@@ -808,7 +808,7 @@ void dumpPinboard()
      */
     @Override
     public void   exceptionArrived(ExceptionEvent ce) {
-        _log.info(" exceptionArrived "+ce);
+        _log.info(" exceptionArrived {}", ce);
```
kofemann: here we probably want to have _log.info(" exceptionArrived {}", ce.toString()); to avoid logging of a stack trace

modules/cells/src/main/java/dmg/cells/services/login/TelnetSAuth_A.java
```diff
@@ -130,7 +130,7 @@ public boolean isPasswordOk( InetAddress host , String user , String passwd ){
              }
              return true ;
          }catch( Exception e ){
-            _log.info( "Exception in TelnetSAuth_A : "+ e ) ;
+            _log.info( "Exception in TelnetSAuth_A : {}", e ) ;
```
kofemann: the same here

kofemann:
ok to test

kofemann:
Thanks!

Issue #3425, opened 2017-08-07
Ssh2Admin -> AdminPublicKeyAuthenticator not checking if user belongs to Admin group


solen0id:
Hey all,

It looks like the AdminPublicKeyAuthenticator.authenticate() method is not checking if the users GID belongs to the admin-group. When building for system test, this allowed me to log in as an arbitrary (and/or non-existent) user, as long as the public-key used to authenticate is present in authorized_keys. 

Is this something that would not occur in a production system with proper user management and is unique to the system test package?

kofemann:
The assumption is that people with public key in authorized_keys file are admins. In case of username+pass, any valid user can connect to the admin interface, but only admins will be let in. IOW, looks like it works as designed.

stefaus:
I think the AdminPublicKeyAuthenticator should at very least check if the username belongs to an existing user, better if belongs to admin group.
Optional would be to bind public keys to usernames in authorized_keys2 file, to avoid possible impersonation fraud.

kofemann:
Your are right. We do not check, that the key matches the user. This is of course if not correct.

paulmillar:
I think the original topic (no checking if user member-of admin group) has been dealt with by @kofemann comment "The assumption is that people with public key in authorized_keys file are admins." So this works as expected.

The additional work (binding identity in authorized_keys2 file to user identity) is certainly appreciated, but slightly different from the original question.

Therefore, I think we can close this issue.

PR #3459, opened 2017-08-11
ssh: add username check to pubkey authentication

stefaus:
Motivation:

At the moment there is no check if a username belongs to a public key in
authorized_keys2 file, which leads to a possible impersonation fraud.
More information in Issue #3425.

Modification:

Changed Ssh2Admin#AdminPublickeyAuthenticator to check if used public key
and username in key file match.
Additionally add a reason to the logged authentication event in case
mapping fails or host check fails.

Result:

Username and key must match, otherwise authentication fails and get
logged with information about failure to domain.access file.

Target: master
Require-notes: yes
Require-book: yes

Signed-off-by: Stefan Moll <stefan@stefaus.de>

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
+++ b/modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
@@ -34,9 +34,10 @@ import java.security.PublicKey;
 import java.security.spec.InvalidKeySpecException;
 import java.util.EnumSet;
 import java.util.List;
-import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


paulmillar:
ok to test

paulmillar:
retest this please

modules/dcache/src/main/java/org/dcache/services/ssh2/Ssh2Admin.java
```diff
@@ -69,6 +70,12 @@
     private static final Logger _log = LoggerFactory.getLogger(Ssh2Admin.class);
     private static final Logger _accessLog =
             LoggerFactory.getLogger("org.dcache.access.ssh2");
+
+    // get the user part from a public key line like
+    // ... ssh-rsa AAAA...... user@hostname
```
kofemann: I would do '@hostname'

kofemann:
Well done, Thanks!

PR #3462, opened 2017-08-14
Task 3 -  Modernizing: Replace Guava Methods with native Java

rwenz3l:
Motivation:
Modernizing dCache code to reduce dependencies and rely on Native Java

Modification:
Remove Google Guava Methods and Replace with native methods.
Completely where possible, otherwise partially)

Result:
Reduces dependencies and calls to non-native classes.

Target: master

Signed-off-by: Robin Wenzel info@robin-wenzel.de

### DIFF
--- a/modules/chimera/src/main/java/org/dcache/chimera/FsInode.java
+++ b/modules/chimera/src/main/java/org/dcache/chimera/FsInode.java
@@ -16,8 +16,6 @@
  */
 package org.dcache.chimera;
 
-import com.google.common.primitives.Longs;
-
 import java.nio.ByteBuffer;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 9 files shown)


alrossi:
OK to test

anupash:
Some are still left,

e.g.
```
modules/dcache/src/main/java/org/dcache/pinmanager/PinManagerCLI.java:                Long id = Longs.tryParse(s);
```
You can use Long#parseLong. But not the differences between Guava Longs and Java Long class

https://google.github.io/guava/releases/21.0/api/docs/com/google/common/primitives/Longs.html#tryParse-java.lang.String-

anupash:
Update the commit message to follow the dCache commit messages style. Module field can have multiple module fields.

module: summary line

Motivation:

Modification

PR #3467, opened 2017-08-31
srmclient: remove non-functioning script with BASH dependencies

paulmillar:
Motivation:

The srmcp script contains some BASHisms that prevent it from being
used in environments where /bin/sh is not bash.

Modification:

Since the BASH-specific features (arrays) were limited to some
functionality that was actually broken, this patch simply removed the
offending part and refactors the argument parsing accordingly.

Result:

Much simpler code that is also POSIX compliant.

Target: master
Ticket: http://rt.dcache.org/Ticket/Display.html?id=9256
Request: 3.2
Request: 3.1
Patch: https://rb.dcache.org/r/10434/
Acked-by: Dmitry Litvintsev
Require-notes: no
Require-srmclient-notes: yes
Require-book: no

### DIFF
--- a/modules/srm-client/src/main/bin/srmcp
+++ b/modules/srm-client/src/main/bin/srmcp
@@ -2,31 +2,6 @@
 
 @INIT_SCRIPT@
 
-args=$*
-
-error() {
-    echo "Can't copy from \"${1}\" protocol to \"${2}\"  protocol"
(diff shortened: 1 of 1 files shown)


paulmillar:
retest this please

paulmillar:
retest this please

PR #3468, opened 2017-09-01
common: fix bug in histogram max index computation

alrossi:
Motivation:

The configure method for CountingHistograms sets
the highest bin value on the basis of the largest
value in the dataset.  If this is undefined, Double.MAX_VALUE
is used.  This, however, results in an effective infinity
as the highest index, and results in aggregate plots
where all the values are squashed into the first bin.

Modification:

If the metadata max value is undefined (because the dataset
is empty), set the max index to the preset number of bins - 1.

Result:

Plots created from data aggregation where some of the set
members may be empty do not result in an x-axis
with the highest value being infinity.

Target: master
Require-notes: yes
Require-book: no
Request: 3.2
Acked-by: Tigran


mksahakyan:
retest this please

mksahakyan:
retest this please

PR #3469, opened 2017-09-01
dcache-restful-api: extract the locations from storage info and add t…

alrossi:
…o JSON attributes

Motivation:

StorageInfo has a List<URI> locations field, but no standard getter and setter.
However, the object is being used to populate the JSONAttributes returned
by the namespace service.  It is of interest to us to expose all known tape
URIs in dcache-view.

Modification:

Extract this list separately and set it on JSONAttibutes.

The alternative (which may be preferable to some) would be to change
the locations() method to getLocations() and add setLocations (currently
there is only addLocation(URI)).  However, this will make for a much
larger patch.

Result:

SURIs are now visible in the admin view.

Target: master
Request: 3.2
Require-notes: no
Require-book: no
Acked-by: Paul


mksahakyan:
retest this please

mksahakyan:
retest this please

PR #3470, opened 2017-09-01
resilience: fix bug in formatting and handling of cache exception types

alrossi:
Motivation:

The utility which returns a CacheException object with
an appropriate message and type based on the failure type
and operation uses string formatting to process arguments.
The way the code is written, however, is subject to bad
formatting exceptions, particularly if there are unexpected
null values passed in as parameters (for instance, if
an exception cause is null).  In some of these cases,
the actual exception is masked because of the formatting
exception, causing the removal of the operation from
the map to fail.

Modification:

Change the method such that it accounts for nulls; this
also requires respecting the contract that the format
strings used will always have three % placeholders.

In addition to this, the "Will retry" message appended
to the failure message should not be included if the
error is fatal.

Result:

Correct formatting will not block proper removal
from the operation map and potentially freeze
progress in the resilience service as a whole.

Target: master
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16
Require-notes: yes
Require-book: no
Acked-by: Tigran

### DIFF
--- a/modules/dcache-resilience/src/main/java/org/dcache/resilience/handlers/FileOperationHandler.java
+++ b/modules/dcache-resilience/src/main/java/org/dcache/resilience/handlers/FileOperationHandler.java
@@ -586,7 +586,7 @@ public class FileOperationHandler {
         }
 
         Serializable exception = msg.getErrorObject();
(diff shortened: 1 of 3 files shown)


mksahakyan:
retest this please

mksahakyan:
retest this please

PR #3474, opened 2017-09-01
resilience: distinguish correctly between file not in repository and …

alrossi:
…file not found

Motivation:

If resilience tries to copy or remove a file, and it is no longer
in the namespace, this should be a fatal error, differently from
the discovery that a replica is no longer in cache.  These two
errors, however, are currently treated the same way.

Modification:

Make FileNotFound always a fatal error.

Result:

Correct behavior (immediate abort of operation) when
file is not in namespace.

Target: master
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16
Require-notes: yes
Require-book: no
Acked-by: Tigran


mksahakyan:
retest this please

mksahakyan:
retest this please

PR #3478, opened 2017-09-01
resilience: handle all cases where no locations for file may be disco…

alrossi:
…vered

Motivation:

When resilience discovers that there are currently no replicas
for a file which is still in the namespace, it should take action
(depending on the inaccessible file handler implementation).

One common cause for this may be that all the pools containing
these replicas may be offline.  But there are also conditions
in which a query of the namespace may return zero locations;
whether this is a bug or not, resilience should not fail or
become blocked under these circumstances.

Modification:

Add the necessary code paths for handling the missing conditions.
This includes an additional method on the inaccessible file handler.

JUnit tests have been adjusted to account for a fuller range of
situations when all replicas are missing.

Result:

Fuller coverage for handling this potential condition without
blocking the operation.

Target:  master
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16
Require-notes: yes
Require-book: no
Acked-by: Tigran


mksahakyan:
retest this please

mksahakyan:
retest this please

PR #3484, opened 2017-09-05
webdav: adjust header parsing to be case insensitive

paulmillar:
Motivation:

In RFC 7230, section 3.2, it states that HTTP header names should be
case insensitive.  Despite this, the standard milton mechanism for
acquiring header information is case sensitive, which (in turn) means
that several HTTP header values in dCache are similarly case sensitive.
Affected headers are RequireChecksumVerification, Credential,
Authorization (only when used to auto-select OIDC delegation) and the
TransferHeader prefix.

We have received reports that this case sensitivity has caused problems
with certain clients (e.g., go).

Modification:

Switch to using the HttpServletRequest object to acquire header
information, which is documented as case insensitive.  Additionally, the
parsing of RequireChecksumVerification is moved from
RemoteTransferHandler to CopyFilter.

Result:

Headers that control third-party transfers are now case insensitive, so
should work with more clients.

Target: master
Request: 3.1
Request: 3.0
Request: 2.16
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/10414/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/CopyFilter.java
+++ b/modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/CopyFilter.java
@@ -72,8 +72,6 @@ import org.dcache.webdav.transfer.RemoteTransferHandler.TransferType;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.concurrent.TimeUnit.MINUTES;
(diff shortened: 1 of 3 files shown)


paulmillar:
retest this please

mksahakyan:
retest this please

mksahakyan:
retest this please

mksahakyan:
retest this please

PR #3492, opened 2017-09-07
nfs: shutdown callback ScheduledExecutorService on shutdown

paulmillar:
Motivation:

The nfs doors do not shut down cleanly, but rely on the cell
infrastructure to kill off threads that may have been started.  This
slows down the shutdown sequence, which risks the daemon script killing
the process outright.

Modification:

Shutdown the ScheduledExecutorService as part of the NFS door shutdown
sequence.

The Threads that are created for this are given a more descriptive name.

Result:

dCache shutdown is faster and less likely to suffer problems.

Target: master
Request: 3.2
Request: 3.1
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/10447/
Acked-by: Tigran Mkrtchyan

Conflicts:
	modules/dcache-nfs/src/main/java/org/dcache/chimera/nfsv41/door/NFSv41Door.java


paulmillar:
retest this please

PR #3503, opened 2017-09-07
dcache-restful-api: return incomplete info instead of throwing NoSuch…

alrossi:
…ElementException

Motivation:

It is possible under some circumstances that cell info cannot be
collected from a service that is mapped by the RoutingManager.

When this occurs, and the Frontend service tries to access the
missing info, the code currently throws a NoSuchElementException.
This behavior, unfortunately, nullifies the entire operation,
resulting in HTTP 500 error being reported in the browser.

A better solution would be to return a data object with an
"Unknown" state indicated for that service.

Modification:

Add the code to the service implementation to return this
placeholder.

The patch also removes an unnecessary null check in the code
which processes the returned future.

Result:

The stack trace below should not appear; the service should
be marked as in an unknown state on the frontend.

Target: master
Require-book: no
Require-notes: no
Request: 3.2
Acked-by: Tigran

### DIFF
--- a/modules/dcache-restful-api/src/main/java/org/dcache/restful/services/cells/CellInfoServiceImpl.java
+++ b/modules/dcache-restful-api/src/main/java/org/dcache/restful/services/cells/CellInfoServiceImpl.java
@@ -61,19 +61,18 @@ package org.dcache.restful.services.cells;
 
 import java.util.Arrays;
 import java.util.Map;
-import java.util.NoSuchElementException;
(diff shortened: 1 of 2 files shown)


mksahakyan:
retest this please

mksahakyan:
retest this please

mksahakyan:
retest this please

paulmillar:
retest this please

mksahakyan:
retest this please

PR #3515, opened 2017-09-11
common: fix time computation in TimeseriesHistogramTest

alrossi:
Motivation:

Recently we observed a failure of one of the unit tests
for the TimeseriesHistogram (stack trace given in the
testing section).

Examination of the code has not revealed the cause of the failure,
which so far has not been systematically reproducible.

However, a similar potential bug does exist regarding
the computation of 'now', which is inconsistent and may
occasionally create a situation when the buffer is
erroneously rotated during the test.

Modification:

Make sure that the time of the last bin matches the
update time exactly by initializing a single "now"
and reusing it.

Result:

This fixes the (unobserved, potential) bug.
Whether it also fixes the observed failure remains to
be seen.

Target: master
Request: 3.2
Require-book: no
Require-notes: no
Acked-by: Paul

### DIFF
--- a/modules/common/src/main/java/org/dcache/util/histograms/TimeseriesHistogram.java
+++ b/modules/common/src/main/java/org/dcache/util/histograms/TimeseriesHistogram.java
@@ -251,7 +251,7 @@ public class TimeseriesHistogram extends HistogramModel
                     break;
             }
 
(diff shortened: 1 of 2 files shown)


mksahakyan:
retest this please

PR #3538, opened 2017-09-20
cells: better error reporting to ensure bugs are understood

paulmillar:
Motivation:

Observed the following error on prometheus:

    18 Sep 2017 09:40:55 (System) [] URL [file:/usr/share/dcache/services/zookeeper.batch]: line 42: Command failed ((3) null; caused by: null)

This is problematic as the error message provides almost no information
on what went wrong -- making it almost impossible to understand or fix.

Modification:

Ensure that 'null' message is never logged -- if the message contains
nothing useful, show the class name instead.

The above example could be due to a RuntimeException (e.g., an NPE)
being wrapped with another exception.  Wrapping a RuntimeException is
itself a bug, but one that is compensated for by checking each cause and
showing the normal stack-trace if one is found.

Result:

Better error reporting for certain failure modes.

Target: master
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/10469/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/cells/src/main/java/dmg/cells/nucleus/CellShell.java
+++ b/modules/cells/src/main/java/dmg/cells/nucleus/CellShell.java
@@ -1743,7 +1743,7 @@ public class CellShell extends CommandInterpreter
                     } else if (!(error instanceof CommandEvaluationException)) {
                         String msg =
                             Exceptions.getMessageWithCauses(error);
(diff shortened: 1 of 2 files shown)


mksahakyan:
retest this please

PR #3557, opened 2017-09-27
dcache-restful-api: Add POST for enable/disable or kill mover to pool…

alrossi:
… info service

Motivation:

As a concession to legacy support of commands issued
through the webadmin interface, it has been agreed that
the new RESTful API will also allow disabling and enabling
of pools as well as killing movers from the active transfers
page.

Modification:

Add the necessary code to the frontend service to support
this.

Result:

These legacy commands are supported in the new frontend.

Target: master
Request: 3.2
Acked-by: Paul
Require-notes: yes
Require-book: yes


kofemann:
ok to test

mksahakyan:
retest this please

PR #3561, opened 2017-09-28
dcache-xrootd: Fix login handshake to support xrootd clients (> 4.7.0)

vingar:
Motivation:

In the 4.7 releases, xrootd client started enforcing protocol requirements
for kXR_login which, unfortunately, broke access to dCache.
xrootd client expects an answer with a 16-character session ID from the door and then
the pool after the redirection.

The login spec can be found at: http://xrootd.org/doc/dev45/XRdv310.htm#_Toc464248819

Without it, the 4.7.0 client declared an error and retries after a while. This will also
fail and so we go into what appears to be a hang. Eventually (after several minutes)
the login will fail with an error.

Modification:

The login answer following the spec has been added in the xrootd pool.

Result:

It works for both xrootd 4.6.1 and 4.7.0.

Target: master
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/10510
Committed: master@


jstarek:
retest this please

Issue #3547, opened 2017-09-25
psu(x) match command fails with FNAL storage unit syntax


alrossi:
I discovered that neither the old page nor the webadmin page for Pool Selection Setup works when issuing a simulated match request.   

Adding a print stack trace, I discovered:

Caused by: dmg.util.CommandThrowableException: (3) java.lang.IllegalArgumentException: Invalid format: expected<x:y@z> got <enstore> from ac_psux_match_$_5
       at dmg.util.command.AcCommandExecutor.execute(AcCommandExecutor.java:177)
       at dmg.cells.nucleus.CellAdapter.executeCommand(CellAdapter.java:238)
       at org.dcache.cells.UniversalSpringCell.executeCommand(UniversalSpringCell.java:195)
       at dmg.cells.nucleus.CellAdapter$1.doExecute(CellAdapter.java:104)
       at org.dcache.util.cli.CommandInterpreter.command(CommandInterpreter.java:129)
       at dmg.cells.nucleus.CellAdapter$1.command(CellAdapter.java:121)
       at dmg.cells.nucleus.CellAdapter.command(CellAdapter.java:223)
       at dmg.cells.nucleus.CellAdapter.executeLocalCommand(CellAdapter.java:924)
       at dmg.cells.nucleus.CellAdapter.messageArrived(CellAdapter.java:855)
       at dmg.cells.nucleus.CellNucleus$DeliverMessageTask.run(CellNucleus.java:1218)
       at org.dcache.util.BoundedExecutor$Worker.run(BoundedExecutor.java:251)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
       at dmg.cells.nucleus.CellNucleus.lambda$wrapLoggingContext$7(CellNucleus.java:754)
       at dmg.cells.nucleus.CellNucleus$$Lambda$27/461591680.run(Unknown Source)
       ... 1 more
Caused by: java.lang.IllegalArgumentException: Invalid format: expected<x:y@z> got <enstore>
       at diskCacheV111.vehicles.GenericStorageInfo.valueOf(GenericStorageInfo.java:397)
       at diskCacheV111.poolManager.PoolSelectionUnitV2.matchLinkGroupsXml(PoolSelectionUnitV2.java:1588)
[...]

PR #3563, opened 2017-09-28
dcache: fix bug in PoolSelectionUnitV2 match()

alrossi:
Motivation:

See issue #3547.

Manifested by the broken commands psu / psux match as well
as on the Pool Selection Setup pages (httpd and webadmin).

The problem lies in the FileAttributes being created using
a StorageUnit; hence the STORAGECLASS, HSM and CACHECLASS
attributes are never defined.

Modification:

Obtain storage class, hsm and cache class from the storage
unit, not directly from the attributes.

Result:

The exception is no longer thrown and matching can
proceed as normal.

Target: master
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16
Require-notes: yes
Require-book: no
Bug: #3547
Acked-by: Tigran


jstarek:
retest this please

PR #3571, opened 2017-10-04
dcache-restful-api: fix aggregation issues for pool info and move agg…

alrossi:
…regation to history service

Motivation:

In making the pool info service stateless, collection of timeseries
information for queue status and file lifetime on the pools was
moved into a separate service (history).

This, however, introduced several regressions.  The first was
that new containers were being created for both pools and
pool groups at each pass of the collection, so that the
data was never being updated, simply overwritten.
The second, however, was the oversight in having all
the aggregation still occur in the frontend service. This
means that the timeseries data for individual pools survives
service restarts, but the aggregated data does not.

Modification:

The minor bugs concerning updating vs overwriting have
been repaired.  But more importantly, the aggregation of
mover request and file lifetime histogram data is now
all done in the history service, and is simple requested
by the pool info service (frontend).  This modification
entailed factoring out some methods into common utilities,
the creation of another abstract class with two different
implementations (frontend vs history), and a number of
other attendant changes.

Note that the frontend service still aggregates space
data, since this data can always be reconstructed and
is not persisted to disk by the history service.

Result:

The proper behavior with respect to aggregated (pool group)
timeseries data is now observed.

Target: master
Request: 3.2
Require-notes: no
Require-book: no
Acked-by: Tigran

### DIFF
--- a/modules/dcache-history/src/main/java/org/dcache/services/history/pools/PoolHistoriesRequestProcessor.java
+++ b/modules/dcache-history/src/main/java/org/dcache/services/history/pools/PoolHistoriesRequestProcessor.java
@@ -74,24 +74,27 @@ import java.util.Map;
 import java.util.Map.Entry;
 
 import diskCacheV111.pools.json.PoolCostData;
+import diskCacheV111.util.CacheException;
 import org.dcache.pool.classic.json.SweeperData;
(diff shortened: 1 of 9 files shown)


alrossi:
Resubmitting combined pull request for RB 10430 and 10523

PR #3573, opened 2017-10-04
apache sshd: upgrade to 1.3.0 due to security vulnerability

DmitryLitvintsev:
Fermilab CST flagged dCache admin port as vulnerable due to 👍 

The remote host is running a version of SSH that is older than (or as old as) version 1.2.23.

The remote version of this software is vulnerable to a known plaintext attack, which could allow an attacker to insert encrypted packets in the client - server stream that will be deciphered by the server, thus allowing the attacker to execute arbitrary commands on the remote server

Solution: Upgrade to version 1.2.25 of SSH which solves this problem.

The upgrade to 1.3.0 apache sshd solves the problem.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -498,7 +498,7 @@
             <dependency>
                 <groupId>org.apache.sshd</groupId>
                 <artifactId>sshd-core</artifactId>
-                <version>1.2.0</version>
+                <version>1.3.0</version>
             </dependency>
             <dependency>
                 <!-- Newer versions have two problems. A performance regression causes it to block on non-responding network


kofemann:
are you sure you need this as ssh-1.2.x and apache-sshd-1.2.x are completely different  products. Moreover, newer versions of sshd pull newer versions of boncycasel, which may break grid related staff.

DmitryLitvintsev:
Tigran: I am *somewhat* sure. I re-deployed dCache w/ this 1.3.0 apache sshd and asked Fermilab CST to re-scan. They reported that the scan passed clean. 
It is quite possible that the scan just detects version and if it is > "1.2.x" it reports success. 
The end result - we are not bothered by security w/ this patch.

PR #3581, opened 2017-10-10
pool: dont disable pool if mover cancelled before open

paulmillar:
Motivation:

A bug in gfal2 results in FTP transfers being aborted some 50 ms after
being initiated.  This results in the door killing the mover shortly
after the pool received the PoolDeliverFile message.  If the mover is
not queued, but not yet fully started, the file open results in an
AsynchronousCloseException being thrown.  As an IOException, this is
treated as a problem with the repository.  The result is the pool will
disabling itself.

Modification:

Adjust the open semantics to include the possibility of a mover being
cancelled.  This is represented by an InterruptedIOException.

Result:

A pool no longer disables itself if the client aborts a transfer shortly
after initiating it.

Target: master
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16
Require-notes: yes
Require-book: no
Ticket: http://rt.dcache.org/Ticket/Display.html?id=9269
Patch: https://rb.dcache.org/r/10498/
Acked-by: Tigran Mkrtchyan

Conflicts:
	modules/dcache/src/main/java/org/dcache/pool/movers/AbstractMover.java

Conflicts:
	modules/dcache/src/main/java/org/dcache/pool/movers/MoverChannelMover.java

### DIFF
--- a/modules/dcache-nfs/src/main/java/org/dcache/chimera/nfsv41/mover/NfsMover.java
+++ b/modules/dcache-nfs/src/main/java/org/dcache/chimera/nfsv41/mover/NfsMover.java
@@ -21,6 +21,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.io.InterruptedIOException;
 import java.nio.channels.CompletionHandler;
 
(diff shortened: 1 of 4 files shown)


jstarek:
retest this please

PR #3590, opened 2017-10-10
dcap: add TTL information to dcap messages

paulmillar:
Motivation:

The dcap door implements asynchronous requests and includes support for
internal retries.  This means that, if the dcap-door timer indicates
a problem (e.g., a message being lost), the door will resend the message.

PoolManager will (by default) suspend requests for a data stored only on
offline pools.  The dcap door's internal retry will result in multiple
requests if the pool remains down for any appreciable length.
PoolManager will collapse all such requests into a single request
handler instance; so, apart from some additional memory overhead, this
does not present a problem.

The issue comes when the pool comes back up again.  PoolManager attempts
to deliver replies for all the requests.  If the pool has been offline
for some time, then some of clients will have disconnected, killing the
session-specific cell.

For these cells, the domain hosting the dcap door will be unable to
deliver the PoolMgrSelectReadPool response messages to the cell.  The
domain then generates an error message and delivers this back to
PoolManager.

Therefore, when a pool comes back up again, PoolManager becomes very
slow at responding while its message queue is swamped from failed
delivery messages, as it sends the backlog of PoolMgrSelectReadPool
replies.

Modification:

Include a TTL value in dcap messages to PnfsManager and PoolManager.

For PnfsManager, this will likely have no effect, but PoolManager will
remove expired messages periodically.  This will reduce the number of
replies PoolManager sends when the pool comes back up again.

Result:

Less impact from dcap users when an offline pool comes back up again.

Target: master
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16
Ticket: http://rt.dcache.org/Ticket/Display.html?id=9250

### DIFF
--- a/modules/dcache-dcap/src/main/java/diskCacheV111/doors/DCapDoorInterpreterV3.java
+++ b/modules/dcache-dcap/src/main/java/diskCacheV111/doors/DCapDoorInterpreterV3.java
@@ -29,6 +29,8 @@ import java.util.concurrent.ConcurrentMap;
 
 import javax.security.auth.Subject;
 
+import java.util.concurrent.TimeUnit;
+
 import diskCacheV111.poolManager.PoolSelectionUnit.DirectionType;
 import diskCacheV111.poolManager.RequestContainerV5;
 import diskCacheV111.util.AccessLatency;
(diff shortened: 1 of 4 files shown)


paulmillar:
retest this please

PR #3593, opened 2017-10-11
dcache-restful-api: change the way in which transfers and restores su…

alrossi:
…pport paging

Motivation:

The current way in which services returning large lists of data
(active transfers, staging requests) support paging (offset, limit)
is by caching snapshots.   This is done on a session-by-session basis.

This is not, however, scalable.  It is also unnecessary.

The aim of supporting paging this way was to prevent clients from
receiving listings where potential additions and deletions scramble
the indexing the client considers valid.  This can be done in a less
memory-unfriendly way, though it does place a slightly heavier burden
on the client.

Modification:

Since the data stored for these services is not randomly updated by
asynchrous puts, but is refreshed only periodically and all at once,
the underlying list during the interval between refreshes can be
considered a universal snapshot.  We capture its identity by
associating a UUID token with it.  At each refresh, this token
changes.

When a client requests data, it receives back the token of the
snapshot to which it corresponds.  The client can choose two
courses of action if it wishes to page through the data in chunks.

The first is to return the token to the server.  If the token
has changed (expired), the server then returns a null token with
an empty list and offsets 0, -1.  The client understands this
null token means the list has changed and can then call refresh
to get the up-to-date list.

The second option is simply never to pass the token back to the
server.  In this case, the server will use the underlying
list to process the request, including the offset into it.  This,
of course, could include new elements the client did not see or
be missing elements which were originally there, but the client
may consider this behavior more desirable than sending the user
back to offset 0 if the list has changed during navigation.

Some refactoring has been done to the RESTful API and services
to generalize and render more DRY the code with this shared behavior.
A new data object encapsulates the access to the stored snapshot.
It has been provided with a JUnit test.

Result:

No server-side maintenance/caching of session-based snapshots.

Target:  master
Request: 3.2
Require-notes: no
Require-book: no
Acked-by: Tigran


jstarek:
retest this please

PR #3611, opened 2017-10-19
nfs: change the way how directory cookies are generated

kofemann:
Motivation:
Directory cookies used by clients to identify a file in the directory
listing. They expected to be permanent and unique per file per directory.
As dCache uses file inumber as a directory cookie, in case of hard-links
in the same directory we get two files files with the same cookie. This
confuses a) TreeSet, as both entries are treated as the same, b) nfs client,
as it expects then to be unique.

Modification:
To avoid chimera db schema change, we introduce a cookie which are a combination
of file's inumber + names hash. This guarantees, that newly created entries will
have higher cookie number. To avoid negative numbers, the higher bit is
set to zero (0).

Result:
dCache correctly lists directory with hard links in it.

Acked-by: Paul Millar
Target: master, 3.2
Require-book: no
Require-notes: yes
(cherry picked from commit f2f9e624a6d0f97125059c0fb6830aca0552a010)


jstarek:
Can I suggest correcting the spelling of the method name? direcotryCookieOf -> directoryCookieOf ;-)

kofemann:
You should have commented on  it during  review. I will fix in master, but
backport make no sense.


-kofemann

jstarek:
Agreed. In it goes...

PR #3627, opened 2017-10-25
dcache-frontend: adjust REST API for Pool Info Resources

alrossi:
Motivation:

Inconsistent use of POST for methods to kill  movers and change pool mode.

Modification:

Use DELETE with path to kill movers.  Use PATCH with path and json request
to change pool mode.

NOTE:  I have decided to let the bulk operations go for the moment.

Result:

REST API adheres to best practice.

Target: master
Request: 3.2
Require-notes: yes
Require-book: yes
Acked-by: Paul


alrossi:
Note:  the @PATCH annotation does not seem to be available with this version of the jersey library, so I left @POST for the mode change.

PR #3628, opened 2017-10-25
systemd: adding a global dcache.service which pulls in dcache@*.service

calestyo:
Motivation:
It would be nice to have a “global” dcache.service, which can be used tostart/
stop/reaload/restart all the dcache@*.services (i.e. the single domains).

Also, it would be much better not to have to enable each dcache@*.service (which
all have different names) all the nodes of the cluster, just to get everything
started.
Enabling dcache.service can be used to have all dCache domains started, while
one can still fall back to only enable single dcache@*.service (and leave
dcache.service on a node).

Modification:
Introduce such a dcache.service unit, and modify the generated units, so that
they're “pulled in”.

Target: master
Require-notes: yes
Require-book: yes
Request: 3.2

### DIFF
--- a/skel/lib/systemd/system-generators/dcache-generator
+++ b/skel/lib/systemd/system-generators/dcache-generator
@@ -4,6 +4,10 @@ set -e
 
 @DCACHE_LOAD_CONFIG@
 
+
+wantdir="$1/dcache.service.wants"
+mkdir "$wantdir"
+
 for domain in $(getProperty dcache.domains); do
     RESTART_DELAY="$(getProperty dcache.restart.delay "$domain")"
     USER="$(getProperty dcache.user "$domain")"
@@ -16,6 +20,9 @@ for domain in $(getProperty dcache.domains); do
 	[Unit]
 	Description=dCache $domain domain
 	After=network.target zookeeper.service
+	PartOf=dcache.service
+	ReloadPropagatedFrom=dcache.service
+	Before=dcache.service
 
 	[Service]
 	Type=simple
@@ -35,4 +42,6 @@ for domain in $(getProperty dcache.domains); do
 	[Install]
 	WantedBy=default.target
 	EOF
+
+    ln -s "$1/dcache@$domain.service" "$wantdir/dcache@$domain.service"
 done
--- a/skel/lib/systemd/system/dcache.service
+++ b/skel/lib/systemd/system/dcache.service
@@ -1,0 +1,14 @@
+#systemd service for managing all dCache domains on the system. This service is
+#actually a systemd target, but as targets cannot be reloaded a service is used.
+
+[Unit]
+Description=dCache
+
+[Service]
+Type=oneshot
+ExecStart=/bin/true
+ExecReload=/bin/true
+RemainAfterExit=on
+
+[Install]
+WantedBy=default.target


calestyo:
Note that I have no idea how your Java-ish build systems work... and whether the newly added `skel/lib/systemd/system/dcache.service` needs to be added somewhere to get into the packages ;-)

skel/lib/systemd/system/dcache.service
```diff
@@ -0,0 +1,14 @@
+#systemd service for managing all dCache domains on the system. This service is
+#actually a systemd target, but as targets cannot be reloaded a service is used.
+
+[Unit]
+Description=dCache
+
+[Service]
+Type=oneshot
+ExecStart=/bin/true
```
vingar: Shouldn't it be the dcache start command  ? same comment for below

calestyo: No, the "global" unit, will pull in the instance units (i.e. the dcache@domain.service units)...Actually in my opinion, dcache start/stop should eventually fully go away (and at least on systemd system already start to give warnings that it will be phased out).

vingar: yes you're right and my tests to validate the pull requests succeeded: sudo systemctl start  dcache.service ; sudo systemctl status dcache.service; sudo systemctl stop dcache.service

calestyo: I already tried to start that in the other issue,... I think dCache developers should decide on how they want to implement systemd support,... i.e. properly or just making it work somehow.

The proper way likely implies that some things dCache does now itself (e.g. the settings mentioned in the other ticket, or dcache start/stop) is mostly moved out of dCache itself.
For as long as we need to support SL6 I'd suggest to rather make wrappers for that, than to hold development towards the direction where things will sooner or later anyway go to. :)

skel/lib/systemd/system-generators/dcache-generator
```diff
@@ -35,4 +42,6 @@ for domain in $(getProperty dcache.domains); do
 	[Install]
```
gbehrmann: Can you please remove the `[Install]` section at the same time? Since you cannot 'enable' a generated unit, there is no point in having this section. The symlink you add below is the substitute for this.

skel/lib/systemd/system-generators/dcache-generator
```diff
@@ -35,4 +42,6 @@ for domain in $(getProperty dcache.domains); do
...
 	WantedBy=default.target
 	EOF
+
+    ln -s "$1/dcache@$domain.service" "$wantdir/dcache@$domain.service"
```
gbehrmann: I wouldn't mind if you renamed the generated unit to use a dash rather than an at symbol. As you pointed out, the purpose of the at symbol is different from what we do here.

gbehrmann:
I will leave the actual merge to @vingar or somebody else still being payed to work on dCache. I suggest this gets merged before the next bug fix release though.

skel/lib/systemd/system-generators/dcache-generator
```diff
@@ -4,6 +4,10 @@ set -e
 
 @DCACHE_LOAD_CONFIG@
 
+
+wantdir="$1/dcache.service.wants"
+mkdir "$wantdir"
```
gbehrmann: You may want to add a `-p` to this.

skel/lib/systemd/system-generators/dcache-generator
calestyo: Uhm... at least the Debian postgresql guys have it in their instance unit, too (even though ours isn't a true instance unit).

Further, this should go in a separate commit, I think... i.e. one that cleans up the generated units... these commits are just about adding the global one.

skel/lib/systemd/system-generators/dcache-generator
calestyo: While this would make life possibly easier for us for now... I don't think this is the right way to go.

We should in fact have instance units here... as all the domains are instances of dCache (just doing different things).
We should IMO rather go the hard way, pull out the settings from dCache which are anyway not ought to be there (in the systemd world) and then do it with instance units and "@".

skel/lib/systemd/system-generators/dcache-generator
calestyo: Debian had that as well, IIRC, but I'm sceptical.... the dir should already be there and if not, something else is wrong in the system,... and we shouldn't try to hide the other problem in that case.

gbehrmann: You mean should NOT be there? Because if it already exists from a previous run, then your command would fail. That's why I suggested adding `-p`. Are you certain that systemd deletes the dir before rerunning the generator after a daemon reload?

skel/lib/systemd/system-generators/dcache-generator
gbehrmann: Sure sure - but that's not going to fix the issue for 3.2 right now. As I said, I don't care much what you do in 3.3. We need a fix for 3.2 now. I realise it is my fault that what we got doesn't work. I would appreciate if you could help with this simple fix for 3.2 now - what you do in the future is up to you.

I am not trying to argue anything you are saying about what we should do in the next release. I just want a solution for the issue that right now prevents update to 3.2 on Debian.

skel/lib/systemd/system-generators/dcache-generator
gbehrmann: The postgresql guys do not generate a dynamic unit - they just add the symlinks. It does make sense to have it in `postgresql@.service` and in `dcache@.service` since these are not dynamically generated and thus can be 'enabled'.

Leave it in if you like - it will not allow you to enable the generated units though, but It will not cause harm either.

skel/lib/systemd/system-generators/dcache-generator
calestyo: I meant, the parent dirs should already be there (created by systemd)... AFAIU, the dir itself (i.e. the wantdir) should NOT be there cause my understanding was, that whenever the generators run, the whole /run/.../ stuff is flushed before.

If it was not (or if there was some other situation, in which the generators run, but the whole stuff isn't flushed away before), then I think it would be actually good for us to notice the failure. Why?
Because we "depend" on the flushing to get rid of e.g. de-configured domains in such cases.

This (having the clean up of the generated stuff) is also the reason why I want dCache's cached config to move away from /var/lib/dcache/config ...
Cause *if* we shall make use of generated per-domain EnvironmentFile= ... than these also need to be cleaned up at least from eventually (not necessarily on any system reload though)... having that stuff in /run/ which is typically tmpfs, makes me feel better ;-)

gbehrmann: Yeah, of course the parent directory is there. As long as you are certain the wantdir does not exist, then fine with me. What I was not certain about was whether the flushing would just delete the files/links or also the directories.

And if you want to move the config cache, be my guess. That stuff was written many years ago.

skel/lib/systemd/system-generators/dcache-generator
calestyo: The fix for 3.2 is simply to get the global dcache.service file... this gives people back the functionality they had before: have everything started or nothing.
It just lacks what's advertised for in the release notes (that one can selectively enable/disable domains).

I don't think it's your fault... I think your solution is actually pretty elegant, and I have no idea why systemd puts obstacles in the way not supporting this :-(

And I don't argue either :-) I'm just trying to find out which is the way we go for 3.3.
I think for 3.2 the best thing is to simply merge the global unit file... and remove the advertisement for per-domain enable/disable in the release notes.

gbehrmann: You mean merge the global unit file without updating the generator to create the symlinks? How would that work?

skel/lib/systemd/system-generators/dcache-generator
calestyo: You are right.

I leave it for now, and let's clean it up once we know where we'll eventually go with the whole thing.

skel/lib/systemd/system-generators/dcache-generator
calestyo: No, sorry, I wasn't clear:
I meant "for 3.2: merge this very pull request"... which gives the global unit file, and the necessary changes to the generator

skel/lib/systemd/system-generators/dcache-generator
calestyo: Well at least it works for me on 3.2 (on a test dcache)... even when reloading systemd multiple times.
I have no strong opinion about the -p ... if someone wants it ... I'd add it... if not the benefit is that we notice if it doesn't get flushed as we assume anyway :-)

As for the moving of the config cache... I tried to... but this seems to be coded in many places in dCache... so it's not done by changing path.properties... but I guess one needs to hack it into the build system.. and I didn't dare to touch this O:-)

paulmillar:
Chris, could you rebase this patch against the current tip of master?

calestyo:
sure & done

paulmillar:
Thanks!

vingar:
Ulf Tigerstedt did the validation and confirmed that's working. This can go in the next patch releases!

calestyo:
Thanks. You guys should still have a look at how we proceed with the topics/questions/issues brought up in #3630 ... even though this PR here is also kind of a workaround for that.

jstarek:
We'll do that, but at the moment, I still wait for a VM for the testing to be commissioned. -- Could you please provide us with a PR against 3.2 since you requested backporting? Thanks!

calestyo:
See #3656.

PR #3631, opened 2017-10-27
systemd: do not explicitly set StandardOutput=

calestyo:
StandardOutput= defaults to DefaultStandardOutput, which already defaults to
journal.
It doesn't make that much sense to explicitly set it here again (to make sure
it's always "journal") and most services I've seen here on Debian don't
(explicitly) set it either.

If we would want to set it explicitly, we should also do so for
StandardError=inherit.

Target: master
Request: 3.2
Require-notes: yes
Require-book: no

### DIFF
--- a/skel/lib/systemd/system-generators/dcache-generator
+++ b/skel/lib/systemd/system-generators/dcache-generator
@@ -26,7 +26,6 @@ for domain in $(getProperty dcache.domains); do
 	ProtectSystem=full
 	ProtectHome=true
 	NoNewPrivileges=true
-	StandardOutput=journal
 	SyslogIdentifier=dcache@$domain
 	Environment="CLASSPATH=$CLASSPATH" "LD_LIBRARY_PATH=$JAVA_LIBRARY_PATH"
 	$( [ -z "$USER" ] || echo "User=$USER" )


calestyo:
If this patch is accepted, the release notes should probably get a note, that DefaultStandardOutput will now be used, which, if unmodified, results in no change.

calestyo:
anything that prevents this from being merged?

vingar:
At NDGF,  We have this version: 
   https://gist.github.com/vingar/de943f2c285943b35491d7d49c942c00

Which includes some differences:
   https://gist.github.com/vingar/5801f37c80910bbe3c4b0062912f9dd3

Should we consider merging them in this pull request ?

cheers, Vincent

vingar:
ah I also noticed some differences with your version against the one on master:
6a7,10
> 
> wantdir="$1/dcache.service.wants"
> mkdir "$wantdir"
> 
18a23,25
> 	PartOf=dcache.service
> 	ReloadPropagatedFrom=dcache.service
> 	Before=dcache.service
36a44,45

calestyo:
Well these are simply because this PR hasn't been merged since more than two weeks and it's based on a older master.

If you look at the diff of this commit, only one line is removed, if that change would be accepted, I can rebase the commit, though that's probably not even necessary.

vingar:
yes, you're right. 
About this, all services systemd starts have their stdout/stderr connected to journald, thus journald really needs to run. It should be well documented and advertised.  I got few comments from Ulf that dcache status still claims it logs to /var/log/dcache/ and access and event logs still go to there.  We should also check how the logs from pool are collected.

calestyo:
I don't think we need to document this: this is the default behaviour of systemd, and it's the default behaviour in Debian as well.
Plus the default behaviour of Debian is, that it anyway picks up everything from journald in rsyslog (I guess everything goes to /var/log/syslog).

And as for Ulfs comments… it should in fact still log to /var/log/dcache, this is what our rsyslog.conf snippet does.

Of course the output of dcache status is wrong in the sense that it (presumably) shows just the value of `dcache.log.dir` which may or may not be correct in systemd (as one may change the rsyslog config, or completely disable that and just use journald).

This is another example of config option that IMHO should (canonically) move out to systemd. In the systemd world, dCache itself shouldn't care at all, where it's logging goes, it should just write to stdout/err and systemd do the rest. (In fact it does that already).
The `dcache` program should probably simply not show that column anymore, as it's beyond dCache's scope where the logging goes to.

The remaining problem is just, that dCache still does (and probably will continue) to use that setting for the event and access logs.

So this is just another one of the questions like what should happen to dcache.user, that the developers need to decide about. Once that's done I can help with patches.

vingar:
Could you post somewhere your rsyslog configuration ? I'll discuss your (more than valid) other points with the developers. One issue will be to support non systemd platform for a while.

Thanks again for the contribution !

calestyo:
> Could you post somewhere your rsyslog configuration?

https://github.com/dCache/dcache/blob/master/packages/fhs/src/main/scripts/dcache.rsyslog

vingar:
sorry it's not related to this pull request, but: 
did you do something to enable the journald to rsyslog communication ? like imjournal in /etc/rsyslog.conf. ForwardToSyslog=yes in journald.conf ? It would be good the describe the recipe somewhere gist ?

calestyo:
`ForwardToSyslog=` is enabled by default on Debian, and I think on upstream systemd as well, see journald.conf(5):

> ForwardToSyslog=, ForwardToKMsg=, ForwardToConsole=, ForwardToWall=

[snip snap]

>            If
>            forwarding to syslog is enabled but nothing reads messages from the
>            socket, forwarding to syslog has no effect. By default, only
>            forwarding to syslog and wall is enabled.

PR #3636, opened 2017-10-29
Improve logging

calestyo:


kofemann:
👍 Thanks.

calestyo:
Thanks for merging... does anyone of you guys perhaps know, how one can get rsyslog to change the program name *in* the logfile from dcache@domain to just domain... without setting a complete format string template for the message... i.e. "just replace the program name, and leave the log message as is otherwise"?

Cause that would be the next step to make the system logging fully look like the "old" logging.

PR #3653, opened 2017-11-02
Fix chmods to fail silently if the directories are missing

tiggi:
Small fix to make the .deb packages install properly.
The /var/lib/dcache/cell-info directory isn't created when the package is installed, so the postinst-script will fail.

Untested releases are fun!

### DIFF
--- a/packages/fhs/src/main/deb/postinst
+++ b/packages/fhs/src/main/deb/postinst
@@ -34,18 +34,18 @@ Please fix this and reinstall this package." >&2
     chown dcache:dcache /var/spool/dcache/star
 
     # protect directories that only dCache should access
-    chmod 700 /var/lib/dcache/alarms
-    chmod 700 /var/lib/dcache/cell-info
-    chmod 700 /var/lib/dcache/credentials
-    chmod 700 /var/lib/dcache/httpd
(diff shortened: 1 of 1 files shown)


paulmillar:
We certainly test dCache quite thoroughly (after each commit, even), but sadly only on CentOS machines, so this problem fell through the gap.  We'll add a Debian machine into the mix to try to catch these problems earlier.

With the patch, an alternative approach is to include the empty directory in the package.  I've created a patch to do this.

tiggi:
ok.

Please add a systemd startup script too. Having to manually start dcache when you restart the computer gets tedious..

paulmillar:
OK, will do -- hopefully for next week's release cycle (Tuesday), but if not, then certainly for the next one.

calestyo:
btw: I don't think that the approach of this PR is the right one.

If we don't fail when the dirs are missing, that means that they won't get the proper permissions.
The user may later create them manually, though...

IMO, the proper solution is simply to create all dirs as required.

jstarek:
Since we have had another pull request that adresses the issue in a more robust manner, I'll close this one. Thanks a lot for the submission, though!

PR #3656, opened 2017-11-08
systemd: adding a global dcache.service which pulls in dcache@*.service

calestyo:
Motivation:
It would be nice to have a “global” dcache.service, which can be used tostart/
stop/reaload/restart all the dcache@*.services (i.e. the single domains).

Also, it would be much better not to have to enable each dcache@*.service (which
all have different names) all the nodes of the cluster, just to get everything
started.
Enabling dcache.service can be used to have all dCache domains started, while
one can still fall back to only enable single dcache@*.service (and leave
dcache.service on a node).

Modification:
Introduce such a dcache.service unit, and modify the generated units, so that
they're “pulled in”.

Target: master
Require-notes: yes
Require-book: yes
Request: 3.2

### DIFF
--- a/skel/lib/systemd/system-generators/dcache-generator
+++ b/skel/lib/systemd/system-generators/dcache-generator
@@ -4,6 +4,10 @@ set -e
 
 @DCACHE_LOAD_CONFIG@
 
+
+wantdir="$1/dcache.service.wants"
+mkdir "$wantdir"
+
 for domain in $(getProperty dcache.domains); do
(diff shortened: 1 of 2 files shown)


jstarek:
Thanks!

calestyo:
could you please also have a look that the two commits from #3636 make it into the next 3.2 patch release?

jstarek:
@calestyo I can create the PR myself, but I'd actually prefer if you posted a PR against 3.2 since that would be a better fit for our normal review workflow.

calestyo:
see #3657

PR #3668, opened 2017-11-12
maven: include dcache.service in Debian packages

calestyo:
Apparently the build system doesn't just pick up the skeleton directories and
include them in the packages (not sure why, cause this is what it should
probably just do).
This causes the new global dcache.service from commit
3bc9c497f194cfdcdf0515775d538ec53967685e to be missing from the built packages.

This commit tells it to at least pick up the whole “skel/lib/systemd/” instead
of just “skel/lib/systemd/system-generators/”.

Target: master
Require-notes: yes
Require-book: no
Request: 3.2

### DIFF
--- a/packages/fhs/src/main/assembly/deb.xml
+++ b/packages/fhs/src/main/assembly/deb.xml
@@ -8,8 +8,8 @@
             <filtered>true</filtered>
         </fileSet>
         <fileSet>
-            <directory>${filtered-skel}/lib/systemd/system-generators</directory>
-            <outputDirectory>lib/systemd/system-generators</outputDirectory>
+            <directory>${filtered-skel}/lib/systemd</directory>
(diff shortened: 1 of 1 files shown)


vingar:
my packaging tests on debian confirmed that it works. the dcache service is not enabled by default after the deb installation but it' fine.

calestyo:
Well _if_ one would want that to be auto-enabled, than this would go in the maintainer scripts.

Debian does it typically for most daemons - but I think it's generally a bad idea, and especially for dCache as this doesn't run in some useful default mode, if no config is in place.

calestyo:
btw: can we have an out-of-the-row minor release of dCache with this fix?
3.2.5 was mostly to get the Debian package working again, but this commit is still missing for that... (and I'd like to upgrade our production dCache from 3.1 -> 3.2 this or next week)

PR #3730, opened 2017-11-26
overhaul logroation on Debian

calestyo:
The current logrotate script has some issues and space for improvements.

First, there is no reason, that dCache sets “weekly”, “compress”, “rotate 52”
(especially why 52?) - these should be taken from the global settings, so an
admin can easily define when he wants rotation and whether and which compression
he wants to have used (e.g. not the standard poor gzip that “compress” implies
but maybe XZ).
If an admin wants to override the global defaults, he still can do so in the
logration configuration for dCache.

Second, in the default configuration, log messages are written to stdout/err,
from where they get collected by journald, from where rsyslogd picks them up and
writes them into files in “/var/log/dcache”.
Because rsyslogd now manages “/var/log/dcache/*.log” we no longer need to use
the “copytruncate”, which is prone to loosing messages (between the copy and the
truncate). Instead, we let rsyslog rotate the file (and it’s important to not
have it compressed already here, so “delaycompress” is set), “create” a new log
file and then we send HUP to rsyslog (or rather some Debian script does this the
correct way for us) to have it opening the new log files.
Thus we shouldn’t loose any messages (and it’s faster too).

“missingok” is left in place, simply because the package doesn’t (and cannot
reliably) create the log files.

I’ve also added “notifempty” which is for some reason not the global default in
Debian. I know, this is the same evil than overriding the global settings with
e.g. “weekly” but I bet that simply everyone wants to have this.

Last but not least I’ve added some warnings that this may possibly not work for
other non-rsyslogd-managed log files (e.g. the access and event logs).

Target: master
Require-notes: yes
Require-book: possibly
Request: 3.2

### DIFF
--- a/packages/fhs/src/main/deb/dcache.logrotate
+++ b/packages/fhs/src/main/deb/dcache.logrotate
@@ -1,7 +1,26 @@
+#Note: The following may work only for dCache logs managed by rsyslog (i.e. not for its access- and event logs, which are managed by logback)!
+
+
+
+
 /var/log/dcache/*.log {
-    compress
-    weekly
-    rotate 52
-    missingok
-    copytruncate
+	create
+	nocopy
+	nocopytruncate
+	missingok
+	
+	#Note: “delaycompress” is important for how the rsyslog-managed logs are rotated.
+	delaycompress
+	
+	postrotate
+		if [ -x /usr/lib/rsyslog/rsyslog-rotate ]; then
+			/usr/lib/rsyslog/rsyslog-rotate
+		else
+			#TODO: Remove this (and the surrounding “if”) once rsyslog versions < 8.27.0-4 are no longer in use (which is still the case in Debian stretch).
+			invoke-rc.d rsyslog rotate > /dev/null
+		fi
+	endscript
+	
+	
+	notifempty
 }


calestyo:
Would be great if some minion O;-) (or a nice developer) could verify whether this actually works (I don't have a test system for this right now)... i.e. start dCache, let it write some log messages, call logrotate, see wether new messages go to the new file.

Other than that I simply proclaim it works™ without having it tested so far :-D.

calestyo:
Uagh… wait with merging that... just noticed that /usr/lib/rsyslog/rsyslog-rotate is only available in Debian buster… so I'll have to amend the commit.

calestyo:
I've updated the commit to work with older Debians… and also tested it on our production dCache... seems to work fine.

Could be merged.

jstarek:
Hi, sorry for the delay in reviewing this. First, thanks for pointing out the issues with our current logrotate config, like the copytruncate issue. We'd like to ship a more robust config here.

After talking to some people from dCache operations and development, I think we'd like to see this approached a bit differently, however.

- First, wrt your comment about "weekly" and "rotate 52": We feel that moving options like rotation frequency, retention time etc. to dCache-internal config options is not really a good fit. This is about how a service behaves in a normal OS environment, so we'd prefer keeping these things separate -- those items should stay in the logrotate config. And, sure, gzip is a poor choice, but available everywhere. Which is what we prefer since,

- secondly, we feel that this file is (and indeed should be) a rather trivial placeholder for whatever configuration is really needed at sites. So, do you think the postrotate script section could be simplified?

calestyo:
> First, wrt your comment about "weekly" and "rotate 52": We feel that moving options like rotation frequency, retention time etc. to dCache-internal config options is not really a good fit. This is about how a service behaves in a normal OS environment, so we'd prefer keeping these things separate -- those items should stay in the logrotate config. And, sure, gzip is a poor choice, but available everywhere. Which is what we prefer since,

Not sure what you mean here... I never said the rotation frequency should become part of dCache… actually I said it shouldn't even be part of the logrotate config that dCache ships.

Same for gzip... dCache shouldn't mandate this... it should be the global logrotate config from which this is inherited.

The dcache logrotate snippet should only contain tuff necessary for the dcache specifics (i.e. the HUP thingy)... and not override the global admin choices.
This is beneficial in another way, in that people who want to use different settings (e.g. not gzip) don't have to adapt the config each upgrade.


Don't think it can be simplified that much… effectively it's a one liner, which needs the if and the alternative way of handling it, for older debian versions.. once these become unsupported, the extra stuff goes away.

packages/fhs/src/main/deb/dcache.logrotate
```diff
@@ -1,7 +1,26 @@
+#Note: The following may work only for dCache logs managed by rsyslog (i.e. not for its access- and event logs, which are managed by logback)!
+
+
+
+
```
vingar: empty lines

vingar:
I've deployed this logrotate script on the ndgf pre-pro instance for further validation. I think two logrotate scripts will be necessary imo as this one assume sjournalctl, systemctl, etc which is not guarantee to be there on all instances. @calestyo What do you think ?

packages/fhs/src/main/deb/dcache.logrotate
calestyo: Well I think such config files shouldn't be possible candidates for the IOCCC ;-)
The comment in the beginning would apply to any logrotate section thay may be added sooner or later to this file (right now only `/var/log/dcache/*.log {}`)... but I think the general goal should be to get all logs rotated by the standard systemtools (systemd/journald and/or logrotate.

So I've added some empty lines that this comment is actually noticed and to indicate it applies to all the following an not just the next section.

packages/fhs/src/main/deb/dcache.logrotate
```diff
@@ -1,7 +1,26 @@
...
-    rotate 52
-    missingok
-    copytruncate
+	create
+	nocopy
+	nocopytruncate
+	missingok
+	
+	#Note: “delaycompress” is important for how the rsyslog-managed logs are rotated.
+	delaycompress
```
vingar: both `delaycompress` and `compress` need to be there according to the man page.

packages/fhs/src/main/deb/dcache.logrotate
```diff
@@ -1,7 +1,26 @@
+#Note: The following may work only for dCache logs managed by rsyslog (i.e. not for its access- and event logs, which are managed by logback)!
+
+
+
+
 /var/log/dcache/*.log {
-    compress
-    weekly
-    rotate 52
```
vingar: I would keep both `weekly` and `rotate 52` for backward compatibility

packages/fhs/src/main/deb/dcache.logrotate
```diff
@@ -1,7 +1,26 @@
...
+		if [ -x /usr/lib/rsyslog/rsyslog-rotate ]; then
+			/usr/lib/rsyslog/rsyslog-rotate
+		else
+			#TODO: Remove this (and the surrounding “if”) once rsyslog versions < 8.27.0-4 are no longer in use (which is still the case in Debian stretch).
+			invoke-rc.d rsyslog rotate > /dev/null
+		fi
+	endscript
+	
+	
+	notifempty
```
vingar: better to move `notifempty` up and have the postrotate block at the end.

vingar:
few comments and one remark: with rsyslog now, we have a mixture of log files created by the syslog:sylog user and the dcache:dcache ones to get the right permissions we needed few (nightmare) changes. I was wondering how did you deal with this aspect ?

jstarek:
Closing due to long inactivity now, just to keep our PR list shorter. We do have full systemd support on the agenda and will change the rotation concept during that work. Feel free to re-open when the PR has been updated and the outstanding issues resolved.

PR #3744, opened 2017-11-29
maven: fix dcache.service permissions

calestyo:
This fixes my commit 62ca394df9147a0e900f5047114abeaccfb26143 in which I didn’t
notice that maven sets the file permission bits of all the files of that set to
0755.

systemd however complains about unit files that have the executable bit set
with:
>systemd[1]: Configuration file /lib/systemd/system/dcache.service is marked
>executable. Please remove executable permission bits. Proceeding anyway.

Target: master
Require-notes: no
Require-book: no
Request: 3.2

### DIFF
--- a/packages/fhs/src/main/assembly/deb.xml
+++ b/packages/fhs/src/main/assembly/deb.xml
@@ -8,8 +8,14 @@
             <filtered>true</filtered>
         </fileSet>
         <fileSet>
-            <directory>${filtered-skel}/lib/systemd</directory>
-            <outputDirectory>lib/systemd</outputDirectory>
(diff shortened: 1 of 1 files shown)


calestyo:
built the package with maven, and it seems the perms are now fine

kofemann:
Thanks!

kofemann:
@calestyo we have 4.0 now. Our policy is to backport to all intermediate branches as well, IOW, if you think it should go into 3.2, then it have to go into 4.0 as well.

calestyo:
sure... 4.0 wasn't out when I made the PR

Issue #3738, opened 2017-11-28
DB upgrade failure


calestyo:
Hey.

We're just in downtime upgrading from 3.1.14-1 -> 3.2.7-1.

`dcache database upgrade` brought this:
```
PnfsManager@filesystem: 
INFO  - Successfully acquired change log lock
INFO  - Reading from public.databasechangelog
INFO  - org/dcache/chimera/changelog/changelog-master.xml: org/dcache/chimera/changelog/changeset-2.15.xml::1.1::behrmann: New row inserted into t_inodes
ERROR - org/dcache/chimera/changelog/changelog-master.xml: org/dcache/chimera/changelog/changeset-2.15.xml::1.1::behrmann: Change Set org/dcache/chimera/changelog/changeset-2.15.xml::1.1::behrmann failed.  Error: org.postgresql.util.PSQLException: ERROR: column "ipnfsid" of relation "t_dirs" does not exist
  Position: 44
liquibase.exception.DatabaseException: org.postgresql.util.PSQLException: ERROR: column "ipnfsid" of relation "t_dirs" does not exist
  Position: 44
	at liquibase.executor.jvm.JdbcExecutor$ExecuteStatementCallback.doInStatement(JdbcExecutor.java:316) ~[liquibase-core-3.3.2.jar:na]
	at liquibase.executor.jvm.JdbcExecutor.execute(JdbcExecutor.java:55) ~[liquibase-core-3.3.2.jar:na]
	at liquibase.executor.jvm.JdbcExecutor.execute(JdbcExecutor.java:122) ~[liquibase-core-3.3.2.jar:na]
	at liquibase.database.AbstractJdbcDatabase.execute(AbstractJdbcDatabase.java:1227) ~[liquibase-core-3.3.2.jar:na]
	at liquibase.database.AbstractJdbcDatabase.executeStatements(AbstractJdbcDatabase.java:1210) ~[liquibase-core-3.3.2.jar:na]
	at liquibase.changelog.ChangeSet.execute(ChangeSet.java:550) ~[liquibase-core-3.3.2.jar:na]
	at liquibase.changelog.visitor.UpdateVisitor.visit(UpdateVisitor.java:43) [liquibase-core-3.3.2.jar:na]
	at liquibase.changelog.ChangeLogIterator.run(ChangeLogIterator.java:73) [liquibase-core-3.3.2.jar:na]
	at liquibase.Liquibase.update(Liquibase.java:200) [liquibase-core-3.3.2.jar:na]
[...]

PR #3748, opened 2017-11-29
chimera: correct previous attempt to fix 'lost+found' directory permi…

paulmillar:
…ssion

Motivation:

A previous patch attempted to fix the problem that the 'lost+found'
directory was created with the wrong permissions.  That patch was faulty
for two reasons:

  1. if the 'lost+found' directory was deleted then the new
     changeset would attempt to recreate it, using the wrong
     schema.

  2. if the 'lost+found' directory was not deleted, no attempt is made
     to fix the permissions.  This is unfortunate as the directory is
     not shown in NFS, due to the incorrect permissions.

Modification:

The old changeset is reverted and a new changeset is introduced.  This
checks that the directory exists and that is has the incorrect
permissions before attempting to correct the permissions.

Result:

The 'lost+found' directory permissions is updated without causing
problems if that directory has been removed or permissions have been
modified.

Target: master
Request: 4.0
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16
Closes: #3738
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/10655/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/chimera/src/main/resources/org/dcache/chimera/changelog/changeset-2.15.xml
+++ b/modules/chimera/src/main/resources/org/dcache/chimera/changelog/changeset-2.15.xml
@@ -5,7 +5,8 @@
         xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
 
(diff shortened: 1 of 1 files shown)


paulmillar:
retest this please

PR #3751, opened 2017-11-29
systemd: stop all dCache service units in Debian prerm

calestyo:
The Debian package’s prerm maintainer script only stopped “dcache@*” units.

First, I think it’s better to really explicitly stop “dcache@*.service” as these
are the only ones we create at the moment and why should we stop anything of
similar names that the user may have created himself.

Second, it didn’t stop the later introduced “dcache.service”. I forgot to add
this, cause normally on proper Debian source packages this would be added
automatically by Debhelper.

Target: master
Request: 3.2
Require-notes: no
Require-book: no


kofemann:
thanks!

Issue #3752, opened 2017-12-04
NPE on pool startup


kofemann:
If you type 'info' command just after pool started, then NPE is logged:

3:45:36 PM [dcache-lab002-A-1] [admin] Command failed due to a bug, please contact support@dcache.org.dmg.util.CommandPanicException: (1) Command failed: java.lang.NullPointerException
	at org.dcache.util.cli.AnnotatedCommandExecutor.execute(AnnotatedCommandExecutor.java:156) ~[common-cli-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at dmg.cells.nucleus.CellAdapter.executeCommand(CellAdapter.java:238) ~[cells-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at org.dcache.cells.UniversalSpringCell.executeCommand(UniversalSpringCell.java:195) ~[dcache-core-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at dmg.cells.nucleus.CellAdapter$1.doExecute(CellAdapter.java:104) ~[cells-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at org.dcache.util.cli.CommandInterpreter.command(CommandInterpreter.java:129) ~[common-cli-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at dmg.cells.nucleus.CellAdapter$1.command(CellAdapter.java:121) ~[cells-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at dmg.cells.nucleus.CellAdapter.command(CellAdapter.java:223) ~[cells-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at dmg.cells.nucleus.CellAdapter.executeLocalCommand(CellAdapter.java:915) ~[cells-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at dmg.cells.nucleus.CellAdapter.messageArrived(CellAdapter.java:855) ~[cells-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at dmg.cells.nucleus.CellNucleus$DeliverMessageTask.run(CellNucleus.java:1218) [cells-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
[...]

PR #3753, opened 2017-12-04
pools: fix NPE from info command at startup

alrossi:
Motivation:

order to support JSON representations.  The
refactoring neglected to include a previous
null check on one of the HsmFlushController fields.

Modification:

Restore the null check.

Result:

An NPE should not be thrown on startup if the
info command is run on the pool. (See below).

Target: master
Request: 3.3
Request. 3.2
Fixes: #3752
Acked-by: Tigran
Require-notes: yes
Require-book: no


alrossi:
3.3 is a mistake. Ignore

PR #3793, opened 2018-02-09
frontend: Map requests with bad JSON to HTTP 400 Bad Request status code

paulmillar:
Motivation:

Jackson is used by Jersey to deserialise JSON input by creating a object
of the right type/class.  If the supplied JSON input is somehow bad,
Jackson throws an exception.  These Jackson exceptions are currently
treated as bugs: a stack-trace is logged and the client sees a 500
Internal Error.

Modification:

Update custom error handling to work with Jackson exceptions.

Result:

Bad client-supplied JSON input results in a HTTP 400 status code,
instead of 500 status code.  The error message is hopefully more useful.

Target: master
Require-notes: yes
Require-book: no
Request: 4.0
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16
Patch: https://rb.dcache.org/r/10719/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache-restful-api/src/main/java/org/dcache/restful/providers/ErrorResponseProvider.java
+++ b/modules/dcache-restful-api/src/main/java/org/dcache/restful/providers/ErrorResponseProvider.java
@@ -17,15 +17,17 @@
  */
 package org.dcache.restful.providers;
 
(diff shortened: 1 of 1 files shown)


paulmillar:
retest this please

PR #3869, opened 2018-02-26
dcache-resilience: handle properly RuntimeExceptions from tasks

alrossi:
Motivation:

A RuntimeException may be lost; specifically, the error is not logged
and does not result in the task being removed from the queue.

Modification:

Add wrapper class to ensure that a RuntimeException is logged.  (This has been
turned into an abstract class, ResilientTask, to hold code in common.) The task
also reports this error back to the containing object so the status is updated.
This allows the existing task removal process to remove it from the list of running tasks.

Also applied, for consistency, to the BrokenFileTask, even though no special
handling of the error is required.

Result:

Bugs are handled correctly.

Target: master
Request: 4.0
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16
Require-notes: yes
Require-book: no
Acked-by: Paul

### DIFF
--- a/modules/dcache-resilience/src/main/java/org/dcache/resilience/data/FileOperation.java
+++ b/modules/dcache-resilience/src/main/java/org/dcache/resilience/data/FileOperation.java
@@ -391,6 +391,7 @@ public final class FileOperation {
 
     public synchronized void submit() {
         if (task != null) {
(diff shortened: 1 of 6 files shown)


jstarek:
retest this please

PR #3884, opened 2018-02-28
migration: Remove org.dcache.pool.migration.ReverseOrder

Laz0rSquid:
Motivation:

Java8 have introduced Comparator#reversed, we don't need our own class to do that.

Modification:

Removed ReverseOrder.java from the migration directory.
Modified code in MigrationModule.java to reflect the changes, instead of:
    return new ReverseOrder(new Order());
now it does:
    return new Order.reversed();

Result:

ReverseOrder.java is removed, code now uses the method of the standard Comparator library.

Target: master
Require-notes: no
Require-book: no
Signed-off-by: Marcus Kopp <marcus.kopp86@gmail.com>


kofemann:
retest this please

kofemann:
Thanks!

PR #3894, opened 2018-03-04
dmg.utils.Gate: Replace dmg.utils.Gate with java.util.concurrent.Sema…

Laz0rSquid:
…phore

Motivation:

When java standard library did not provide a semaphore-like abstraction, dCache had implemented it's own one. Since java.util.concurrent.Semaphore is available since java5, the custom implementation can be replaces with one which provided by JDK.

Modification:

Replaced Gate with Semaphore, deleted Gate.java and GateTest.java

Result:

Project now uses standard implementation of Semaphore

Fixes: #38228
Acked-by:
Target: master
Require-book: no
Require-notes: no


modules/cells/src/main/java/dmg/cells/nucleus/SystemCell.java
```diff
@@ -1,12 +1,15 @@
 package dmg.cells.nucleus ;
 
 import com.google.common.base.Throwables;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
+import dmg.util.AuthorizedString;
```
kofemann: please do not re-organize imports

modules/cells/src/main/java/dmg/cells/nucleus/SystemCell.java
```diff
@@ -66,15 +51,15 @@
...
+   private final Semaphore              _shutdownLock = new Semaphore(0);
 
    private class TheKiller extends Thread {
       @Override
       public void run(){
          _log.info("Running shutdown sequence");
          kill() ;
          _log.info("Kill done, waiting for shutdown lock");
-         _shutdownLock.check() ;
+         _shutdownLock.tryAcquire() ;
```
kofemann: you break the semantics here: Gate#check blocks until gate is open. Semaphore#tryAcquire returns right away with true/false result.

kofemann:
The contributor did not follow-up the patch. Closing.

PR #3895, opened 2018-03-07
dcache: avoid NPE from initialization race in RestoreRequestsReceiver…

alrossi:
… in HttpPooMgrEngineV3

Motivation:

During system test the following NPE was observed:

06 Mar 2018 10:57:32 (httpd) [PoolManager PoolManagerGetRestoreHandlerInfo] Uncaught exception in thread httpd-0
java.lang.NullPointerException: null
        at diskCacheV111.poolManager.RestoreRequestsReceiver.messageArrived(RestoreRequestsReceiver.java:120) ~[dcache-core-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_102]
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_102]
        ...

The NPE stems from the fact that initialization has not occurred before the
RequestHandler in PoolManager sends the message.

This seems to be particularly the case for the HttpPoolMgrEngineV3, which is
constructed out of band (not by Spring injection), and initialized in afterStart.

Modification:

Move the initialization to the constructor, since the thread is actually
started there.

Also, initialized the webapp and frontend instances in the spring context
and remove redundant initializa call in the class,
and eliminated the unnecessary synchronized block (the cache returns
a concurrent map, so no synchronization is needed).

Result:

We should not see NPEs from this race.  Also, initialization sequence
is better.

Target: master
Request: 4.0
Require-notes: yes
Require-book: no
Acked-by: Paul

### DIFF
--- a/modules/dcache-frontend/src/main/java/org/dcache/restful/util/restores/RestoreCollector.java
+++ b/modules/dcache-frontend/src/main/java/org/dcache/restful/util/restores/RestoreCollector.java
@@ -94,7 +94,6 @@ public class RestoreCollector extends
     public void initialize(Long timeout, TimeUnit timeUnit) {
         pnfsHandler = new PnfsHandler(pnfsStub);
         pnfsHandler.setSubject(Subjects.ROOT);
-        receiver.initialize();
         super.initialize(timeout, timeUnit);
     }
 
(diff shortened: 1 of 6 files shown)


mksahakyan:
retest this please

mksahakyan:
retest this please

PR #3900, opened 2018-03-08
dcache-frontend: avoid null dereferencing for incomplete pool history…

alrossi:
… data

Motivation:

If the history service is not running or there are other issues
with pool connectivity, this could lead to incomplete or missing
JSON data from the history service.

Currently there is a nonNull filter in place, but it is not
sufficient.  Each map operation which accesses the next
child element must do the same.

Modification:

Add the missing filters.

Also, a similar filter in a related class has been
changed from (e) -> e != null to Objects::nonNull
for the sake of consistency.

Result:

Peaceful behavior in the absence of data.

Target: master
Request: 4.1
Request: 4.0
Require-notes: yes
Require-book: no
Acked-by: Paul


mksahakyan:
retest this please

mksahakyan:
retest this please

mksahakyan:
retest this please

PR #3904, opened 2018-03-09
ftp: ensure half-closed connections are fully closed on return

paulmillar:
Motivation:

PassiveConnectionHandler mistakenly believes that the
SocketChannel#isInputShutdown method returns true whenever the remote
connection has closed its half of the TCP connection.

In fact, Java provides no mechanism to discover whether a TCP connection
is (half-) closed.  The *only* way to discover this is by reading from
the channel and receiving the end-of-stream marker (typically, -1 is
returned).

Currently, the client closing their connection results in a loop.  The
connection is returned to PassiveConnectionHandler and added to the
idle-connections list.  The selector marks the connection as active and
a corresponding handler is created only for this handler to discover the
connection is closed.  The handler then returns the connection to the
PassiveConnectionHandler.

Modification:

Mark remote-closed connections as such, by calling
SocketChannel#shutdownInput.

Result:

Remote closed connections are handled correctly.

An unreleased regression is fixed.

Target: master
Request: 4.1
Require-notes: no
Require-book: no
Patch: https://rb.dcache.org/r/10818/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache-ftp/src/main/java/org/dcache/ftp/proxy/EDataBlockNio.java
+++ b/modules/dcache-ftp/src/main/java/org/dcache/ftp/proxy/EDataBlockNio.java
@@ -66,10 +66,16 @@
 
 package org.dcache.ftp.proxy;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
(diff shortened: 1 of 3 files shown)


mksahakyan:
retest this please

PR #3941, opened 2018-04-09
pool: fix error message for failed active FTP transfers

paulmillar:
Motivation:

Active transfers (as opposed to passive) has dCache establish the TCP
connection with the data recipient, over which the file's content will
flow.  Under ideal circumstances, the dCache pool will establish this
connection, rather than the door establishing the connection and
proxying the data.  If the pool experiences a problem establishing the
TCP connection, an IOException is caught a propagated to the door,
although the message sent wraps the original IOException to provide
additional context information (e.g., address & port number).

Unfortunately, the IOExceptions from OpenJDK often fail to include a
message, resulting error responses like:

    451 General problem: Problem while connected to [2001:67c:1148:200::111]:20276: null

Modification:

Update wrap method to build a suitable message by using the cause class
simple name if the message is null: this is often enough to identify the
nature of the problem.

Result:

Provide a more complete error message should the pool fail to connect to
the remote party as part of an active (i.e., non-passive) FTP transfer.

Target: master
Require-notes: true
Require-book: false
Request: 4.1
Request: 4.0
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16

### DIFF
--- a/modules/cells/src/main/java/dmg/util/Exceptions.java
+++ b/modules/cells/src/main/java/dmg/util/Exceptions.java
@@ -58,14 +58,17 @@ public class Exceptions
      * enclosingType is reached.  If enclosingType does not support wrapping
      * then {@literal cause} is returned and a warning is logged.
      * <p />
-     * Note that the wrapped exception will contain only a message and possibly
-     * the {@literal cause} as the triggering exception; in particular, any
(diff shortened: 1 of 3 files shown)


paulmillar:
retest this please

PR #3966, opened 2018-04-19
ftp: returned error is too vague for meaningful investigation

paulmillar:
Motivation:

dCache will fail a command if a client requests a different IP version
(e.g., IPv4 instead of IPv6) and that interface has no such interface.

The returned error is so vague that the cause is completely unclear.

Modification:

Update error message to provide a reasonable summary of the problem.

Result:

Problems where a client requests an IP-version that isn't supported will
generate an error message that better describes the problem.

Target: master
Request: 4.1
Request: 4.0
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16
Ticket: http://rt.dcache.org/Ticket/Display.html?id=9388


paulmillar:
retest this please

PR #3978, opened 2018-04-24
dcache-resilience: fix computation of operation count when storage re…

alrossi:
…quirements change

Motivation:

    There is a small regression in the way resilience computes
    the number of operations necessary to adjust copies when
    a storage unit definition changes.  Currently it computes
    this as 2 X (required - 1).  But this fails when there
    is a reduction in the number of required copies.

Modification:

    The correct computation is:  max(required, current locations)
    + required - 2.  This ensures that all removals can take
    place.  (Note that if required >= current, the formula
    is identical to the previous one, whereas in the opposite
    case the count is larger than before; tag redistribution
    changes should thus be covered in any case).

Result:

    Correct number of operation passes occur during scans.

Target: master
Request: 4.1
Request: 4.0
Request: 3.2
Request: 3.1
Request: 3.0
Request: 2.16
Require-notes: yes
Require-book: no
Acked-by: Tigran


mksahakyan:
retest this please

PR #4000, opened 2018-05-28
no longer disable Java ECC algorithms

calestyo:
By now everyone should have been able to install a current version of JDK in
which ECC algorithms are properly supported.
There's no reason to keep the far more secure algorithms for anyone else per
default, just because a minority might not yet have upgraded.

Thus, changed the default of not using ECC algorithms.

Target: master
Request: 4.1
Request: 4.0
Require-notes: yes
Require-book: no


paulmillar:
Thanks Chris!

PR #4002, opened 2018-05-28
webdav: do not return 404 for OPTIONS request targeting absent entity

paulmillar:
Motivation:

Milton's OPTIONS handler returns a 404 error if an OPTIONS request
targets an entity that does not exist.  This behaviour deviates from
Apache httpd server and results in failed uploads for dcache-view.

Modification:

Introduce a work-around where non-existing Resource is handled inside
the ResourceHandlerHelper.

Milton uses the Resource to deduce the list of supported methods.  This
patch introduces a hard-coded list of supported methods for non-existing
files that matches what DcacheFileResource provides; however, future
work should use Milton's mechanism (e.g., by injecting a dummy
DcacheFileResource) to prevent the hard-coded list becoming outdated.

Result:

OPTIONS does not return a 404 Status code when targeting on non-existing
entity.

dcache-view uploads succeed.

Target: master
Request: 4.1
Request: 4.0
Request: 3.2
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/10977
Acked-by: Olufemi Adeyemi

### DIFF
--- a/modules/dcache-webdav/src/main/java/org/dcache/webdav/DcacheResourceHandlerHelper.java
+++ b/modules/dcache-webdav/src/main/java/org/dcache/webdav/DcacheResourceHandlerHelper.java
@@ -1,5 +1,6 @@
 package org.dcache.webdav;
 
+import com.google.common.collect.ImmutableList;
 import io.milton.http.AuthenticationService;
 import io.milton.http.HandlerHelper;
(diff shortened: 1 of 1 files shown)


paulmillar:
retest this please

paulmillar:
retest this please

PR #4007, opened 2018-06-06
doors: support advertising multiple addresses in LoginBroker

paulmillar:
Motivation:

By default, Login Broker advertises all available interfaces.

However, if additional interfaces should be published (e.g., DNS
aliases) then a door is limited to a single hostname.

Modification:

Add support for a door advertising multiple hostnames or IP addresses.
This change is designed to be backwards compatible with the existing
file format.

Add documentation describing the format of a property that is currently
poorly described.

Result:

dCache doors can advertise multiple interfaces, including DNS aliases.

Target: master
Request: 4.1
Request: 4.0
Request: 3.2
Ticket: http://rt.dcache.org/Ticket/Display.html?id=9410
Require-notes: yes
Require-book: yes
Patch: https://rb.dcache.org/r/10962
Acked-by: Tigran Mkrtchyan

Conflicts:
	modules/cells/src/main/java/dmg/cells/services/login/LoginBrokerPublisher.java


paulmillar:
retest this please

PR #4031, opened 2018-06-19
BUILDING.md - installing a new Maven

onnozweers:
Centos 7 has an ancient version of Maven, surprise surprise. Here's a hint how to install a newer version.


kofemann:
Thanks Onno!

PR #4052, opened 2018-06-22
frontend.authn.protocol by default https

onnozweers:
I think it's safer if the frontend interface is https by default, not http, especially now that it is going to have more admin capabilities.


paulmillar:
I agree

PR #4057, opened 2018-06-27
dcache-resilience: improve inaccessible file accounting

alrossi:
Motivation:

When multiple pools go offline it is possible that
all replicas for a given resilient file become unreadable.
If the file is not CUSTODIAL, and thus cannot be restored
from tape, the discovery of such a file during scanning
will generate an error in the 'history errors' listing,
in the resilience domain .resilience log, and will also
raise a general alarm concerning the pool.

There currently exists a command, 'inaccessible', which
generates a listing of the pnfsids on a given pool
which in the current state of dCacche have no readable
replicas.  However, this command takes a while to complete
(asynchronously), and the output is written to a file
which must be viewed by logging in.

Modification:

1.  Add 'referring pool' to the error output to enable
    grep'ing the resilience log for a given scanned pool
    (see below).
2.  Add options to the existing command to check status
    of the job, to list the contents of the file for
    that pool, and to clean up/delete the file.

_Also fixes a bug (uncaught NoSuchElementException)
in the validation of the pools._

Result:

Hopefully enhanced information and usability of this
functionality.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Acked-by: Tigran

### DIFF
--- a/modules/dcache-resilience/src/main/java/org/dcache/resilience/admin/ResilienceCommands.java
+++ b/modules/dcache-resilience/src/main/java/org/dcache/resilience/admin/ResilienceCommands.java
@@ -64,7 +64,9 @@ import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
 
+import java.io.BufferedReader;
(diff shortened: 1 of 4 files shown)


jstarek:
retest this please

PR #4074, opened 2018-07-04
webdav: pass on status message phrase to client

paulmillar:
Motivation:

When the WebDAV door is proxying a transfer and that transfer fails, that
failure could be due to some aspect of the client request (i.e., a 400
status code).

Currently the WebDAV door always returns a status "500 Internal Error" to
the client if a transfer fails.  This is true even if the pool returns a
400 status code (indicating the client did something wrong).  This is
both uninformative and fails to convey that the problem is with the
client's request.

Modification:

Update door to return a 400 error if a proxied transfer fails with the
pool returning a 400 error.  The status message is also relayed back to
the client.

Result:

Clients receive better feedback if a proxy transfer failed due to bad
client interaction.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/11039/
Acked-by: Tigran Mkrtchyan


paulmillar:
retest this please

paulmillar:
retest this please

Issue #3814, opened 2018-02-12
Replace CacheEntryFilter with java.util.function.Predicate


kofemann:
The CacheEntryFilter used by migration module to build various filters when selecting files to migrate. Since java 8 a  Predicate interface is available, which intended to be used in the similar context and can replace home-made analogue. Moreover, a collection of such filters can be replaces with chained predicates.

PR #4091, opened 2018-08-08
#3814 Replacement of CacheEntryFilter by java.util.function.Predicate

marcelmu:
(unfortunately no JUnit tests for Job class)


modules/dcache/src/main/java/org/dcache/pool/migration/NotFilter.java
```diff
@@ -1,23 +1,24 @@
...
 
+import java.util.function.Predicate;
 import org.dcache.pool.repository.CacheEntry;
 
 /**
  * Repository entry filter wrapping another filter. It accepts all
  * entries not accepted by the other filter.
  */
-public class NotFilter implements CacheEntryFilter
+public class NotFilter implements Predicate<CacheEntry>
```
kofemann: As you use standard Predicate, then there are no reasons to keep NotFilter(which is not used anyway)

kofemann:
ok to test

modules/dcache/src/main/java/org/dcache/pool/migration/NotFilter.java
marcelmu: I cleaned up the classes in the second commit of this pull request and removed NotStickyOwnerFilter.
I indeed overlooked NotFilter.

marcelmu: Fixed in 1fcf3a53b955616eb7c06c1e7a5b880880dd96f8

kofemann:
thanks!

Issue #4083, opened 2018-07-28
unclean shutdown of history if interrupted during start-up


paulmillar:
The following stack-trace was logged on prometheus

```
28 Jul 2018 08:01:26 (history) [] Uncaught exception in thread pool-5-thread-1
org.springframework.remoting.RemoteProxyFailureException: Failed to fetch pool monitor: null; nested exception is java.lang.InterruptedException
        at org.dcache.poolmanager.RemotePoolMonitor.getPoolMonitor(RemotePoolMonitor.java:210) ~[dcache-core-5.0.0-SNAPSHOT.jar:5.0.0-SNAPSHOT]
        at org.dcache.poolmanager.RemotePoolMonitor.getPoolSelectionUnit(RemotePoolMonitor.java:119) ~[dcache-core-5.0.0-SNAPSHOT.jar:5.0.0-SNAPSHOT]
        at org.dcache.services.history.pools.PoolTimeseriesServiceImpl.listPools(PoolTimeseriesServiceImpl.java:168) ~[dcache-history-5.0.0-SNAPSHOT.jar:5.0.0-SNAPSHOT]
        at org.dcache.util.collector.pools.PoolInfoCollector.collectData(PoolInfoCollector.java:101) ~[dcache-core-5.0.0-SNAPSHOT.jar:5.0.0-SNAPSHOT]
        at org.dcache.util.collector.pools.PoolInfoCollector.collectData(PoolInfoCollector.java:82) ~[dcache-core-5.0.0-SNAPSHOT.jar:5.0.0-SNAPSHOT]
        at org.dcache.services.collector.CellDataCollectingService.run(CellDataCollectingService.java:221) ~[dcache-core-5.0.0-SNAPSHOT.jar:5.0.0-SNAPSHOT]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [na:1.8.0_181]
        at java.util.concurrent.FutureTask.run(FutureTask.java:266) [na:1.8.0_181]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180) [na:1.8.0_181]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293) [na:1.8.0_181]
[...]

PR #4094, opened 2018-08-10
cells: add handling of RemoteProxyFailureException nested Interrupted…

alrossi:
…Exception to UncaughtException handler

Motivation:

The RemotePoolMonitor nests any InterruptedException thrown
in a spring RemoteProxyFailureException (RuntimeException).
This is done in order to preserve the original PoolMonitor
API, whose methods do not declare checked exceptions.

Not handling this special case can produce unnecessary
stack traces on shutdown (see Testing).

While the referencenced issue concerns the History
service, the problem is more widespread.  Of all
the instances where this could potentially occur,
only one handles it.

Modification:

Instead of wrapping each individual site with
a try catch which would result in a rethrow
if the nested exception is not an InterruptedException
(an anti-pattern), an if-clause is added
to the SystemCell uncaughtException method.

Result:

There should not be any stack traces resulting
from RemotePoolMonitor thread interruptions.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Bug: #4083
Acked-by: Tigran

### DIFF
--- a/modules/cells/pom.xml
+++ b/modules/cells/pom.xml
@@ -31,6 +31,10 @@
         <artifactId>curator-recipes</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-context</artifactId>
+    </dependency>
+    <dependency>
         <groupId>jline</groupId>
         <artifactId>jline</artifactId>
     </dependency>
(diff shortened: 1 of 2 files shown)


jstarek:
retest this please

Issue #4084, opened 2018-07-30
EOF failure on history service


alrossi:
see:

RT 9461] dCacheDomain crashed - queuedFlush.metadata java EOFException
RT 9462] Re: [Fes-tg] dCacheDomain crashed - queuedFlush.metadata java EOFException
RT 9463] Re: [Fes-tg] dCacheDomain crashed - queuedFlush.metadata java EOFException
RT 9464] Re: [Fes-tg] dCacheDomain crashed - queuedFlush.metadata java EOFException
RT 9465] Re: [Fes-tg] dCacheDomain crashed - queuedFlush.metadata java EOFException
RT 9466] Re: [Fes-tg] dCacheDomain crashed - queuedFlush.metadata java EOFException

http://rt.dcache.org/Ticket/Display.html?id=9466
http://rt.dcache.org/Ticket/Display.html?id=9465
http://rt.dcache.org/Ticket/Display.html?id=9464
http://rt.dcache.org/Ticket/Display.html?id=9463
http://rt.dcache.org/Ticket/Display.html?id=9462
http://rt.dcache.org/Ticket/Display.html?id=9461

alrossi:
Looks like their specific problem was that connectivity issue cause history service to fail on write, and leave one or more .json files incomplete.  Upon restart, the service fails because it tries to read back incomplete files (getting EOF).

alrossi:
Need to handle Gson/Json parse exceptions (=RuntimeExceptions, not IOExceptions) explicitly.

https://rb.dcache.org/r/11074/

PR #4097, opened 2018-08-13
dcache-history: handle Gson syntax errors explicitly

alrossi:
Motivation:

Since the history service is not transactional, network
hiccups can cause JSON file corruption.  If the service
contains corrupt or incomplete files in the
pool-history directory, it will fail to (re)start.

Modification:

Handle the JsonParseExceptions, which extend RuntimeException
(not IOException), explicitly.  When these occur,
also delete the file and map an empty placeholder info
object until the next refresh pass reconstructs the file.

Result:

Should no longer experience restart failures, such as the
one given in Testing.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Bug: #4084
Acked-by: Tigran


jstarek:
retest this please

PR #4101, opened 2018-08-13
dcache-xrootd: repair handling of delayed sync errors to client

alrossi:
Motivation:

Depending on timing, errors reported to the user client
through fireDelayedSync (third-party client) currently
may not appear.  This is particularly the case when
there is an error set on the tpc client before the first
sync call from the user client.  One example of this
is when there is no authentication module to handle
the gsi authentication request.

Modification:

Check first for tpc client errors being set on all incoming
user client requests, and report them immediately.

Result:

The correct error message will reach the user client.

Target: master
Request: 4.2
Acked-by: Tigran


jstarek:
retest this please

PR #4140, opened 2018-08-23
core: ensure pool/poolmanager communication receives errors

paulmillar:
Motivation:

Commit 0b94ca04 updated the doors to use PoolManagerStub, instead of a
CellStub, when sending messages to pool-manager and pool (when starting
a mover).

Unlike CellStub, the PoolManagerStub neglected to set the reply-required
flag.  This results in some error responses (such as running out of
capacity in a space reservation) being simply being discarded.  The
corresponding transfer then times out with a generic error, rather than
indicating that there was a problem.  Typical error message:

    Request to [>SpaceManager@local ... ] timed out.

Modification:

Update PoolManagerStub so it adds the reply-required flag when accepting
a messages.

Result:

Fix the problem where certain transfer failures, such as attempting to
use a space-reservation that has insufficient capacity, resulted in the
door eventually reporting a time-out problem to the client.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/11105
Acked-by: Albert Rossi


mksahakyan:
retest this please

PR #4176, opened 2018-09-03
ftp: add ability to log client-aborted transfers

paulmillar:
Motivation:

We are observing many complaints of aborted transfers.  Unfortunately,
dCache provides insufficient information to allow detailed post-mortem
analysis of why the transfer failed.

Modification:

Add a subclass of FTPException to identify when the client has aborted a
transfer.  This allows us to distinguish between transfers that fail for
some other reason (e.g., permission denied, insufficient capacity).

Add the 'ftp.enable.log-aborted-transfers' dCache configuration option
to control whether client-aborted transfers are logged.

Note that this patch disables this logging by default.  This is to
facilitate back-porting to supported branches; however, I am inclined to
enable this logging by default on master as such transfer cancellations
are almost always a cause for concern and, by logging this information,
a post-mortem may be able to deduce whether dCache (or the other party)
is to blame.

Support in AbstractFTPDoor to log the current status of the transfer,
triggered by the client aborting the transfer.  This includes collecting
some additional (potentially useful) information.

Some ancillary changes to support production of readable output.

Result:

dCache now has the ability to log the current status of a transfer at
the point the client decided to abort an FTP transfer.  This should
support a post mortem investigation on why a transfer was cancelled.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Requires-notes: yes
Requires-book: yes
Patch: https://rb.dcache.org/r/11124/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/common/src/main/java/org/dcache/util/LineIndentingPrintWriter.java
+++ b/modules/common/src/main/java/org/dcache/util/LineIndentingPrintWriter.java
@@ -1,0 +1,126 @@
+/* dCache - http://www.dcache.org/
+ *
+ * Copyright (C) 2018 Deutsches Elektronen-Synchrotron
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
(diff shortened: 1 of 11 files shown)


mksahakyan:
retest this please

PR #4194, opened 2018-09-13
dcache-dcap: add uid/gid to transfer info for plain/anonymous dcap

alrossi:
Motivation:

For plain dcap, the session subject received from login does not
contain uid or gid principals, but it is this subject which, via
IoDoorEntry, is conveyed to the transfer info.  These transfers
should nevertheless still be identifiable as to their initiator.

Modification:

If the subject used to construct IoDoorEntry does not contain
uid or gid principals, add them based on the uid and gid passed
during the "hello" command.

Result:

DCAP writes and reads no longer lack the uid/gid attributes.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Acked-by: Dmitry


paulmillar:
Please delay the back-port until the issue of a potential security vulnerability is resolved.

alrossi:
Will repost with new patch

PR #4201, opened 2018-09-20
webdav+transfermanagers: support TPC pull with targeted macaroons

paulmillar:
Motivation:

Fix support for WebDAV Third-Party-Copy (TPC) pull requests that are
authorised with a macaroon only valid for writing a specific file.

Modification:

Currently the transfermanager is responsible for creating the file for a
pull-request.

It is desirable that the WebDAV door checks whether the user can
transfer a file, as this allows the door to return an HTTP error rather
than returning 200 OK and an error as the final update in the streaming
response.  Therefore the WebDAV door checks if the user is allowed to
read/write (for push/pull transfers, resp.) the local file before
creating the request with transfermanager.

For non-macaroon authorised transfers, the granularity of file creation
is directories.  For macaroons, the granularity is per file.  Therefore
the existing infrastructure does not permit checking whether a user is
allowed to create an individual file.

To resolve this problem, responsibility for creating the file (for
WebDAV-triggered pull requests) is moved from the transfermanager to the
WebDAV door.  The transfermanager is still responsible for creating the
file for srmCopy pull requests.

Result:

It is possible to transfer a file using a targeted macaroon.

NOTE: both the webdav door and transfermanagers must be updated before
the fix is effective.

Target: master
Require-notes: yes
Require-book: no
Ticket: http://rt.dcache.org/Ticket/Display.html?id=9468
Patch: https://rb.dcache.org/r/11180/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/transferManager/TransferManagerMessage.java
+++ b/modules/dcache-vehicles/src/main/java/diskCacheV111/vehicles/transferManager/TransferManagerMessage.java
@@ -1,4 +1,6 @@
 package diskCacheV111.vehicles.transferManager;
+
+import diskCacheV111.util.PnfsId;
 import diskCacheV111.vehicles.Message;
 
 import org.dcache.auth.attributes.Restriction;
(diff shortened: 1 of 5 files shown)


paulmillar:
retest this please

PR #4207, opened 2018-09-20
dcache: add configuration for the Kafka producer timeout

paulmillar:
Motivation:

The kafka producer library has a property that controls how long it
spends when attempting to send an event.  If it takes too long, the
event is discarded and an error is logged.

Currently, the pool, and the xrootd, nfs and webdav doors hardcode this
timeout value to 0.  The dcap and ftp doors also have hard-code values,
albeit with a non-zero value.

For at least pools, this value is too small, as I have observed events
being discarded on an otherwise unloaded system, and unloaded kafka
instance.

Modification:

Introduce a configuration properties to control the Kafka timeout.  The
default is set to some (hopefully) reasonable default value.

Update all producers to use this configuration property.

Result:

The timeout used by dCache when attempting to send a Kafka event is now
adjustable via the configuration properties dcache.kafka.maximum-block
and dcache.kafka.maximum-block.unit.

The default timeout for pools, and the xrootd, nfs and webdav doors is
now non-zero.  This should fix the problem of kafka events being lost
under normal operational conditions.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Require-notes: yes
Require-book: no
Patch: https://rb.dcache.org/r/11177/
Acked-by: Marina Sahakyan


paulmillar:
retest this please

PR #4209, opened 2018-09-20
webdav: use TLS credential directly for gridsite

paulmillar:
Motivation:

An HTTP third-party copy works with the pool making an HTTP GET or PUT
request to some remote server.  This request may require dCache to
provide some form of credential with which it authorises the transfer.

One way to authorise the transfer is by presenting an X.509 credential.
To do this, the client delegates an X.509 credential using the GridSite
protocol and connects to the WebDAV door using an X.509 credential with
the same DN.  The client may specify the "Credential" header value
"gridsite" to ensure that the delegated credential is used.

The WebDAV door selects the correct delegated credential by matching the
delegated credential's DN with the DN of the X.509 credential the client
presented during the TLS handshake.  The WebDAV door obtains the latter
DN by extracting the GlobusPrincipal from the logged in user.

The problem is that it is not guaranteed that the logged in user will
have a GlobusPrincipal, even if the user presented an X.509 credential
during the TLS handshake.

For non-macaroon requests, it requires gPlazma be configured so that the
x509 plugin is always run.  Although this is a common deployment, it
isn't guaranteed.  For macaroon requests, the logged in user never
contains a GlobusPrincipal.

Modification:

Update the gridsite delegated credential selection to take the DN
directly from the TLS handshake credential (if any).

This patch also modifies the default behaviour of webdav TPC if no
"Credential" HTTP header is supplied.  With this patch, if the client
presents an X.509 credential during the TLS handshake then dCache will
(by default) attempt to use a delegated credential by default.

Result:

dCache can now transfer data with a remote site, authenticating with
that remote site using a delegated X.509 credential, but authenticating
locally with a macaroon.

Target: master
Ticket: http://rt.dcache.org/Ticket/Display.html?id=9492
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2

### DIFF
--- a/modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/CopyFilter.java
+++ b/modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/CopyFilter.java
@@ -1,6 +1,6 @@
 /* dCache - http://www.dcache.org/
  *
- * Copyright (C) 2014 Deutsches Elektronen-Synchrotron
+ * Copyright (C) 2014 - 2018 Deutsches Elektronen-Synchrotron
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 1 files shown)


paulmillar:
retest this please

paulmillar:
retest this please

Issue #4258, opened 2018-10-10
xrdcp fails to extract archive zip file


vingar:
Motivation
----------

xrdcp fails to extract archive zip file against dcache:

```
2018-10-10 11:29:39,010 DEBUG   Executing: xrdcp -vf root://grid-dc.rzg.mpg.de:1094//pnfs/rzg.mpg.de/data/atlas/dq2/atlasdatadisk/rucio/data16_13TeV/db/3f/DRAW_RPVLL.14552406._000091.zip.1?xrdcl.unzip=DRAW_RPVLL.11106701._002823.pool.root.1 -z DRAW_RPVLL.11106701._002823.pool.root.1 file:///afs/cern.ch/user/m/mlassnig/data16_13TeV/DRAW_RPVLL.11106701._002823.pool.root.1.part
2018-10-10 11:29:39,303 DEBUG   xrdcp status: 54
2018-10-10 11:29:39,303 DEBUG   xrdcp stdout: 
[0B/0B][100%][==================================================][0B/s]  
Run: [ERROR] Server responded with an error: [3015] Not a file
```

paulmillar:
Rod also reported that the problem was observed with LRZ running dCache v4.1.16; that this is likely NOT a regression.

Following Rod's description, I was able to reproduce the problem with prometheus, using the following commands:

```
paul@celebrimbor:~$ zip -j0 test.zip /bin/bash
  adding: bash (stored 0%)
paul@celebrimbor:~$ unzip -v test.zip 
Archive:  test.zip
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
 1099016  Stored  1099016   0% 2017-05-15 21:45 ddbc6e90  bash
--------          -------  ---                            -------
 1099016          1099016   0%                            1 file
paul@celebrimbor:~$ globus-url-copy file://`pwd`/test.zip gsiftp://prometheus.desy.de/Users/paul/test.zip
paul@celebrimbor:~$ xrdcp --zip bin/bash.zip root://prometheus.desy.de:1094/Users/paul/test.zip /tmp/
[0B/0B][100%][==================================================][0B/s]  
Run: [ERROR] Server responded with an error: [3015] Not a file
```

bbockelm:
Is it possible to set the debug output to level 3 to see precisely where `xrdcp` is choking?

paulmillar:
[...]

PR #4264, opened 2018-10-11
xrootd: add support for kXR_stat on open files

paulmillar:
Motivation:

Occationally, the xrootd client requires `kXR_stat` on an open file
handle to succeed.  It is not documented in the protocol specificiation
whether this is a reasonable expectation; but, given this is easy to
implement, it makes sense to support it.

Modification:

Update the pool to return information about the file if the `kXR_stat`
request targets an open file handle.  Requests that target a path
continue to be redirected back to the door.

Result:

The xrootd client is happer; in particular, the `--zip` option now
works.

NOTE this patch depends on xrootd4j >= 3.3.2 or >= 3.2.3.

Target: master
Request: 4.2
Requires-notes: yes
Requires-book: no
Closes: #4258
Patch: https://rb.dcache.org/r/11239/
Acked-by: Tigran Mkrtchyan
Acked-by: Albert Rossi


paulmillar:
retest this please

PR #4273, opened 2018-10-15
pool: update HTTP TPC to support retrying GET and HEAD requests for DPM

paulmillar:
Motivation:

If DPM is calculating a checksum, then any RFC 3230 (i.e., with a
'Want-Digest' header) GET or HEAD request returns '202 Accepted' respond
status line and an HTML page as the response entity.  Since dCache
considers any 2xx response as success, the HTML page is accepted as the
file's contents, resulting in data corruption.

Modification:

Add a special case that treats a 202 status code ('Accepted') for GET or
HEAD requests as the server indicating the result is not yet ready, and
retry the request after a short delay.

Only accept a 200 status code ('OK') when processing a GET response: all
other status codes are treated as indicating an error.

Result:

dCache now supports a DPM-specific HTTP extension that indicates the
checksum calculation is not yet complete, avoiding potential data
corruption with third-party copy.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/11242/

### DIFF
--- a/modules/dcache-vehicles/src/main/java/diskCacheV111/util/ThirdPartyTransferFailedCacheException.java
+++ b/modules/dcache-vehicles/src/main/java/diskCacheV111/util/ThirdPartyTransferFailedCacheException.java
@@ -1,5 +1,7 @@
 package diskCacheV111.util;
 
+import org.dcache.util.Exceptions;
+
 /**
  * Some unspecified problem when transferring a file between dCache and some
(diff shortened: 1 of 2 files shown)


paulmillar:
retest this please

paulmillar:
retest this please

PR #4281, opened 2018-10-15
dcache-frontend: add path filter to transfers

alrossi:
Motivation:

We would now like to provide backend filtering and sorting
capabilities on this attribute.

Modification:

Add the filter to the transfers API and implement it
analogously with the other string filters.  Also
add 'path' to the sort fields.

Result:

Filter box and multisort on file paths now possible.
Note: this patch also removes the catch block for NoSuchElementException
from the resource GET method.

Target: master
Request: 4.2
Acked-by: Dmitry


jstarek:
retest this please

jstarek:
If this needs admin-facing documentation, let‘s do that right now

Issue #4242, opened 2018-10-09
dcache-frontend: 500 error when getting info for a cell 


vingar:
```
$ curl -k   -u user#admin:pwd  -X GET https://srm.ndgf.org:3880/api/v1/cells/topo
{"errors":[{"message":"Internal error: java.lang.ArrayIndexOutOfBoundsException: 1","status":"500"}]}
```
In the server logs:
```
09 Oct 2018 16:44:00 (Frontend-piggy) [] Uncaught exception in thread jetty-72892
java.lang.ArrayIndexOutOfBoundsException: 1
	at org.dcache.restful.services.cells.CellInfoServiceImpl.getCellData(CellInfoServiceImpl.java:142) ~[dcache-frontend-4.2.8.jar:4.2.8]
	at org.dcache.restful.resources.cells.CellInfoResources.getCellData(CellInfoResources.java:137) ~[dcache-frontend-4.2.8.jar:4.2.8]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_181]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_181]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_181]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[na:1.8.0_181]
	at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory.lambda$static$0(ResourceMethodInvocationHandlerFactory.java:76) ~[jersey-server-2.26.jar:na]
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher$1.run(AbstractJavaResourceMethodDispatcher.java:148) ~[jersey-server-2.26.jar:na]
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.invoke(AbstractJavaResourceMethodDispatcher.java:191) ~[jersey-server-2.26.jar:na]
	at org.glassfish.jersey.server.model.internal.JavaResourceMethodDispatcherProvider$TypeOutInvoker.doDispatch(JavaResourceMethodDispatcherProvider.java:243) ~[jersey-server-2.26.jar:na]
[...]

PR #4294, opened 2018-10-16
dcache-frontend: fix array out of bounds exception in cell info service

alrossi:
Motivation:

See GitHub #4242

NOTE:  technically, the full domain name is needed to get live information,
since the cell's key to the cached value is the fully-qualified domain name.

Modification:

Check for length of split array of input.

Result:

Do not generate exception when input is of type -X GET https://srm.ndgf.org:3880/api/v1/cells/topo (i.e., no domain specified).
The information returned on such a cell, however, will be incomplete.

Target: master
Request: 4.2
Request: 4.1
Bug: #4242
Acked-by: Dmitry
Acked-by: Paul


jstarek:
retest this please

PR #4305, opened 2018-10-18
macaroons: include macaroon id in error message

paulmillar:
Motivation:

The macaroon's identifier is included in the Subject, if the macaroon is
valid.

If the macaroon is rejected then the macaroon id is currently not
included in the error message. This makes it hard to trace which
macaroon has been rejected, when it was requested, with which caveats,
etc.

Modification:

Update the message describing why a macaroon was rejected so it also
includes the macaroon's ID.

Result:

It is now possible to identify which macaroon is rejected, and so
investigate further why this is happening.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/11254/
Acked-by: Tigran Mkrtchyan


jstarek:
retest this please

paulmillar:
retest this please

PR #4320, opened 2018-10-23
dcache-frontend: add path filter to transfers

alrossi:
Motivation:

We would now like to provide backend filtering and sorting
capabilities on this attribute.

Modification:

Add the filter to the transfers API and implement it
analogously with the other string filters.  Also
add 'path' to the sort fields.

Result:

Filter box and multisort on file paths now possible.
Note: this patch also removes the catch block for NoSuchElementException
from the resource GET method.

Target: master
Request: 4.2
Acked-by: Dmitry


alrossi:
I accidentally deleted the original one.

alrossi:
There is no need to document this in the book, as the RESTful API is not documented resource by resource there.  One consults SWAGGER for such information.  This is auto generated.

PR #4344, opened 2018-10-24
core: revert custom bouncycastle version

paulmillar:
Motivation:

The custom bouncycastle JCE provider was found to fail on one of our
production instances.  The result was an inability to log into the admin
interface.

Modification:

Revert back to the standard bouncycastle version.

Result:

Admins should be able to use the admin interface again.

Request: 4.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/11282/
Acked-by: Tigran Mkrtchyan


jstarek:
Thanks for the PR, but I already reverted directly on the affected branches.

PR #4356, opened 2018-10-26
webdav: improve logging of TPC requests

paulmillar:
Motivation:

When investigating problems with HTTP TPC transfers, various useful
pieces of information are missing:

    what actually went wrong (as reported back to the client),

    which credential source was used,

    whether checksum verificiation was required,

    the source or destination URL.

Additionally, both TPC successful and failured operations are currently
logged at INFO level.  This is in contrast with other access log
entries, where failing operations are logged at either WARN or ERROR
level.

Modification:

Record information about the COPY request in various attributes.

Update the access log file so these attributes are recorded, if present.

Update the log level so TPC failures are logged at WARN level.

Result:

The admin has more logged information in the webdav access log file with
which to diagnose third-party-copy problems.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/11266/
Acked-by: Albert Rossi


jstarek:
retest this please

PR #4368, opened 2018-10-26
webdav: fail TPC request early on unknown hostname

paulmillar:
Motivation:

The client can write any hostname as the third-party-copy source or
destination URL.  If the host is unknown then the request is accepted
and poolmanager will (eventually) fail the request.

It is more efficient and provides better feedback if such bad requests
are failed as quickly as possible.

Modification:

Check within WebDAV door if the hostname was resolved.  If not, fail the
request.

Result:

Fail fast on HTTP third-party-transfers where the hostname is unknown.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/11289/
Acked-by: Tigran Mkrtchyan


jstarek:
retest this please

paulmillar:
retest this please

PR #4384, opened 2018-10-29
pool: increase TPC socket timeout for GET requests

paulmillar:
Motivation:

DPM will block HTTP GET requests while calculation checksums.
Therefore the default one minute timeout may be too short.

Modification:

Add a GET-specific timeout that allows for longer delays.

Result:

HTTP TPC Transfers involving DPM become more reliable.

Target: master
Requires-notes: yes
Requires-book: no
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Patch: https://rb.dcache.org/r/11301/
Acked-by: Tigran Mkrtchyan


jstarek:
retest this please

PR #4420, opened 2018-11-06
srm: gridsite fix querying validity of delegated credential

paulmillar:
Motivation:

The gridsite endpoint allows clients to delegate a credential.  It also
allows clients to manage delegated credentials.

There is a bug where a delegated credential with a primary FQAN (from a
VOMS server that dCache trusts) is not found when the client queries its
expiry time, but it is found when the client subsequently attempts to
delegate a credential.

This inconsitent behaviour prevents some gridsite clients, such as
davix, from delegating.

Modification:

Update RequestCredentialStore interface to support querying for a
credential with given name, ignoring any FQAN.  The credential with the
longest remaining validity is returned, irrespective of whether or not
that credential has a primary FQAN or the value of that primary FQAN.

Result:

Clients that use the gridsite protocol, such as davix, can now delegate
their credential.

Target: master
Require-notes: yes
Require-book: no
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Patch: https://rb.dcache.org/r/11327/
Acked-by: Dmitry Litvintsev

### DIFF
--- a/modules/srm-server/src/main/java/org/dcache/gridsite/SrmCredentialStore.java
+++ b/modules/srm-server/src/main/java/org/dcache/gridsite/SrmCredentialStore.java
@@ -56,8 +56,7 @@ public class SrmCredentialStore implements CredentialStore
     @Override
     public X509Credential get(DelegationIdentity id) throws DelegationException
     {
-        RequestCredential credential =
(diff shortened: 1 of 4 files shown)


paulmillar:
retest this please

Issue #4461, opened 2018-11-26
alarms: NPE when sending an alarm


ahaupt:
dCache version 4.2.18

When sending an alarm via the alarm service:

```
[trajan-vm2] (alarms@alarmDomain) admin > send "Test alarm"
sending alarm to trajan-vm2.zeuthen.desy.de:9867
```
... a backtrace in alarms log file is produced:

```
26 Nov 2018 13:43:49 (alarms) [] Uncaught exception in thread pool-5-thread-1
java.lang.NullPointerException: null
	at org.dcache.alarms.logback.LogEntryHandler$LogEntryTask.run(LogEntryHandler.java:148) ~[dcache-core-4.2.18.jar:4.2.18]
	at org.dcache.util.BoundedExecutor$Worker.run(BoundedExecutor.java:251) ~[dcache-common-4.2.18.jar:4.2.18]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [na:1.8.0_191]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [na:1.8.0_191]
	at java.lang.Thread.run(Thread.java:748) [na:1.8.0_191]
```

... and no mail is sent although mail sending is enabled here. This bug looks similar to #2578

alrossi:
Andreas,

I think your problem is independent of the MDC issue. It has to do with the received field being null (the JAVA autoboxing code will throw an NPE when you do something like this:

Integer i = null;

if (i == 0) {
...
}

and that is what seems to be the case here. However, I am not sure how this is actually occurring.

What database are you using ... RDBMS or XML?

Thanks, Al



________________________________________________
Albert L. Rossi
Application Developer & Systems Analyst III
Scientific Computing Division, Data Movement Development
FCC 229A
Mail Station 369 (FCC 2W)
Fermi National Accelerator Laboratory
Batavia, IL 60510
(630) 840-3023


________________________________

ahaupt:
Hi Al,

it's the default, so XML. And it gets filled with alarms, so this works.

```
[trajan-vm2] /root # ll /var/lib/dcache/alarms/store.xml 
-rw-r--r--. 1 dcache dcache 970 Nov 26 16:20 /var/lib/dcache/alarms/store.xml
```

Cheers,
Andreas

alrossi:
Hi Andreas,


yes, indeed I can reproduce the issue using XML.


[...]

PR #4473, opened 2018-11-30
alarms: fix persistence.xml configuration

alrossi:
Motivation:

Evidently the default value of the retainValues
property changed somewhere between original implementation of
alarms and the latest update to version 4.0+.

This value needs to be true so that we can reference
the values of the object after the transaction.

Currently, we are not setting the property for the
XML plugin.

Modification:

Explicitly set it to true as we do for RDBMS.

Result:

NPE no longer appears, and the referenced object
is guaranteed to have its original field values
(created by conversion from the logging event).

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Bug: #4461
Acked-by: Paul


jstarek:
retest this please

PR #4480, opened 2018-12-05
ftp/webdav: fix bypass of restrictions

paulmillar:
Motivation:

Restrictions are the mechanism for enforcing macaroon (and other) path
related limitations. This requires doors to send the path of a file.
Not all doors do this, resulting in the following warning:

    (PnfsManager) [door:... GFTP-... PnfsSetFileAttributes 00...F60] Restriction check by-passed due to missing path; please report this to <support@dCache.org>

The problem is present if the client specifies a checksum value with
either an FTP or WebDAV upload.

Modification:

Update WebDAV and FTP doors to always use the file's path.

Result:

No more warnings about restrictions being by-passed if the client
specifies a checksum value when uploading via FTP or WebDAV.

Target: master
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Requires-notes: yes
Requires-book: no
Ticket: https://rt.dcache.org/Ticket/Display.html?id=9565
Patch: https://rb.dcache.org/r/11416/
Acked-by: Albert Rossi


jstarek:
retest this please

PR #4511, opened 2019-01-14
DO NOT MERGE Revert "libs: upgrade to bouncycastle 1.54,  CANL to 2.5.0 and voms-a…

paulmillar:
DO NOT MERGE

…pi-java to 3.3.0"

This reverts commit eafbd5c1b11d3525787c80136d8dc81b52f44427.

It looks like this commit causes a problem for some VOMS servers:

14 Jan 2019 16:40:13 (gPlazma) [sfY:20:srm2:ls SRM-dcache-dot3 Login AUTH voms] Validation failure 1IY4 for DN "CN=Samuel Louis Bein,CN=739893,CN=sbein,OU=Users,OU=Organic Units,DC=cern,DC=ch": [[canlError]:CAnL certificate validation error: Signature of a CRL corresponding to this certificates CA is invalid, [invalidAcCert]:LSC validation failed: AA certificate chain embedded in the VOMS AC failed certificate validation!, [aaCertNotFound]:AC signature verification failure: no valid VOMS server credential found.]

### DIFF
--- a/modules/gplazma2-grid/src/main/java/org/dcache/gplazma/plugins/X509Plugin.java
+++ b/modules/gplazma2-grid/src/main/java/org/dcache/gplazma/plugins/X509Plugin.java
@@ -7,6 +7,7 @@ import eu.emi.security.authn.x509.proxy.ProxyUtils;
 import org.bouncycastle.asn1.ASN1Encodable;
 import org.bouncycastle.asn1.ASN1ObjectIdentifier;
 import org.bouncycastle.asn1.ASN1Sequence;
+import org.bouncycastle.asn1.DERObjectIdentifier;
 import org.bouncycastle.asn1.DERSequence;
 import org.bouncycastle.asn1.x509.GeneralName;
 import org.bouncycastle.asn1.x509.PolicyInformation;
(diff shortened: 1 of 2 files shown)


DmitryLitvintsev:
Paul, 

I do not think this is a good idea to revert this. How about we understand how to fix the issue? 

See, w/o this patch xrootd basically is not functional. 

ATM, Ryan Taylor is testing voms plugin patched dCache RPM. I would like to 
see the results of that

paulmillar:
We now have this problem here at DESY.  We also have a user who has experience the problem and is willing to test changes.  We also have a test instance with which we've been able to reproduce the problem.

This patch, which (as the title says) is NOT TO BE MERGED is merely to provide an RPM for testing, and to apply the usual testing.

The primary aim is to confirm that the bouncy-castle + CaNL patch is causing this problem and, by reverting that patch, we have one way of fixing it.

I understood that the xrootd problem was affecting ~1% of connections, randomly (not always the same user).

Once we know this, we can either test a potential fix, or make a new dCache release that reverts the BC+CaNL patch.

Given that this is preventing certain users for using dCache *right now* 100% of the time, my preference would be to revert the BC+CaNL patch.

I haven't seen this patch for VOMS you mentioned.  Is it on RB?

DmitryLitvintsev:
1) The patch is in RPM built  for Ryan. Not on RB. Since I cannot test it I rely on Ryan to test it for me. 
     I cannot reproduce the problem (needless to say we do not have any issues at Fermilab). 
    
2) 1% xrootd failure *per file*. Normally users run w./ 100 files per input, so the failure rate 
     is about 60% for our users here. 

Since almost 100 % of transfers here at Fermilab are xrootd - we were not really functional and users were *very* unhappy (to day the least). 

I would rather by-pass VOMS validation if this is where it fails in VOMS plugin rather than  revert the library update.

paulmillar:
I'm closing this pull-request, as it is no longer needed.

Please post any proposed patch on RB as quickly as possible, to facilitate a rapid review process.

PR #4525, opened 2019-01-22
Update cookbook-transport-security.md

onnozweers:
Mention webdav.redirect.allow-https, and dCache 5 & Java 11.

### DIFF
--- a/docs/TheBook/src/main/markdown/cookbook-transport-security.md
+++ b/docs/TheBook/src/main/markdown/cookbook-transport-security.md
@@ -8,7 +8,9 @@ To make sure data is safely encrypted during transfer, one should:
 1. Use the WebDAV protocol
 2. Configure dCache WebDAV doors not to redirect transfers (because currently it redirects to HTTP and not HTTPS)
 3. Have a well configured certificate or chain in each WebDAV door
-4. Have a well configured Java 8 or better
+4. Use Java 8 or better, with a well configured java.security
+
+Since 4.2, there is a new option `webdav.redirect.allow-https` which is `false` by default. If set to true, WebDAV doors should redirect clients to HTTPS movers on the pools. Your pools must have host certificates and have `pool.enable.encrypted-transfers = true` for this to work. This new feature should be considered experimental; test it before using it.
 
 Please note, that IT security is an ever changing field. Make sure you follow the latest developments. If security in dCache is a priority for you, be sure to follow the dCache user mailinglist.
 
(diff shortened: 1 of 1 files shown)


onnozweers:
Oh, and I couldn't find the chapter in the table of contents... Did I overlook it, or is the table perhaps outdated?

kofemann:
Thanks, Onno! Looks like we have lost the reference....

docs/TheBook/src/main/markdown/cookbook-transport-security.md
```diff
@@ -8,7 +8,9 @@ To make sure data is safely encrypted during transfer, one should:
 1. Use the WebDAV protocol
 2. Configure dCache WebDAV doors not to redirect transfers (because currently it redirects to HTTP and not HTTPS)
 3. Have a well configured certificate or chain in each WebDAV door
-4. Have a well configured Java 8 or better
+4. Use Java 8 or better, with a well configured java.security
+
+Since 4.2, there is a new option `webdav.redirect.allow-https` which is `false` by default. If set to true, WebDAV doors should redirect clients to HTTPS movers on the pools. Your pools must have host certificates and have `pool.enable.encrypted-transfers = true` for this to work. This new feature should be considered experimental; test it before using it.
```
paulmillar: As a writing style, we try to avoid phrases like "since 4.2" and "new option".

The reason is that each version of the book targets and describes a specific version of dCache.  For example, `master` branch targets (currently) dCache v5.1.  So it shouldn't mention dCache v5.0, v4.2, ...

Descriptions of how dCache has changed between  specific versions is (of course) very useful, especially for admins who are upgrading.   However, such version-comparing information belongs in the release notes, rather than in The Book.

As a suggestion, the following provides (roughly) the same information, but doesn't mention `4.2` or that the property is new:

    There is the `webdav.redirect.allow-https` option, which is `false` by default.  If set to true, ...

although you might like to rephrase the paragraph slightly, as the initial sentence is a little awkward in this version.

docs/TheBook/src/main/markdown/cookbook-transport-security.md
```diff
@@ -45,7 +47,7 @@ Configuring Java
 
 dCache relies on Java for its transport security. Hardening Java is necessary to get the maximum security. If not configured properly, Java may support ciphers and protocols that are considered weak.
 
-To begin with, make sure you run Java 8 or newer. It's impossible to get an A rating in the Qualys SSL test without Java 8 or newer. Java 11 would be best because it adds support for TLSv1.3, but dCache has not been tested with Java 11 (as of October 2018).
+To begin with, make sure you run Java 8 or newer. It's impossible to get an A rating in the Qualys SSL test without Java 8 or newer. If you run dCache 5, you can use Java 11 which adds support for TLSv1.3.
```
paulmillar: As above, the information is certainly very useful, but shouldn't reference a dCache version.

This patch is targeting `master` (==> dCache v5.1), so you can say simply (rephrasing, slightly)

    Using Java 11 will add support for TLSv1.3

(or similar)

onnozweers:
Hi Paul, thanks for the feedback; I hadn't realized that. I'll do some more commits to correct things. I'll separate the commits, so that they can be applied to the correct branches.

PR #4528, opened 2019-01-22
backport of series of patched to improve CEPH based pools

kofemann:


vingar:
would it be possible to upgrade also the rados4j dependency ? or in another pull request ?

PR #4623, opened 2019-02-11
webdav/frontend: add switch to reject macaroons sent unencrypted

paulmillar:
Motivation:

Bearer tokens are susceptible to being stolen.  To prevent this, it is
normally required that such tokens are only ever sent over an encrypted
channel.

dCache cannot enforce this behaviour, as it is the clients decision what
it sends over the network.  However, we can deliberately "break" such
clients, to act as a strong indicator that something is wrong.

Modification:

Update AuthenticationHandler to reject any macaroon sent over an
unencrypted channel.

Note: the default is set to not enforce this behaviour, as we want to
back-port this patch.  There should be a follow-up patch (targeting just
'master') where the default value is updated.

Result:

Sites may configure dCache to reject any macaroons send over an
unencrypted channel.  This is following security recommendations.  The
default behaviour is to continue accepting macaroons sent over an
unencrypted channel to avoid breaking existing deployment.

Target: master
Requires-notes: yes
Requires-book: no
Request: 5.0
Request: 4.2
Patch: https://rb.dcache.org/r/11522/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache-frontend/src/main/resources/org/dcache/frontend/frontend.xml
+++ b/modules/dcache-frontend/src/main/resources/org/dcache/frontend/frontend.xml
@@ -405,6 +405,8 @@
                     <property name="readOnly" value="${frontend.authz.readonly}"/>
                     <property name="enableBasicAuthentication" value="${frontend.authn.basic}"/>
                     <property name="realm" value="${frontend.authn.realm}"/>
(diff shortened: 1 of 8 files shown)


mksahakyan:
retest this please

PR #4632, opened 2019-02-12
module: README.md

einhorntee:
Motivation: I want my dCache to be better than the original

Modification: Change title in README.md

Result: title is now better than dcache

Target: master


einhorntee:
Was only a trial and should probably not be merged...

PR #4637, opened 2019-02-12
ACL: Neue Textdatei erstellt

gjyligashi83:
Text erstellt

### DIFF
--- a/modules/cells/src/main/java/dmg/cells/network/LocationMgrTunnel.java
+++ b/modules/cells/src/main/java/dmg/cells/network/LocationMgrTunnel.java
@@ -17,7 +17,7 @@
  */
 package dmg.cells.network;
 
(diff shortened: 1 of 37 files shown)


modules/cells/src/main/java/dmg/cells/network/LocationMgrTunnel.java
```diff
@@ -17,7 +17,7 @@
  */
 package dmg.cells.network;
 
-import org.slf4j.Logger;
+import org.slf4j.LOGGER;
```
kofemann: I believe this change  (and all following import changes) is incorrect

kofemann:
ok to test

jstarek:
Closing this pull request as it seems to have been more of a test.

Issue #4670, opened 2019-02-13
Update Pgpass class to use file system  API to check file permissions 


kofemann:
Pgpass file uses `stat` binary check file permission. This is not required as java7 have introdiced new API that allows to discover posix file attributes.

Additionally, URI class can be used to identify is string a url or not.

tomdittrich:
I will give it a try.

tomdittrich:
@kofemann 
> Additionally, URI class can be used to identify is string a url or not.

Do you mean something like this? https://stackoverflow.com/a/5719282/8041403
Or is there a extra URI class within the project?

kofemann:
@tomdittrich I was mo thinking about  [URI#create](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#create-java.lang.String-), But feel free to have an alternative solution.

tomdittrich:
`URI.create` have little problems to parse `-jdbcUrl=jdbc:postgresql:database` correctly. So you can check for yourself if this kind of uri schema is neccesary.

I have some suggestions for future education issues:
- replace all the System.out.prints with a logger for better debugging
- a proper error handling with exceptions (maybe you have to touch the classes, which call the pgpass methos)
- check file permission for non-posix systems
- a refactoring of the whole class  (avoid class members, more JavaDoc ...)

paulmillar:
@tomdittrich `URI.create` should be parsing `jdbc:postgresql:database`, not `-jdbcUrl=jdbc:postgresql:database`.  The latter is an option, which are (usually) parsed using the `Args` class.

tomdittrich:
@paulmillar 
I know. My post is not correct but I used it in the right way within the code. For `URI.create` you have to eliminate the `jdbc:` part with `substring(5)`

I talked about it with @kofemann 
He said it's okay if we leave it how it is for now.

tomdittrich:
@kofemann 
From my point of view, everything should be ready. There are no merge conflicts within the pull request too.

PR #4682, opened 2019-02-20
Pgpass file permission for POSIX (Issue #4670)

tomdittrich:
Issue #4670 
Please read the comments or commits for more information.
Thanks so far ...

### DIFF
--- a/modules/dcache/src/main/java/diskCacheV111/util/Pgpass.java
+++ b/modules/dcache/src/main/java/diskCacheV111/util/Pgpass.java
@@ -5,12 +5,16 @@
 package diskCacheV111.util;
 
 import java.io.BufferedReader;
-import java.io.BufferedWriter;
+import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.OutputStreamWriter;
-import java.io.PrintWriter;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.attribute.PosixFilePermission;
+import java.nio.file.attribute.PosixFilePermissions;
+import java.util.Set;
 
 /**
  *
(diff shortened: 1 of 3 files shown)


kofemann:
ok to test

kofemann:
@tomdittrich Shall I squash all commits into a single one?

tomdittrich:
If it doesn't violate the principle of "atomar", I don't see any problem. Thanks.

modules/dcache/src/main/java/diskCacheV111/util/Pgpass.java
```diff
@@ -75,51 +79,93 @@ private boolean parseUrl(String url) {
...
+     * @return permissions right? true
+     * @throws IOException
+     */
+    private boolean checkPgFilePermissions(String referencePermissionInput) throws IOException {
+
+        if(checkIfOsIsPosixCompliant()){
+            return checkPgFilePermissionsForPosix(referencePermissionInput);
+
+        } else {
+            System.out.println("Error reading permissions for '" + _pwdfile + "'. OS is not POSIX compliant");
```
paulmillar: Use slf4j to log errors, not `println`.

You should add a static field member, something like:

    private static final Logger LOGGER = LoggerFactory.getLogger("diskCacheV111.util.Pgpass");

modules/dcache/src/test/java/diskCacheV111/util/PgpassTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Some tests for the {@link Pgpass} class
+ *
+ * @author Tom Dittrich
+ * @version 0.1
+ */
+public class PgpassTest {
+
+    final String FILEPATH = "src/test/resources/pgpassTestFile.txt";
```
paulmillar: The `src` directory in a maven project is reserved exclusively for checked out code.  All dynamically created files (of any type) belong in the `target` directory.

Perhaps a better approach for unit-testing is to use an in-memory filesystem and configure it (in the test) exactly as required.  This avoids any accidental problems (wrong permissions, file already exists, ...)

For examples on how to do this with JimFS, see:
modules/gplazma2-grid/src/test/java/org/dcache/gplazma/util/IGTFInfoFileTest.java
modules/gplazma2-multimap/src/test/java/org/dcache/gplazma/plugins/GplazmaMultiMapFileTest.java

modules/dcache/src/main/java/diskCacheV111/util/Pgpass.java
```diff
@@ -75,51 +79,93 @@ private boolean parseUrl(String url) {
     }
 
     public String getPgpass(String hostname, String port, String database, String username) {
-        //
+        String result;
+
+        if (!checkIfFileExists()) {
+            System.out.println("File '" + _pwdfile + "' not exist");
```
paulmillar: Use slf4j logger, not println (see other comment)

modules/dcache/src/test/java/diskCacheV111/util/PgpassTest.java
tomdittrich: Sounds better, you are right.
Maybe it's a good idea for a future edu-issue? It can be combined with writing more tests or make the class better testable.

tomdittrich:
> There are calls to `println`. These must be converted to use slf4j.

I think I solved it.

modules/dcache/src/test/java/diskCacheV111/util/PgpassTest.java
kofemann: I think this should be fixed before code get merged. something  like `File#createTempFile` will do the job.

kofemann:
squashed and merged. Thanks!

PR #4720, opened 2019-03-13
webdav: fix resource name for door root

paulmillar:
Motivation:

Further reports of the following error:

    Your resource factory returned a resource with a different name to that requested!!!
    Requested: null returned: atlas - resource f[...]

An earlier patch (a42022b) attempted to address this problem but was
incomplete.

Modification:

Consider the user's and door's root when deciding the name of a
resource.

Result:

The error message "Your resource factory returned a resource with a
different name to that requested!!!" error message should no longer
appear.

Target: master
Requires-notes: yes
Requires-book: yes
Request: 5.0
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Ticket: https://rt.dcache.org/Ticket/Display.html?id=9650
Patch: https://rb.dcache.org/r/11595/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache-webdav/src/main/java/org/dcache/webdav/DcacheResource.java
+++ b/modules/dcache-webdav/src/main/java/org/dcache/webdav/DcacheResource.java
@@ -77,7 +77,7 @@ public class DcacheResource
     @Override
     public String getName()
     {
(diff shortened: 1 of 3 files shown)


mksahakyan:
retest this please

PR #4736, opened 2019-03-18
resilience: add ability to log resilience activity (incoming)

alrossi:
Motivation:

Providing an activity log, where resilience records its interactions
with other dCache components, may prove useful in understanding
behaviour.

Modification:

Log any cell message received by resilience.

Result:

It is now possible to record resilience activity (on the receiving
end), which may prove useful.

Target: master
Requires-notes: yes
Requires-book: no
Request: 5.0
Request: 4.2
Acked-by: Paul


mksahakyan:
retest this please

Issue #4738, opened 2019-03-20
TheBook: ban.conf example is broken


onnozweers:
Dear dCache developers,

In The Book, there is an example of a ban.conf file. Unfortunately, when I copied and pasted this example and modified our gplazma.conf to use it, EVERYONE was denied access because of a syntax error.

The example from https://github.com/dCache/dcache/blob/master/docs/TheBook/src/main/markdown/config-gplazma.md:

```
# Ban users by principal
alias dn=org.globus.gsi.jaas.GlobusPrincipal
alias kerberos=javax.security.auth.kerberos.KerberosPrincipal
alias fqan=org.dcache.auth.FQANPrincipal
alias name=org.dcache.auth.LoginNamePrincipal

ban name:ernie
ban kerberos:BERT@EXAMPLE.COM
ban com.example.SomePrincipal:Samson
```

I added a line like this:
```
ban dn:/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=....
```

From the log:

```
20 Mar 2019 10:08:33 (gPlazma) [admin ACCOUNT banfile] Bug in plugin: 
java.lang.IllegalArgumentException: No matching class found: org.globus.gsi.jaas.GlobusPrincipal
        at org.dcache.auth.Subjects.principalsFromArgs(Subjects.java:502) ~[dcache-common-5.1.0-SNAPSHOT.jar:5.1.0-SNAPSHOT]
        at org.dcache.gplazma.plugins.BanFilePlugin.principalsFromFile(BanFilePlugin.scala:82) ~[gplazma2-banfile-5.1.0-SNAPSHOT.jar:5.1.0-SNAPSHOT]
        at org.dcache.gplazma.plugins.BanFilePlugin.$anonfun$bannedPrincipals$1(BanFilePlugin.scala:89) ~[gplazma2-banfile-5.1.0-SNAPSHOT.jar:5.1.0-SNAPSHOT]
        at org.dcache.gplazma.plugins.FileCache.getOrFetch(FileCache.scala:13) ~[gplazma2-banfile-5.1.0-SNAPSHOT.jar:5.1.0-SNAPSHOT]
        at org.dcache.gplazma.plugins.FileCache.getOrFetch$(FileCache.scala:10) ~[gplazma2-banfile-5.1.0-SNAPSHOT.jar:5.1.0-SNAPSHOT]
[...]

PR #4739, opened 2019-03-20
Fixing example of ban.conf that breaks dCache.

onnozweers:
See also https://github.com/dCache/dcache/issues/4738.


paulmillar:
Thanks Onno!

PR #4804, opened 2019-04-13
pool: avoid throwing a RuntimeException for non-bugs

paulmillar:
Motivation:

The pool uses a lambda function to create a mover, wrapping any
CacheException thrown while creating the mover with a RuntimeException.

The code that handles RuntimeExceptions is faulty, as it checks whether
the root cause is a CacheException, rather than the immediate cause.
The CacheException may, itself, have a cause that isn't a
CacheException.

The result is that a CacheException is re-thrown as a RuntimeException,
which is then logged as a bug in dCache.

Modification:

Use a custom RuntimeException (UncheckedCacheException) to wrap the
CacheException within the lambda.  The catch block is now specific
enough to catch only the CacheException, leaving other RuntimeExceptions
to propagate.

Result:

Certain error cases, where a pool is unable to create a mover are no
longer logged as a bug in dCache.

Target: master
Request: 5.1
Request: 5.0
Request: 4.2
Request: 4.1
Request: 4.0
Request: 3.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/11688/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/pool/classic/MoverRequestScheduler.java
+++ b/modules/dcache/src/main/java/org/dcache/pool/classic/MoverRequestScheduler.java
@@ -1,8 +1,8 @@
 package org.dcache.pool.classic;
 
-import com.google.common.base.Throwables;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import java.io.InterruptedIOException;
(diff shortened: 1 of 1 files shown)


kofemann:
retest this please

Issue #4825, opened 2019-05-07
PinManager logging explodes


calestyo:
Hi.

I've just upgraded from 4.2.26 to .32 and when I started just the core domains the logging of the Pinmanager went crazy repeating:
```
May  7 01:44:31 lcg-lrz-dcache0 dcache@storagesystem0[14685]: 2019-05-07 01:44:31+02:00 (PinManager) [] Unable to clear sticky flag because pool lcg-lrz-dc08_0 is unavailable
May  7 01:44:31 lcg-lrz-dcache0 dcache@storagesystem0[14685]: 2019-05-07 01:44:31+02:00 (PinManager) [] Unable to clear sticky flag because pool lcg-lrz-dc29_6 is unavailable
May  7 01:44:31 lcg-lrz-dcache0 dcache@storagesystem0[14685]: 2019-05-07 01:44:31+02:00 (PinManager) [] Unable to clear sticky flag because pool lcg-lrz-dc28_5 is unavailable
May  7 01:44:31 lcg-lrz-dcache0 dcache@storagesystem0[14685]: 2019-05-07 01:44:31+02:00 (PinManager) [] Unable to clear sticky flag because pool lcg-lrz-dc02_4 is unavailable
May  7 01:44:31 lcg-lrz-dcache0 dcache@storagesystem0[14685]: 2019-05-07 01:44:31+02:00 (PinManager) [] Unable to clear sticky flag because pool lcg-lrz-dc24_1 is unavailable
May  7 01:44:31 lcg-lrz-dcache0 dcache@storagesystem0[14685]: 2019-05-07 01:44:31+02:00 (PinManager) [] Unable to clear sticky flag because pool lcg-lrz-dc30_1 is unavailable
May  7 01:44:31 lcg-lrz-dcache0 dcache@storagesystem0[14685]: 2019-05-07 01:44:31+02:00 (PinManager) [] Unable to clear sticky flag because pool lcg-lrz-dc27_6 is unavailable
May  7 01:44:31 lcg-lrz-dcache0 dcache@storagesystem0[14685]: 2019-05-07 01:44:31+02:00 (PinManager) [] Unable to clear sticky flag because pool lcg-lrz-dc31_1 is unavailable
May  7 01:44:31 lcg-lrz-dcache0 dcache@storagesystem0[14685]: 2019-05-07 01:44:31+02:00 (PinManager) [] Unable to clear sticky flag because pool lcg-lrz-dc30_2 is unavailable
[...]

PR #4898, opened 2019-05-28
pinmanager: add logging info

onnozweers:
In issue #4825 we've seen that it may be helpful to understand whether a pinmanager assumes the leader role. I hope with these two lines, it can be understood from the logs whether that happens. In case the messages are incorrect or inaccurate, please feel free to correct them.

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/pinmanager/PinManager.java
+++ b/modules/dcache/src/main/java/org/dcache/pinmanager/PinManager.java
@@ -181,6 +181,7 @@ public class PinManager
         @Override
         public void isLeader()
         {
+            _log.info("Assuming leader role. Scheduling Expiration and Unpin tasks.");
             expirationFuture = executor.scheduleWithFixedDelay(
                     new FireAndForgetTask(expirationTask),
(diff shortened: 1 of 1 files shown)


modules/dcache/src/main/java/org/dcache/pinmanager/PinManager.java
```diff
@@ -196,6 +197,7 @@ public void isLeader()
         @Override
         public void notLeader()
         {
+            _log.info("Not assuming leader role. Probably there is another PinManager that already is entitled to the leader role.");
```
paulmillar: The text here is a little misleading: the `notLeader` method is called when a pinmanager that was the leader is demoted so it is no longer the leader. Suggest something like:

    Dropping leader role. [...]

Also, it is usually (almost always) a bad idea to speculate in log messages: report what you know to be true; e.g.,

    Dropping leader role.  Cancelling expiration and unpin tasks.

onnozweers: Thanks for clarifying this Paul!

paulmillar:
ok to test

PR #4899, opened 2019-05-28
pinmanager unpinprocessor: add debugging info

onnozweers:
Here are some suggestions for debugging info to have a better idea what the pinmanager is doing and whether it is processing pins in "unpinning" state. This helped me to better understand issue #4825. Feel free to adjust messages and log levels (perhaps some things should be trace level instead of debug).

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/pinmanager/UnpinProcessor.java
+++ b/modules/dcache/src/main/java/org/dcache/pinmanager/UnpinProcessor.java
@@ -81,8 +81,10 @@ public class UnpinProcessor implements Runnable
     private void upin(Semaphore idle, Executor executor, Pin pin) throws InterruptedException
     {
         if (pin.getPool() == null) {
+            _logger.debug("No pool found for pin {}, pnfsid {}; no sticky flags to clear", pin.getPinId(), pin.getPnfsId());
             _dao.delete(pin);
         } else {
+            _logger.debug("Clearing sticky flag for pin {}, pnfsid {}", pin.getPinId(), pin.getPnfsId());
             clearStickyFlag(idle, pin, executor);
         }
     }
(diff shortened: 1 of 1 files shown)


paulmillar:
ok to test

paulmillar:
I like the idea of this patch, but I think it could be done "better".  For example, the pin deletion messages (I think) belong within the DAO and I would add some of the logged information in the context, rather than explicitly in the message.

onnozweers:
Hi Paul, I agree, but it's a bit difficult for me because the DAO does not have a logger yet, and it's not a class but an interface, and I think I've reached the edge of my very limited Java skills.

paulmillar:
Sure, that's not a problem -- I can add something there.

paulmillar:
Patch: https://rb.dcache.org/r/11770/

Turns out there's quite a few bugs in how logging is handled in PinManager, but I think I've got them all, now.

With this patch, you no longer need the debug statements: `_logger.debug("Deleting pin` ...

paulmillar:
@onnozweers, could you rebase your patch against the current tip of master and retest?

I think you'll find the last two debug lines (`_logger.debug("Deleting pin`...) are not needed any more.

jstarek:
ping @onnozweers  ;-)

onnozweers:
I have full confidence that logging now has improved significantly. Wouldn't it just be better if I withdrew this pull request?

onnozweers:
I've looked into rebasing but it looks complicated to me.
Can I close this pull request and submit a new one, based on the current master? That I can do.

paulmillar:
Hi Onno,

Of course!  feel free to close this pull request and create a fresh one.

onnozweers:
Continued in #4948.

PR #5023, opened 2019-07-24
gplazma2-argus: logger changed to correct standard

s0559289:
gplazma2-argus: logger changed to correct standard

Motivation: logger variable name is different in modules

Modification: name changed from _log to LOGGER

Result: logger variable name is now LOGGER

Target: master
Require-book: no
Require-notes: no


kofemann:
ok to test

PR #5028, opened 2019-07-24
Chimera logger unification

marcel3007:
chimera: changed name of Logger 

Motivation: unification of Logger

Modification: changed all names to LOGGER

Result: unificated logger name

Target: master
Require-book: no
Require-notes: no


kofemann:
ok to test

PR #5029, opened 2019-07-24
Logger unification

Arnika777:
gplazma2-voms: Logger name was changed

Motivation: every Logger shoud have the same name

Modification: Logger name was changed

Result:

Target: master
Require-book: no
Require-notes: no


kofemann:

PR #5030, opened 2019-07-24
Logger unification

nimaazha:


modules/acl/src/main/java/org/dcache/acl/mapper/AclUnixMapper.java
```diff
@@ -25,7 +25,7 @@
  */
 public class AclUnixMapper {
 
-    private static final Logger logger = LoggerFactory.getLogger("logger.org.dcache.authorization." + AclUnixMapper.class.getName());
+    private static final Logger LOGGER = LoggerFactory.getLogger("LOGGER.org.dcache.authorization." + AclUnixMapper.class.getName());
```
kofemann: I think 
```
LoggerFactory.getLogger("LOGGER.org.dcache.authorization." 
```

is unintended change and  should stay lowercase

kofemann:

PR #5033, opened 2019-07-24
Logger unified cells

nimaazha:
Module: Cells : Logger name was changed

Motivation: every Logger should have the same name

Modification: Logger name was changed

Result:

Target: master
Require-book: no
Require-notes: no

### DIFF
--- a/modules/acl/src/main/java/org/dcache/acl/mapper/AclMapper.java
+++ b/modules/acl/src/main/java/org/dcache/acl/mapper/AclMapper.java
@@ -25,17 +25,17 @@ import org.dcache.auth.Subjects;
  */
 public class AclMapper {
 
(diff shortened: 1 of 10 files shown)


modules/acl/src/main/java/org/dcache/acl/mapper/AclMapper.java
```diff
@@ -25,26 +25,26 @@
  */
 public class AclMapper {
 
-    private static final Logger logger = LoggerFactory.getLogger("logger.org.dcache.authorization." + AclMapper.class.getName());
+    private static final Logger LOGGER = LoggerFactory.getLogger("LOGGER.org.dcache.authorization." + AclMapper.class.getName());
```
kofemann: we want to change only variable name `logger` to `LOGGER`. The reference in configuration, which is used LoggerFactory.getLogger(...) shoudl stay is.

kofemann:

PR #5078, opened 2019-09-20
scripts: set hsmcp and infoprovidercms executable

jstarek:
Motivation:

Github ticket 3158 noted that the default HSM script is not set
as executable during the installation. Likewise, the
infoprovidercms.rb script also needs this set.

Other files in /usr/share/dcache/lib/ could also be set executable
by default. However, since most are templates that are meant to be
customized by admins, this patch only modifies these two files'
permissions.

Modification:

The RPM specfile and the DEB's postinst script were modified
so that the hsmcp.rb and infoprovidercms.rb scripts are installed
with 755 permissions.

Result:

Simple setups with tape connection work out-of-the-box.

Target: master
Require-notes: yes
Require-book: no
Request: 5.2
Request: 5.1
Request: 5.0
Request: 4.2

### DIFF
--- a/packages/fhs/src/main/deb/postinst
+++ b/packages/fhs/src/main/deb/postinst
@@ -42,6 +42,8 @@ Please fix this and reinstall this package." >&2
     chmod 700 /var/lib/dcache/statistics
 
     # allow somewhat more relaxed permissions elsewhere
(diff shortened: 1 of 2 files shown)


jstarek:
retest this please

PR #5112, opened 2019-10-17
How to enable scitokens in gplazma.conf

onnozweers:
Here's a hint for in the gplazma.properties file on how to enable scitokens by modifying the gplazma.conf.


paulmillar:
Thanks, Onno!

PR #5269, opened 2020-01-15
Fix tape-reserved size calculation

IATkachenko:
Calculation of used and reserved tape sizes looks like copy-paste and
condition in sum() for reserved was no changed.
Fixed now.


paulmillar:
OK to test

paulmillar:
Thanks Igor!

joanmarcriera:
Hi, 

May I ask what tape systems do you use? We have a T950 and a TS4500, and we are looking for a piece of software that can manage both, with an S3 entry point if possible, so we integrate with dCache or similar in the future. 
Can you advise/share what kind of tape management are you using?


Best regards
Marc

Issue #5359, opened 2020-03-16
Stacktrace in Log


mksahakyan:
16 Mar 2020 15:47:15 (Frontend-release-6-1) [] Uncaught exception in thread pool-55-thread-56
java.lang.NullPointerException: Cannot merge statistics, metadata object was null.
        at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:900)
        at org.dcache.util.histograms.HistogramMetadata.mergeStatistics(HistogramMetadata.java:179)
        at org.dcache.util.collector.pools.PoolInfoCollectorUtils.mergeLastAccess(PoolInfoCollectorUtils.java:333)
        at org.dcache.restful.util.pool.PoolHistoriesHandler.getAggregateFileLifetime(PoolHistoriesHandler.java:193)
        at org.dcache.restful.util.pool.PoolHistoriesHandler.update(PoolHistoriesHandler.java:265)
        at org.dcache.util.collector.pools.PoolInfoAggregator.lambda$aggregateDataForPoolGroups$0(PoolInfoAggregator.java:100)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
        at java.base/java.util.HashMap$ValueSpliterator.forEachRemaining(HashMap.java:1675)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
        at org.dcache.util.collector.pools.PoolInfoAggregator.aggregateDataForPoolGroups(PoolInfoAggregator.java:95)
        at org.dcache.restful.util.pool.PoolDataRequestProcessor.postProcess(PoolDataRequestProcessor.java:105)
        at org.dcache.util.collector.RequestFutureProcessor.waitUntilFinished(RequestFutureProcessor.java:233)
[...]

PR #5361, opened 2020-03-18
dcache-history,dcache-frontend: guard against unconfigured sweeper hi…

alrossi:
…stogram

Motivation:

GH Issue #5359.

The NPE stack trace arises because the pool
data could be sent with a default sweeper data
object.  This object currently contains
a default counting histogram object, but
that object in turn has internal fields,
in particular, the metadata field, which
have not been initialized, causing
a NPE on processing.

Modification:

In this patch we just guard against
processing an uninitialized counting
histogram.

A subsequent patch (for master only)
will modify how the Sweeper initializes
histograms.

Result:

Should not see the NPE reported below.

Target: master
Bug: dcache/issues/5359
Patch: https://rb.dcache.org/r/12264/
Request: 6.1
Request: 6.0
Request: 5.2
Request: 5.1
Request: 5.0
Closes: #5359
Acked-by: Olufemi


mksahakyan:
retest this please

PR #5372, opened 2020-03-19
dcache-frontend: make ErrorResponseProvider return the more specific …

alrossi:
…error message

Motivation:

Return the most specific error message on RESTful calls.

Modification:

Get the root cause error message and return it if available;
absent that, use the exception message; absent that, use
the default.

Include NotFoundExceptions in this procedure.

Result:

More meaningful error responses.

Target: master
Request: 6.1
Request: 6.0
Request: 5.2
Request: 5.1
Request: 5.0
Patch: https://rb.dcache.org/r/12271
Requires-book: no
Requires-notes: no
Acked-by: Lea


mksahakyan:
retest this please

PR #5392, opened 2020-04-20
dcache-xrootd: fix compatible level security for sigver

alrossi:
Motivation:

Fix a bug where setting the
security level to 'compatible'
was forcing a signed hash on
all open calls rather than
just opens for write.

Modification:

Upgrade to xrootd4j 3.5.6

Result:

Compatible level behaves correctly.

Target: 6.1, 6.0, 5.2, 5.1, 5.0
Requires-notes: yes
Requires-book: no

See xrootd4j patch: https://rb.dcache.org/r/12321
master@0406002401d73d39a9b43f48ee769637cabd76b6


mksahakyan:
retest this please

PR #5433, opened 2020-06-02
dcache-xrootd: cancel TPC transfer when client disconnects unexpected…

alrossi:
…ly from pool

Motivation:

See https://github.com/dCache/xrootd4j/issues/9

29 Mar 2019 11:38:58 (dmsdca19-1) [] Uncaught exception in thread xrootd-tpc-client-1
java.lang.IllegalStateException: ChecksumChannel must not be written to after getChecksums

While the cause of the original two-party example
is still not understood, I have not noticed it on our systems
since 2018.   The TPC error, however, occurs sporadically on
our WLCG-DOMA-TPC test stand.

Modfication:

The third-party version appears to be provoked by a premature
disconnect after timeout by the client connected to the pool.
The disconnect on the channel provokes the finalization
of the checksum, but a write from the TPC client is
still in flight and is attempted thereafter, since
the client still carries a handle to the mover channel.

The fix is simply to shutdown the TPC client first when
the pool netty channel goes inactive.

Result:

The ChecksumChannel error is no longer observed for TPC.

Target:  master
Request: 6.1
Request: 6.0
Request: 5.2
Request: 5.1
Request: 5.0
Patch: https://rb.dcache.org/r/12391
Closes: dCache/xrootd4j/issues/9
Acked-by: Dmitry
Acked-by: Lea

### DIFF
--- a/modules/dcache-xrootd/src/main/java/org/dcache/xrootd/pool/XrootdPoolRequestHandler.java
+++ b/modules/dcache-xrootd/src/main/java/org/dcache/xrootd/pool/XrootdPoolRequestHandler.java
@@ -181,6 +181,10 @@ public class XrootdPoolRequestHandler extends AbstractXrootdRequestHandler
         /* close leftover descriptors */
         for (FileDescriptor descriptor : _descriptors) {
             if (descriptor != null) {
+                if (descriptor instanceof TpcWriteDescriptor) {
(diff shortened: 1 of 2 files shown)


mksahakyan:
retest this please

alrossi:
Sorry, I did this for the other branches but must have missed this one.

Issue #3705, opened 2017-11-15
systemd: "dcache start/stop" confuses systemd


calestyo:
Hey.

Just noted, when doing `dcache stop` or `dcache start` then systemd get's a bit confused, namely while the former two are usually considered to "start/stop all dcache on the system" (which people will think corresponds to `dcache.service`), it in fact only starts/stop each single domain (and on systemd each single instance unit) - the "global" service doesn't get started or stopped.

Thus when I started dCache e.g. via `systemctl start dcache.service`, but stop it via `dcache stop` systemd will not notice that `dcache.service` is no longer active.

This is not a big deal for now, but if people would start to use e.g. dcache.service (and it's state in their own units... e.g. as a dependency for a systemd controlled pnfs mount)... or in querying whether dcache runs... it could become one.

TBH, I think workarounds like checking in `domainStart()` and `domainStop()` are just hacks and would sooner or later cause problems.
The main problem here is rather that we interfere again with what's not our business in a systemd world (starting/stopping services) (<--- and this shouldn't be read as a complaint... :-) .. it's just that the situation changes with systemd).

IMO, the proper solution is to deprecate/remove all `dcache stop`/`start` and similar things which are going to be obsolete sooner or later.
But as with other such changes (dcache.user, dcache.log.dir, etc.) that's some bigger changes which you guys should probably discuss first (or maybe we can have some telco about it - but don't assume that I'd consider myself the big systemd expert ;-) so I might be of little help)

Cheers,
Chris.

paulmillar:
We are working on improving the systemd support, which requires some restructuring of our we structure the dCache code.  In essence, we want to "pull out" the platform specific parts so that these may be maintained independently of the dCache code-base.  This change will allow us to have different support scripts: ones for systemd deployments, for sys-v deployments, and for tar-ball deployments.

calestyo:
One question is: will there be so many non-systemd deployments to justify them being supported?

I don't think it's easy to achieve supporting all three equally well, the problem with the config options like dcache.user is a good example for this. Logging is another.

sysvinit will simply die pretty soon in any of the major linux distros, so I personally wouldn't want to put too much effort into a dying system... and as I've said, without many ugly hacks and much inhomogeneity, I think it's barely doable to support more than one.

RHEL7 is already there and the typical production dCache site has IMO only little reason to say that they couldn't update to it (since dCache nodes are usually dedicated).

paulmillar:
[...]

Issue #3630, opened 2017-10-25
dcache@*.service cannot be enabled


calestyo:
Hey.

Seems no one really ever tested the new systemd units ;-)

When one tries to enable a dcache@*.service, to have it automatically startup on next boot, one gets e.g.
```
# systemctl enable dcache@admin.service 
Failed to enable unit: Unit /run/systemd/generator/dcache@admin.service is transient or generated.
```

This is because of the unit files being created completely in the generator (and not just symlinks, to one central `dcache@.service`, where the instance name is used with e.g. `%i`.

#3628 partially solves this, as one can enable the global `dcache.service` I introduce there, but it doesn't give the flexibility which is even advertised in the release notes :-P

I think the only way would be to have a `/lib/systemd/system/dcache@.service`, at first I was kinda surprised that you don't do this anyway, but I guess the reason is you want the per domain config options like `dcache.restart.delay` in the unit file... with values as at the time when systemd is re-execed.

Not sure what's the best approach... creating the whole `dcache@*.service` files in the generator is probably not the cleanest/intended way though.

And I don't think adding a "static" dummy `/lib/systemd/system/dcache@.service` in addition to the `dcache@*.service` would be a good idea.

Any thoughts?

calestyo:
I was asking around at IRC, and what they told me there is in principle that creating real foo@bar.service units (i.e. as real files) is not supposed to be done... and may stop working any time.

When one uses such "instance units" (i.e. with an @) there should be allegedly really just one file that is named foo@.service... and that uses then %i and friends to create instances thereof.

While it's easy to create such a dcache@.service (the generator would still need to be kept for creating symlinks for each existing domain).... I have no real idea right now, how to do the config value parsing...

CC: @gbehrmann

calestyo:
I think the way to go might be the following:

- split out most of the stuff between cat <<-EOF from the generator into a proper dcache@.service instance unit.
- the generator keeps to continue the symlinks (then pointing towards the instance unit)
- in addition, the generator will create an EnvironmentFile= *per* domain

Now that env file is obviously transient... and it should go into some /run/... location.

Am I right that dcache already creates a configuration cache of the config in /etc/ ... and uses that cache instead of the /etc/dcache/ stuff... so that changes to /etc/dcache/* only get picked up when a domain restarts?
[...]

PR #5463, opened 2020-06-29
systemd: convert dcache.service into dcache.target

kofemann:
Motivation:
the systemd service are expect to run processes. The dcache.service,
however, there to group per-domain units together. The original
reason to make it a service was an ability propagate reloads to the
set of dependent units. This makes no sense in case of dcache as (a)
domain processes don't depend on each other and (b) dcache does not
support dynamic configuration reload.

Modification:
change dcache.service into dcache.target which is more suitable for
dcache domain grouping.

Result:
more systemd friendly deployment.

NOTE: this change breaks existing (debian) installations, but impact is
small. Now admins have to run

systemctl start dcache.target

instead of

systemctl start dcache.service

Acked-by:
Target: master
Require-book: yes
Require-notes: yes


kofemann:
This pull request is here to let @calestyo to review and comment as original author of dcache.servcie and one of view systemd users.

calestyo:
Hey...

Well I think technically, using a target would work,... but from my understanding it's conceptually rather not meant for it in the systemd-world.

Targets are kinda vaguely defined as "used for grouping units and as well-known synchronization points during start-up".
I don't think the synchronisation point aspect applies here... this rather means things like networking.target or so.
And grouping... well it obviously would group the dcache domains, but I think this is rather meant for grouping of completely independent services (which the dCache domains are not... they share at least config and so on).

At least I wouldn't know of any clustered service which makes use of targets in these manner... all targets I see on my systems are rather abstract things like networking, multi-user and so on.

I, personally, would have found the construct of using one master dcache.service, which pulls in the dcache@*.service instances the best... it also matches the idea, that one can have multiple instances of the very same service,... like getty@tty1.service, getty@tty2.service.... and analogously for dCache e.g. multiple pools or ftp doors on one node.


There are however many other systemd related issues (just search for systemd... I've reported most of them, IIRC).

But at least some of these rather require some bigger decisions IMO.
I don't think it's a good idea trying to just somehow make the current way of doing things work with systemd - but rather to use the systemd way of doing things.
But this most likely means to more or less completely drop support for sysvinit... and also to make more user-visible changes, since options like dcache.user (and several others) wouldn't have any place then. Also dcache start/stop would need to go away... or dates in log files would be added by systemd (not dCache itself).

Specifically one would also need to decide when configuration changes (in /etc/dcache and other dcache config locations) actually make it do the binaries... this in conjunction with the generator for the dcache@*.service files.
Like when I add a new domain in the config...systemd would only see it when I do a systemctl reload... but other options would get picked up just by restarting the domains.
That difference is IMO a bit ambiguous.

Also, dCache's systemd generator itself works a bit non-standard and thus causes issues (see #3630).

And I do think to remember that dcache's own "process handling" (i.e. the startup of domains done by the dcache script) confuses systemd (in addition to the issues described in #3705)... so that systemd would never recognise when e.g. starting actually failed (e.g. because of config errors).
This also reminds me: the automatic restart of crashed domains... has basically no place in the systemd world, where systemd itself would handle this.


Cheers and hope that helps,
Chris

calestyo:
btw: "(a) domain processes don't depend on each other and" is of course right... but what I tried to explain above... the domains are still "connected", even if just conceptually by sharing e.g. the same config.

So from what I can see, there is e.g. no real difference to getty@*.service/getty.service ... or postgresql@*.service/postgresql.service ... their instances are also independent from each other... but it's a different kind of independence compared to e.g. all the completely unrelated things pulled in by multi-user.target... or networking-pre.target

kofemann:
To my understanding, .service is still expects an executable (that's why you have bin/true there), where .target is pure grouping, a shortcut. Here for example a similar concept for ceph

https://documentation.suse.com/ses/6/html/ses-all/ceph-operating-services.html

Concerning process handling. dcache starts a single java process which directly bound to systems:
```
[nairi] ~/eProjects/dcache $ sudo systemctl status dcache@coreDomain.service
● dcache@coreDomain.service - dCache coreDomain domain
     Loaded: loaded (/run/systemd/generator/dcache@coreDomain.service; generated)
     Active: active (running) since Mon 2020-06-29 16:06:13 CEST; 9s ago
   Main PID: 62491 (java)
      Tasks: 30 (limit: 19031)
     Memory: 265.4M
     CGroup: /system.slice/system-dcache.slice/dcache@coreDomain.service
             └─62491 /usr/bin/java -server -Xmx512m -XX:MaxDirectMemorySize=512m ...
Jun 29 16:06:13 nairi systemd[1]: Started dCache coreDomain domain.
```
and for course systemd can handle restarts, this is on my todo list to make it configurable:

https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=

Of course, the systemd integration is not ideal, but we need more experienced admins to use it to collect feedback and adjust accordingly.

kofemann:
Of course, the way how we run dcache is not systemd friendly... Or the other way around. Systemd doesn't expect to run rumtiple instances of a service on a single host. The generator based approach allows us to offload lot of self written logic to systemd. An alternative can be a generation of first-class service units from puppet/ansible and co or let dcache just start a simple controller process and then pull from a central location the configuration.

Or we offload everything to today's hype - containers and it's orchestration. However, according to our survey, none of the sites are planning to run dcache in a containerized  environment.

calestyo:
Yes, I think .service needs a binary,... but it seems usual to use /bin/true for this (e.g. Postgresql and others do it that way).

> Concerning process handling. dcache starts a single java process which directly bound to systems

Maybe my memory cheated me there... but I thought systemd woudn't have noticed if starting dcache would have actually failed e.g. because of config errors.
In fact it seems the java process is started directly by systemd from the generator units... or is there still any loader code involved?!

> and for course systemd can handle restarts, this is on my todo list to make it configurable:
> https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=

Yes... nice... and I'd really suggest that for such thing the configuration should also take place in systemd... and not kinda hack systemd to use the dcache settings... i.e. I also wouldn't try to make systemd use dcache.user to set User=.

If one knows how, it's also fairly easy to just override single settings like User= for one single unit (even for generated ones).
So we just need to tell people in the release notes what they should do now, if they e.g. used dcache.user before.


> Systemd doesn't expect to run rumtiple instances of a service on a single host.

Why not? With the generators that works pretty nice... and actually systemd serves such units out of the box (e.g. getty - which btw is in fact a target, I mean the "parent" unit... so I was wrong above).
Any I now quite a number of programs that do this... e.g. postgresql in Debian is pretty powerful with respect to that, supporting not only different DB servers but also different versions to all run as postgresql@*.services.


I'd also strongly suggest to continue using the generator based approach... (just perhaps doing it technically different... (see #3630).
I really think that dCache's concept of domains (be they of different types or of the same like multiple doors of the same type) just perfectly fits the dcache@domain.service idea.


As for whether the "parent" unit should be a service or a target... I'm really not sure myself... technically, both works of course... but .targets are vaguely defined in the manpage... in any case it's just a cosmetic question.
If you want a definite answer, it's perhaps best to describe the situation to the systemd developers and ask them what they'd recommend to use.

I just searched the whole Debian archive for which targets exist in all of its packages:
```
389-ds-base: /lib/systemd/system/dirsrv.target
ceph-common: /lib/systemd/system/ceph.target
ceph-fuse: /lib/systemd/system/ceph-fuse.target
ceph-mds: /lib/systemd/system/ceph-mds.target
ceph-mgr: /lib/systemd/system/ceph-mgr.target
ceph-mon: /lib/systemd/system/ceph-mon.target
ceph-osd: /lib/systemd/system/ceph-osd.target
cloud-init: /lib/systemd/system/cloud-config.target
cloud-init: /lib/systemd/system/cloud-init.target
friendly-recovery: /lib/systemd/system/friendly-recovery.target
gitlab: /lib/systemd/system/gitlab.target
knot-resolver: /lib/systemd/system/kresd.target
knot-resolver: /lib/systemd/system/multi-user.target.wants/kresd.target
libvirt-daemon-system: /lib/systemd/system/virt-guest-shutdown.target
nfs-common: /lib/systemd/system/nfs-client.target
obs-api: /lib/systemd/system/obsapi.target
openssh-server: /lib/systemd/system/rescue-ssh.target
pki-server: /lib/systemd/system/pki-tomcatd-nuxwdog.target
pki-server: /lib/systemd/system/pki-tomcatd.target
policycoreutils: /lib/systemd/system/selinux-autorelabel.target
radosgw: /lib/systemd/system/ceph-radosgw.target
rbd-mirror: /lib/systemd/system/ceph-rbd-mirror.target
rdma-core: /lib/systemd/system/rdma-hw.target
resource-agents: /lib/systemd/system/resource-agents-deps.target
systemd: /lib/systemd/system/basic.target
systemd: /lib/systemd/system/blockdev@.target
systemd: /lib/systemd/system/bluetooth.target
systemd: /lib/systemd/system/boot-complete.target
systemd: /lib/systemd/system/cryptsetup-pre.target
systemd: /lib/systemd/system/cryptsetup.target
systemd: /lib/systemd/system/ctrl-alt-del.target
systemd: /lib/systemd/system/default.target
systemd: /lib/systemd/system/emergency.target
systemd: /lib/systemd/system/exit.target
systemd: /lib/systemd/system/final.target
systemd: /lib/systemd/system/getty-pre.target
systemd: /lib/systemd/system/getty.target
systemd: /lib/systemd/system/graphical.target
systemd: /lib/systemd/system/halt.target
systemd: /lib/systemd/system/hibernate.target
systemd: /lib/systemd/system/hybrid-sleep.target
systemd: /lib/systemd/system/initrd-fs.target
systemd: /lib/systemd/system/initrd-root-device.target
systemd: /lib/systemd/system/initrd-root-fs.target
systemd: /lib/systemd/system/initrd-switch-root.target
systemd: /lib/systemd/system/initrd.target
systemd: /lib/systemd/system/kexec.target
systemd: /lib/systemd/system/local-fs-pre.target
systemd: /lib/systemd/system/local-fs.target
systemd: /lib/systemd/system/multi-user.target
systemd: /lib/systemd/system/multi-user.target.wants/getty.target
systemd: /lib/systemd/system/network-online.target
systemd: /lib/systemd/system/network-pre.target
systemd: /lib/systemd/system/network.target
systemd: /lib/systemd/system/nss-lookup.target
systemd: /lib/systemd/system/nss-user-lookup.target
systemd: /lib/systemd/system/paths.target
systemd: /lib/systemd/system/poweroff.target
systemd: /lib/systemd/system/printer.target
systemd: /lib/systemd/system/reboot.target
systemd: /lib/systemd/system/remote-cryptsetup.target
systemd: /lib/systemd/system/remote-fs-pre.target
systemd: /lib/systemd/system/remote-fs.target
systemd: /lib/systemd/system/rescue.target
systemd: /lib/systemd/system/rpcbind.target
systemd: /lib/systemd/system/runlevel0.target
systemd: /lib/systemd/system/runlevel1.target
systemd: /lib/systemd/system/runlevel2.target
systemd: /lib/systemd/system/runlevel3.target
systemd: /lib/systemd/system/runlevel4.target
systemd: /lib/systemd/system/runlevel5.target
systemd: /lib/systemd/system/runlevel6.target
systemd: /lib/systemd/system/shutdown.target
systemd: /lib/systemd/system/sigpwr.target
systemd: /lib/systemd/system/sleep.target
systemd: /lib/systemd/system/slices.target
systemd: /lib/systemd/system/smartcard.target
systemd: /lib/systemd/system/sockets.target
systemd: /lib/systemd/system/sound.target
systemd: /lib/systemd/system/suspend-then-hibernate.target
systemd: /lib/systemd/system/suspend.target
systemd: /lib/systemd/system/swap.target
systemd: /lib/systemd/system/sysinit.target
systemd: /lib/systemd/system/sysinit.target.wants/cryptsetup.target
systemd: /lib/systemd/system/system-update-pre.target
systemd: /lib/systemd/system/system-update.target
systemd: /lib/systemd/system/time-set.target
systemd: /lib/systemd/system/time-sync.target
systemd: /lib/systemd/system/timers.target
systemd: /lib/systemd/system/umount.target
systemd-container: /lib/systemd/system/machines.target
systemd-cron: /lib/systemd/system/cron-daily.target
systemd-cron: /lib/systemd/system/cron-hourly.target
systemd-cron: /lib/systemd/system/cron-monthly.target
systemd-cron: /lib/systemd/system/cron-weekly.target
systemd-cron: /lib/systemd/system/cron.target
wireguard-tools: /lib/systemd/system/wg-quick.target
zfsutils-linux: /lib/systemd/system/zfs-import.target
zfsutils-linux: /lib/systemd/system/zfs-volumes.target
zfsutils-linux: /lib/systemd/system/zfs.target
```

If you take away the ones directly from systemd, there are not that many left. Most of them, AFAICS, are rather abstracts "things"... I mean e.g. even openssh's rescue-ssh.target is actually more like a system state, in the sense of: "don't go to multi-user or graphical.target... but go to a rescue-ssh interface" ...

zfs/nfs/ceph also use targets... not sure why ceph does it,... but for zfs/nfs I'd guess they consider the target also like "system state/synchronisation unit" similar to networking.target... i.e. from that point on, NFS is available... respectively from that point on ZFS mounts will work.

For dCache this applies IMO less, as dCache isn't a typical filesystem for the local host (where dCache runs on).

But again,... this is really just cosmetics.


Cheers,
Chris.

kofemann:
Digging more in this topic.... Systemd supports template units

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/chap-managing_services_with_systemd#sect-Managing_Services_with_systemd-Instantiated_Units

This means that we can have a single unit file that will be used to start
all dcache services, e.g no generator is needed.
This is of course more work, thus let's keep it for future enhancement.


-kofemann

calestyo:
> Digging more in this topic.... Systemd supports template units

Aren't these just the ones created by a generator...or at least the same class of, i.e. ones with an @ .. and the only difference was that template really has a plain file foo@.service associated with it.

I mean that was, IIRC, just the conclusion I drew in #3630, that we need such a template, even if it were empty and the generator creates the actual "sub-units" which then kinda override the template unit.

But if you guys decide that we anyway go the systemd-way and drop thing like dcache.user... it might be just enough to have the template alone (which would be pretty simple then) and no need for a generator at all.
With %i, IIRC, you can refer the instance portion in the template unit, so that would be the domain in our case.

It could get tricky though with things like ${DCACHE_DEFAULTS} since you add these to the executed java... but we cannot directly access it from systemd.
I don't think any of the Environment*= options could be used for that, as we'd need to invoke the parser... so maybe the solution would be to pull these out of daches own config as well (which I think would be the clean way).


> This is of course more work, thus let's keep it for future enhancement. 

Well much work on the dcache side (i.e. pull out all the stuff)... but on the systemd side, I guess we could more or less take the unit file that is now made by the generator.

Cheers,
Chris.

kofemann:
The only benefit I can see is shorter output of `systemctl list-dependencies command`

```

$ systemctl list-dependencies dcache.service
dcache.service
? ??system.slice
? ??basic.target
?   ??firewalld.service
?   ??microcode.service
?   ??rhel-dmesg.service
?   ??selinux-policy-migrate-local-changes@targeted.service
?   ??paths.target
?   ??slices.target
?   ? ??-.slice
?   ? ??system.slice
?   ??sockets.target
?   ? ??avahi-daemon.socket
?   ? ??dbus.socket
?   ? ??dm-event.socket
?   ? ??rpcbind.socket
?   ? ??systemd-initctl.socket
?   ? ??systemd-journald.socket
?   ? ??systemd-shutdownd.socket
?   ? ??systemd-udevd-control.socket
?   ? ??systemd-udevd-kernel.socket
?   ??sysinit.target
?   ? ??dev-hugepages.mount
?   ? ??dev-mqueue.mount
?   ? ??kmod-static-nodes.service
?   ? ??lvm2-lvmetad.socket
...

$ systemctl list-dependencies dcache.target
dcache.target
? ??dcache@coreDomain.service
? ??dcache@gplazmaDomain.service
? ??dcache@namespaceDomain.service
? ??dcache@nfsDomain.service
? ??dcache@poolDomain.service
? ??dcache@poolmanagerDomain.service
$
```
discarding.

calestyo:
> systemctl list-dependencies command

Hmm but that would indicate via colours, whether the units actually run and thus be a partial replacement for dcache status.


Well... I really don't know whats better.... my suggestion... ask at the upstream mailing list what they think is the cleanest way :-)

kofemann:
https://lists.freedesktop.org/archives/systemd-devel/2020-July/044839.html

kofemann:
poking more around looks like using **target** is recommended way for grouping. Moreover, newer versions of systemd support reload as well. @calestyo raise or objection before merge it into master.

calestyo:
Well if that's what upstream recommends, then we should go for it. :-)

PR #5477, opened 2020-07-09
systemd: convert dcache.service into dcache.target

kofemann:
Motivation:
the systemd service are expect to run processes. The dcache.service,
however, there to group per-domain units together. The original
reason to make it a service was an ability propagate reloads to the
set of dependent units. This makes no sense in case of dcache as (a)
domain processes don't depend on each other and (b) dcache does not
support dynamic configuration reload.

Moreover, using target for grouping is more mainstream approach
(ceph, vmware).

Modification:
change dcache.service into dcache.target which is more suitable for
dcache domain grouping.

Result:
more systemd friendly deployment.

NOTE: this change breaks existing (debian) installations, but impact is
small. Now admins have to run

systemctl start dcache.target

instead of

systemctl start dcache.service

Acked-by: Lea Morschel
Target: master
Require-book: yes
Require-notes: yes
(cherry picked from commit b1442851a03b7d50e471377cbeea1fbe2e03571e)


kofemann:
retest this please

kofemann:
retest this please

PR #5482, opened 2020-07-16
pool.properties comment: load distribution with sweeper-margin

onnozweers:
Thanks to Dmitry for explaining this in the user forum.


kofemann:
Thanks, Onno!

PR #5548, opened 2020-09-10
srm-client: drop support for SRM v1.1

paulmillar:
Motivation:

The SRM v1.1 protocol is *long* dead.

Support in dCache for v1.1 was made optional and disabled by default in
2015 (dCache v2.14) and support for SRM v1.1 dropped altogether in 2016
(dCache v2.16).

This code simplication is in preparation for adding support bearer
tokens.

Modification:

Drop all SRM v1.1-only commands.

Drop configuration that allows selection of SRM v1.1 or SRM v2.2.

Drop resulting dead-code.

Result:

srm-client no longer supports SRM v1.1.

Target: master
Request: 6.2
Requires-notes: yes (srm-client)
Requires-book: no
Patch: https://rb.dcache.org/r/12570/
Acked-by: Tigran Mkrtchyan


kofemann:
Why this should go into 6.2?

paulmillar:
I would like to release a version of srm-client that supports bearer token authentication.

Because srm-client releases are bound to dCache releases, the only way of doing that is as part of a dCache release.  If we are to release a new srm-client "soon" (i.e., before v7.0.0 is released) then I'll need to back-port patches to the latest stable branch (6.2) so that they may be included in the latest release.

The first version of the patch adding bearer token support made only the minimum changes.  It added the bearer token support (for SRM v2.2 commands only) and left the rest of the srm-client code otherwise unchanged.

This patch was rejected by Dmitry as it was too big.

The patch's size was a natural consequence of how SRM v2.2 supported was implemented in srm-client: with a fair amount of copy-and-paste style code duplication.  This, in turn, was a result of how SRM v2.2 support was "bolted on" on top of the existing SRM v1.1 support.

Since the SRM v1.1 support was no longer needed and was causing problems, I felt it was time to get rid of it.

However, this means that three patches are needed to support bearer tokens:
  1. this patch, which drops SRM v1.1 support;
  2. a patch refactoring how SRM v2.2 is supported, making code more DRY;
  3. a third patch, adding bearer token support.

Each patch depends on the previous one: the final patch (patch 3) depends on patch 2 and patch 2 depends on this patch.

I have tried to make each of these three patches as small as possible; however, it's certainly true that, together, the three patches are larger than the original, single patch.

kofemann:
The 7.0 is expected somewhere in October... The users of srm-client can wait.

paulmillar:
or in November (like with v3.0.0), or in December (v4.0.0), or in January (v5.0.0).  That said, v6.0.0 was released in October, so there is hope to get this out before 2021.

I was hoping to provide people with an official dCache srm-client release, particularly for the AuthZ hackathon this week, and more generally, to use when verifying a corresponding gfal update.

However, if the official release isn't happening quickly, I can simply provide people an unofficial version that people can use in the meantime.

PR #5559, opened 2020-09-28
pool: http-tpc PUT request are repeatable

paulmillar:
Motivation:

If the remote server takes too long to respond to an HTTP PUT request
that uses the expect-100-continue negotiation then the pool for will
spontaneously send the entity.  This behaviour is standards-compliant
and common across all HTTP client libraries, not just Apache library
used by the pool.  The pool's timeout is just three seconds.

(For dCache, the WebDAV door receiving the HTTP PUT request will likely
redirect the client to the pool.  This can take longer than three
seconds, especially if pools are busy and the mover is queued.)

Therefore, if the receiving dCache door takes too long to reply then the
client will sent the file's content to the door, only to receive the
redirection response.  This then forces the pool to send the file's
content again, which it currently cannot do.  The transfer then fails.

Modification:

Introduce a new HttpEntity class that allows the pool to send the file
content multiple times, as a result of redirection responses.

Result:

The HTTP-TPC PUSH requests (which use the HTTP PUT request to send data)
are now more robust against slow remote servers that redirect the
transfer.

Target: master
Request: 6.2
Request: 6.1
Request: 6.0
Request: 5.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/12536/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/cells/src/main/java/dmg/util/Exceptions.java
+++ b/modules/cells/src/main/java/dmg/util/Exceptions.java
@@ -19,7 +19,7 @@ public class Exceptions
     {
     }
 
-    private static String meaningfulMessage(Throwable t)
+    public static String meaningfulMessage(Throwable t)
     {
         return t.getMessage() != null ? t.getMessage() : t.getClass().getName();
     }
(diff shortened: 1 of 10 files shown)


mksahakyan:
retest this please

Issue #5618, opened 2020-11-18
WebDAV errors observed when executing several simultaneous PUT requests


airnandez:
Hello,

I would like to report some errors we are observing when uploading files to dCache using the WebDAV protocol. The problem occurs specifically when:

* the WebDAV door redirects the requests to the pools (when the client specifies the `Expect: 100-continue` header), and
* the number of simultaneous `PUT` requests goes beyond a threshold that in our configuration is about 30 requests

These are the versions we are running in our installation:

* dCache webDAV door: `dCache/5.2.33`
* dCache pools: `dCache/5.2.24`

The issue is easily reproducible using `curl` and here is an excerpt of the error messages we observe:

```
curl: (35) Encountered end of file
curl: (56) SSL peer reports incorrect Message Authentication Code.
curl: (35) Peer's certificate has an invalid signature.
curl: (35) SSL peer was not expecting a handshake message it received.
curl: (35) Encountered end of file
curl: (35) Encountered end of file
curl: (56) SSL peer reports incorrect Message Authentication Code.
curl: (35) SSL received a malformed Alert record.
curl: (35) SSL peer was not expecting a handshake message it received.
curl: (35) SSL peer was not expecting a handshake message it received.
curl: (35) SSL received a malformed Alert record.
curl: (35) SSL peer was not expecting a handshake message it received.
curl: (35) Encountered end of file
curl: (56) SSL peer reports incorrect Message Authentication Code.
curl: (35) Encountered end of file
curl: (56) SSL peer reports incorrect Message Authentication Code.
curl: (35) Encountered end of file
curl: (35) SSL peer was not expecting a handshake message it received.
curl: (35) SSL received a malformed Alert record.
curl: (35) SSL received a malformed Alert record.
curl: (56) SSL peer reports incorrecct Mesusarge Authentication Code.
```

You can find a shell script to reproduce the issue [here](https://gist.github.com/airnandez/ee641fb0fafc0b473d2fad9ff5d3e17b) together with the indications on how to run it.

In case it matters, we are using grid certificates (as opposed to bearer tokens) for authenticating to both the dCache WebDAV door and the pools.

@ageorget administers our dCache instance and could provide relevant details of our specific configuration, if needed.

cc: @bgounon

airnandez:
Supplemental information. We also observe errors using simultaneous `GET` requests beyond a threshold. Here is an excerpt of the errors we observe when using 64 simultaneous requests:

```
[...]

PR #5624, opened 2020-11-30
pool: https avoid sharing SSLEngine between TCP connections

paulmillar:
Motivation:

Problems have been reported when dCache is configured to redirect
clients to pools with encryption enabled, and dCache is in heavy use
with multiple concurrent uploads.

The problem stems from an SSLEngine instance being shared between
multiple TCP connections.

An SSLEngine is intended to be used once per TCP connection; however,
currently the HttpsTransferService generates a single SSLEngine instance
when Netty server starts and uses this for all TCP connections.

Modification:

Create a new SSLEngine when accepting a new TCP connection.  This
SSLEngine has a lifetime of the TCP connection.

Result:

Fix a problem where, under heavy load, redirected https transfers result
in garbled SSL information, preventing the HTTP client from working.

Target: master
Requires-notes: yes
Requires-book: no
Request: 7.0
Request: 6.2
Request: 6.1
Closes: #5618
Patch: https://rb.dcache.org/r/12676/
Acked-by: Lea Morschel

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/http/HttpsTransferService.java
+++ b/modules/dcache/src/main/java/org/dcache/http/HttpsTransferService.java
@@ -1,6 +1,6 @@
 /* dCache - http://www.dcache.org/
  *
- * Copyright (C) 2017-2018 Deutsches Elektronen-Synchrotron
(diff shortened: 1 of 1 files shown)


mksahakyan:
retest this please

PR #5635, opened 2020-12-14
pool: sort flush queue only if we going to use the result

kofemann:
Motivation:
Even if flush number of active flushes is configured to be zero, the
StorageClassInfo#internalFlush will sort the requests, but discard the
result.

Modification:
Update StorageClassInfo#internalFlush to sort the flush queue right before
the sorted result is needed. Use Java8 stream to build the sorted sublist.

Result:
Reduce the number of useless CPU intensive operations and lock starvation.

Target: master, 7.0, 6.2
Acked-by: Lea Morschel
Acked-by: Paul Millar
(cherry picked from commit 91cfaa5d3f1ca6031aeec6baf533efda0138526c)


mksahakyan:
retest this please

Issue #5642, opened 2021-01-07
Bad timeout management in TransferManager


paulmillar:
For HTTP-TPC, WebDAV door makes periodic queries to (Remote)TransferManager to learn the current status of a transfer.  These queries are done periodically, controlled by `webdav.third-party-transfers.performance-marker-period`, which defaults to 5 seconds.

The WebDAV door gives RemoteTransferManager half that (so 2.5 seconds, by default) to respond to its query asking the current status of the mover.

If the transfer has started, RemoteTransferManager will contact the pool to query the mover's status.  This query currently uses a hard-coded timeout (30 seconds).

This is undesirable as, if the pool is overloaded, queued queries on the pool are processed even though the value will be discarded (creating unnecessary load), there is additional (unnecessary) logging as the final response from TransferManager is discarded (for taking too long), and the WebDAV door reports a generic problem (timeout taking to TransferManager) rather than showing the underlying problem.

These issues should be fixed if the TransferManager were updated so that the pool query timeout is based on the remaining lifetime of the query from the WebDAV door.  This updated message timeout may need to include a factor that takes into account overheads from TransferManager's reponse to WebDAV (if the pool message times out) not being instantaneous; for example, by decreasing the TTL of the pool message.

paulmillar:
The complete fix requires two patches.  One has been back-ported to the supported branches.  The other has only been committed to `master` branch.  Therefore, this issue is resolved completely when a site upgrades to dCache v7.1

PR #5651, opened 2021-01-12
TransferManager: choose better pool query message TTL

paulmillar:
Motivation:

The TransferManager provides the ability to query the current status of
a transfer, which is used by the WebDAV door to return HTTP-TPC progress
markers.

The progress markers have a strict deadline within which they need to
know the transfer's status.  The message from WebDAV to TransferManager
querying the transfer's status has a TTL that reflects this requirement.

However, the query from TransferManager to the pool has a hard-coded
timeout of 30 seconds.  This timeout is far too long.

Modification:

Update TransferManager so that the timeout (and corresponding TTL) for
the pool-querying message is based on the TTL of the incoming query from
the WebDAV door.

Result:

Better recovery in the case where a pool's message queue is overloaded,
as old (and now useless) requests may be simply discarded.

This patch partially addresses #5642.

Target: master

Request: 7.0
Request: 6.2
Request: 6.1
Request: 6.0
Request: 5.2
Requires-notes: no
Requires-book: no
Patch: https://rb.dcache.org/r/12753/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache/src/main/java/diskCacheV111/services/TransferManager.java
+++ b/modules/dcache/src/main/java/diskCacheV111/services/TransferManager.java
@@ -255,7 +255,8 @@ public abstract class TransferManager extends AbstractCellComponent
             return message;
         }
 
-        return handler.appendInfo(message);
+        long poolQueryTimeout = Math.min(envelope.getAdjustedTtl(), 30_000);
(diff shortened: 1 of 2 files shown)


mksahakyan:
retest this please

PR #5700, opened 2021-02-02
dcache-frontend: change transfer rate to compute what it advertised

alrossi:
Motivation:

dcache-view currently displays MB/sec for transfer rate, but the
value is the old KB/sec.

Modification:

Make the auxiliary method on TransferInfo provide MB/sec, which
is more useful.  Retain KB/sec in old page by allowing
it to compute KB/sec there.

Result:

Pages display the values advertised.

Target: master
Patch: https://rb.dcache.org/r/12801
Acked-by: Dmitry
Request: 7.0
Request: 6.2
Request: 6.1
Request: 6.0
Request: 5.2


mksahakyan:
retest this please

PR #5708, opened 2021-02-02
frontend: create a log entry when we return 500 Internal Server Error

paulmillar:
Motivation:

There are situations where dCache frontend will return 500 Internal
Server Error to the client.  However, when this happens, there is a
reasonable expectation that the log file contains details of what went
wrong.

Currently this doesn't happen: frontend logs nothing.  This can make
diagnosing and fixing such problems quite tricky.

Modification:

Add logging so frontend records the problem before throwing
InternalServerErrorException.

Note: this patch deliberately does not modify ErrorResponseProvider
because throwing InternalServerErrorException should only be about
creating the error response for the client, dCache-internal logging
shouldn't happen as a side-effect of this.

The exception handling in QosManagement is simplified to show that it
doesn't throw InternalServerErrorException (or CacheException for that
matter).

BulkServiceCommunicator is updated to propagate any RuntimeException or
Error (e.g,. OutOfMemoryError).  These would otherwise be suppressed.

Result:

Frontend now logs any time it returns 500 Internal Server Error to the
client.

Target: master
Requires-notes: yes
Requires-book: no
Request: 7.0
Request: 6.2
Request: 6.1
Request: 6.0
Request: 5.2
Patch: https://rb.dcache.org/r/12800/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache-frontend/src/main/java/org/dcache/restful/events/streams/metronome/Metronome.java
+++ b/modules/dcache-frontend/src/main/java/org/dcache/restful/events/streams/metronome/Metronome.java
@@ -145,8 +145,9 @@ public class Metronome implements EventStream
             String msg = index == -1 ? e.getMessage() : e.getMessage().substring(0, index);
             throw new BadRequestException("Bad selector value: " + msg);
         } catch (IOException e) {
-            throw new InternalServerErrorException("Unable to process selector: "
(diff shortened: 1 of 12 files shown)


mksahakyan:
retest this please

Issue #5697, opened 2021-02-01
Canceled webdav transfers spamming transfermanagers log


ageorget:
Hi,

I found a lot a repetitive log messages in transfermanagers log file caused by canceled webdav TPC transfers. 
The only way to remove these messages is to restart the webdav door.

`01 Feb 2021 09:33:01 (RemoteTransferManager) [door:webdav-ccdcatli310@webdav-ccdcatli310Domain:AAW56IFr54g webdav-ccdcatli310 CancelTransfer] cannot find handler with id=1611783138446000 for CancelTransferMessage
01 Feb 2021 09:33:02 (RemoteTransferManager) [door:webdav-ccdcatli267@webdav-ccdcatli267Domain:AAW56VbLGpg webdav-ccdcatli267 CancelTransfer] cannot find handler with id=1611786718230000 for CancelTransferMessage
01 Feb 2021 09:33:02 (RemoteTransferManager) [door:webdav-ccdcatli267@webdav-ccdcatli267Domain:AAW56VbFABg webdav-ccdcatli267 CancelTransfer] cannot find handler with id=1611786717830000 for CancelTransferMessage
01 Feb 2021 09:33:02 (RemoteTransferManager) [door:webdav-ccdcatli267@webdav-ccdcatli267Domain:AAW56VbQqHg webdav-ccdcatli267 CancelTransfer] cannot find handler with id=1611786718594000 for CancelTransferMessage
01 Feb 2021 09:33:06 (RemoteTransferManager) [door:webdav-ccdcatli310@webdav-ccdcatli310Domain:AAW56IFr54g webdav-ccdcatli310 CancelTransfer] cannot find handler with id=1611783138446000 for CancelTransferMessage
01 Feb 2021 09:33:07 (RemoteTransferManager) [door:webdav-ccdcatli267@webdav-ccdcatli267Domain:AAW56VbQqHg webdav-ccdcatli267 CancelTransfer] cannot find handler with id=1611786718594000 for CancelTransferMessage
01 Feb 2021 09:33:07 (RemoteTransferManager) [door:webdav-ccdcatli267@webdav-ccdcatli267Domain:AAW56VbLGpg webdav-ccdcatli267 CancelTransfer] cannot find handler with id=1611786718230000 for CancelTransferMessage
[...]

PR #5716, opened 2021-02-03
webdav-transfermanager: fail gracefully if transfermanager restarted

paulmillar:
Motivation:

Restarting the transfermanager with on-going HTTP-TPC transfers results
in transfermanager loosing knowledge of those transfers.  If the client
(e.g., FTS) then aborts the transfer then the WebDAV door contact
transfermanager to cancel the transfer, a request that transfermanager
logs but otherwise ignores.

The result is an endless loop, spamming the log file with entries like:

    cannot find handler with id=1611786718594000 for CancelTransferMessage

every time the WebDAV door tries to send a performace report (typically
5 seconds).

Modification:

Have transfermanager throw an exception if the request to cancel a
transfer targets an unknown transfer.

Have WebDAV react to that specific exception by immediately failing the
transfer.

Result:

Restarting transfermanager while there are ongoing HTTP-TPC transfers no
longer results in an endless loop, with transfermanager logging a
message every ~5 seconds per ongoing HTTP-TPC transfer.

Target: master
Request: 7.0
Request: 6.2
Request: 6.1
Request: 6.0
Request: 5.2
Requires-notes: yes
Requires-book: no
Closes: #5697
Patch: https://rb.dcache.org/r/12804/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/RemoteTransferHandler.java
+++ b/modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/RemoteTransferHandler.java
@@ -483,6 +483,15 @@ public class RemoteTransferHandler implements CellMessageReceiver
                 message.setExplanation("client went away");
                 try {
                     _transferManager.sendAndWait(message);
+                } catch (MissingResourceCacheException e) {
+                    /* Tried to cancel a transfer, but the transfer-manager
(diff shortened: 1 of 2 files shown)


mksahakyan:
retest this please

mksahakyan:
retest this please

PR #5765, opened 2021-03-01
common: Update BearerToken representation to include some information

paulmillar:
Motivation:

Commit c45da26e27 rightly stopped the logged output from including the
complete token.  Logging the complete token is a security risk.

However, it's still useful sometimes to understand with which token a
user is attempting to authenticate.  With c45da26e27, this is no longer
possible.

Modification:

Add a new scheme for representing a token.

For tokens larger than 8 characters (which is most tokens), the first
four and last four characters are shown.  The remaining (middle) portion
is hashed and the hash value is included.

The first four characters is often sufficient to understand what kind of
token is being presented.  Many types of token (e.g., JWT, macaroons)
have an internal structure where a cryptographic hash (used to ensure
the integrity of the token) is placed at the end of the token.
Therefore the last four characters of the token have a high likelihood
of being unique between different tokens.  These two small parts of the
token, together, should allow an admin to easily check the logged
details against some full token, which was obtained some other way.

By including a (cryptographically strong) hashed middle value, it is
highly likely that two different tokens will have different
representations.  Although the admin will not know the complete token,
it should be obvious when two tokens are (very likely) the same, and
when they are different.

For tokens with 8 characters or less, just a hash of the token is used.

A public method is added to allow other dCache code to provide the safe
representation of a bearer token.

The toString method is updated to use this representation.

JavaDoc is added to describe the expectation that the result from the
getToken method is never logged.

Result:

dCache updates how it represents bearer tokens (e.g., in log files) so
that it provides sufficient information to determine whether tokens are
the same, but without revealing the complete token.

Target: master
Requires-notes: yes
Requires-book: no
Request: 7.0
Request: 6.2
Request: 6.1
Request: 6.0
Request: 5.2
Patch: https://rb.dcache.org/r/12864/
Acked-by: Tigran Mkrtchyan
Acked-by: Lea Morschel

### DIFF
--- a/modules/common/src/main/java/org/dcache/auth/BearerTokenCredential.java
+++ b/modules/common/src/main/java/org/dcache/auth/BearerTokenCredential.java
@@ -1,10 +1,13 @@
 package org.dcache.auth;
 
 import com.google.common.base.CharMatcher;
+import com.google.common.hash.Hashing;
 
 import java.io.Serializable;
+import java.util.Base64;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static java.nio.charset.StandardCharsets.US_ASCII;
 
 public class BearerTokenCredential implements Serializable
 {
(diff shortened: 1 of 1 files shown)


paulmillar:
retest this please

PR #5835, opened 2021-04-08
dcache-xroot: fix memory leak, race and other small issues with recon…

alrossi:
…nect to pool

Motivation:

Patch: https://rb.dcache.org/r/12938/
Committed: master@585ffa893b95e0fdad4fb7e1c7201580a2a6de20

attempted to fix a reconnect issue on the pools.
This patch corrects a slow memory leak and a potential
mishandling of a race condition between client and server.
It also makes a better effort to prevent/clean up
orphaned timers.

Modification:

1.  do not call release ALL on expiration or cancel, only
    on close;
2.  call timer.cancel in the TimerTask run method; otherwise,
    the Timer run method goes into indefinite wait and
    it/its Thread will not be garbage collected;
3.  cancel the timer on close;
4.  do not start the timer if the mover has already been
    released, i.e., removed from the mover map.

Also added a little more (debug) logging and renames
two methods.

Result:

Bugs eliminated.

Target: master
Request: 7.0
Request: 6.2
Request: 6.1
Request: 6.0
Request: 5.2
Patch: https://rb.dcache.org/r/12964/
Acked-by: Lea

### DIFF
--- a/modules/dcache-xrootd/src/main/java/org/dcache/xrootd/pool/XrootdPoolRequestHandler.java
+++ b/modules/dcache-xrootd/src/main/java/org/dcache/xrootd/pool/XrootdPoolRequestHandler.java
@@ -358,7 +358,8 @@ public class XrootdPoolRequestHandler extends AbstractXrootdRequestHandler
             /*
              *  Stop any timer in case this is a reconnect.
              */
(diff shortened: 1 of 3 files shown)


mksahakyan:
retest this please

PR #5858, opened 2021-04-22
gplazma: ldap throw exception if no principal is added

paulmillar:
Motivation:

The gPlazma Mapping plugin contract is that a mapping plugin should
either add (at least) one principal or it has failed.  A failed mapping
plugin throws an AuthenticationException to indicate why it failed.

Currently, the ldap plugin never throws an AuthenticationException, even
when it has added no principals.  This makes it hard to use the ldap
plugin with other mapping plugins.

Modification:

Update plugin to throw the expected exception if it fails to add any
principals.  The exception's message provides a short explanation on why
the mapping request failed.

Result:

The LDAP plugin behaviuor now more closely follows that of other mapping
plugins.  This allows deployments where LDAP is tried first and, if that
fails to identify the user, fall-back strategies are used.

Target: master
Request: 7.0
Request: 6.2
Request: 6.1
Request: 6.1
Request: 6.0
Request: 5.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/13000/
Acked-by: Albert Rossi
Acked-by: Tigran Mkrtchyan
Acked-by: Lea Morschel


mksahakyan:
retest this please

mksahakyan:
retest this please

Issue #5870, opened 2021-04-23
frontend/api: readers and writers always zero


onnozweers:
Hi all,

I'm running something like this to collect pool mover info:

```
curl ... -X GET https://dcacheview.grid.surfsara.nl:22881/api/v1/poolgroups/atlas_writediskpools/queues \
| jq -r '.costDataForPools | .[] | .extendedMoverHash.gridftp_queue'
```
I get something like this:
```
{
  "name": "gridftp_queue",
  "active": 2,
  "maxActive": 24,
  "queued": 0,
  "readers": 0,
  "writers": 0
}
{
  "name": "gridftp_queue",
  "active": 0,
  "maxActive": 24,
  "queued": 0,
  "readers": 0,
  "writers": 0
}
{
  "name": "gridftp_queue",
  "active": 1,
  "maxActive": 24,
  "queued": 0,
  "readers": 0,
  "writers": 0
}
...
```
I would expect that readers + writers = active, but this is clearly not the case. Is my understanding incorrect, or is there a bug that prevents values for readers & writers to be passed on?

Kind regards,
Onno

alrossi:
Hi Onno,

I am wondering too whether this has ever worked.   Investigating and will get back.

Al

alrossi:
Onno,

a quick check from production here at Fermi.  Here is the beginning of read-write pool group data:

```{
  "costDataForPools": {
    "rw-stkendca1801-1": {
      "label": "Pool Cost Info",
      "mover": {
        "active": 14,
        "maxActive": 3730,
        "queued": 0,
        "readers": 0,
        "writers": 0
      },
      "store": {
        "active": 40,
        "maxActive": 0,
        "queued": 69,
        "readers": 0,
        "writers": 109
      },
      "restore": {
        "active": 40,
        "maxActive": 0,
        "queued": 150,
        "readers": 0,
        "writers": 190
      },
```

It looks like readers and writers are both active and queued.

The base code which populates named queues (from the Pool class):

```
private static PoolCostInfo.NamedPoolQueueInfo toNamedPoolQueueInfo(MoverRequestScheduler queue)
    {
        return new PoolCostInfo.NamedPoolQueueInfo(queue.getName(),
                                                   queue.getActiveJobs(),
                                                   queue.getMaxActiveJobs(),
                                                   queue.getQueueSize(),
                                                   queue.getCountByPriority(IoPriority.REGULAR),
                                                   queue.getCountByPriority(IoPriority.HIGH));
    }

[...]

PR #5872, opened 2021-04-29
poolV4: fix reader/writer counts

alrossi:
Motivation:

Bug, GitHub #5870
frontend/api: readers and writers always zero

Reporting of readers and writers has two issues:
- only the queued movers are counted for named queues;
- these are counted on the basis of IOPriority.

Modification:

Look at the open mode of the file in mover to determine
whether this is read or write; scan all jobs, not just
those that have been queued.

Result:

Counts of readers and writers should now be correct.

Target: master
Request: 7.1
Request: 7.0
Request: 6.2
Request: 6.1
Request: 6.0
Request: 5.2
Patch: https://rb.dcache.org/r/13009/
Closes: #5870
Requires-notes: yes
Requires-book: no
Acked-by: Dmitry


mksahakyan:
retest this please

Issue #5883, opened 2021-04-30
CORS doesn't seem to work


onnozweers:
Dear dCache devs,

I'm trying to get CORS to work in dCache 6.0.37. For this, I am experimenting with webdav.allowed.client.origins.

My config:
```
[webdav2880-${host.name}Domain]
dcache.java.options.extra = -Djdk.tls.ephemeralDHKeySize=2048
[webdav2880-${host.name}Domain/webdav]
webdav.cell.name=webdav2880-${host.name}
webdav.redirect.on-read=true
webdav.redirect.on-write=true
webdav.redirect.allow-https=true
webdav.authn.basic=true
webdav.authn.accept-client-cert = false
webdav.authn.protocol=https
webdav.net.port=2880
webdav.enable.overwrite=true
webdav.allowed.client.origins = https://dcacheview.grid.surfsara.nl:22880
webdav.loginbroker.tags=webdav2880
```
Then from the dCacheView at https://dcacheview.grid.surfsara.nl:22880 I try to download or display a file. In the Chrome console, I can see this fail with twice this error:

`Access to fetch at 'https://dcacheview.grid.surfsara.nl:2880/pnfs/grid.sara.nl/data/users/onno/disk/bloemen.jpg' from origin 'https://dcacheview.grid.surfsara.nl:22880' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.`

The console seems to show two requests listed in the console:

```
Request URL: https://dcacheview.grid.surfsara.nl:2880/pnfs/grid.sara.nl/data/users/onno/disk/bloemen.jpg
Referrer Policy: strict-origin-when-cross-origin
Provisional headers are shown
authorization: Basic **********************
content-type: image/jpeg
Referer: https://dcacheview.grid.surfsara.nl:22880/
suppress-www-authenticate: Suppress
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.93 Safari/537.36
```
I decoded the "authorization: Basic" value and it seems to be correct (username:password).

And the second:
```
OPTIONS /pnfs/grid.sara.nl/data/users/onno/disk/bloemen.jpg HTTP/1.1
Host: dcacheview.grid.surfsara.nl:2880
Connection: keep-alive
Accept: */*
[...]

PR #5891, opened 2021-05-05
webdav: use Jetty's built-in support for CORS

paulmillar:
Motivation:

Issue #5883 is the latest of several bug-reports that come from dCache's
custom CORS support.  In this case, the underlying problem is that the
WebDAV door passes on CORS preflight requests to the down-stream servlet
(Milton or frontend), in order to allow it to include additional
headers.

Since CORS requests are always unauthenticated, passing requests on to
the downstream servlet will only work if the resource is public, or
there is special code within those downstream serlvets to detect and
support CORS requests.

The precise problem described in issue #5883 comes from attempting to
access a protected resource in dCacheView, which uses WebDAV for the
actual data transfer.  The work-around in Milton to handle such requests
is broken.

Modification:

This patch does two things:

a) It hides CORS pre-flight requests from any downstream handlers.  All
   resources now receive a standard set of CORS headers.

b) It removes the dCache-custom implementation of CORS in favour of
   Jetty's implementation: CrossOriginFilter.

Unfortunately, as Jetty implemnets CORS as a Filter (not a Handler), we
can't use Spring to create an instance.  Instead, this patch keeps the
existing class (CrossOriginResourceSharingHandler), but now this acts
both as a simple bridge between Handler and Filter and is responsible
for instantiating and configuring CrossOriginFilter.

Result:

dCacheView is able to view and download files when authenticated and
when the WebDAV door proxies the transfer.

Target: master
Requires-notes: yes
Requires-book: no
Request: 7.1
Request: 7.0
Request: 6.2
Request: 6.1
Request: 6.0
Request: 5.2
Closes: #5883
Patch: https://rb.dcache.org/r/13017/
Acked-by: Tigran Mkrtchyan


svemeyer:
retest this please

PR #5940, opened 2021-06-02
User guide: bulk-request URLs contain /api/v1

onnozweers:
In the user guide, there is a list of supported bulk-requests. This list however has URLs without `/api/v1`. I tested it with a fresh snapshot and this doesn't work, but with the /api/v1 it works.


paulmillar:
Thanks, Onno!

PR #5953, opened 2021-06-29
chimerashell: add recursive option to 'chown' command

paulmillar:
Motivation:

Ownership is often consistent within some subtree; i.e., all files and
directories within a specific directory are the same.  Therefore, when
changing ownership (the `chown` command), it's not uncommon to want to
update the filesystem recursively.

Modification:

Add the `-R` option to the `chown` command.

The patch also includes some minor refactoring to avoid unnecessary
field members and to throw CommandException instead of
IllegalArgumentException.

Result:

Chimera shell now supports the `-R` option in the `chown` command.

Target: master
Request: 7.1
Request: 7.0
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/13036/
Acked-by: Lea Morschel


kofemann:
retest this please

mksahakyan:
retest this please

PR #5961, opened 2021-07-14
a few macaroon hands-on examples

thdesy:
parsed collected macaroon examples with curl from local wiki into markdown


kofemann:
Thanks @thdesy

PR #5990, opened 2021-08-02
gplazma: add util to convert authzdb file to omnisession

paulmillar:
Motivation:

To support the removal of authzdb plugin, we should provide an easy way
for sites to migrate their existing configuration from using authzdb (as
a session plugin) to omnisession.

Note: it is also possible to use the authzdb plugin as a mapping plugin.
This utility does not help in migrating that use to some alternative
solution.

Modification:

Add utility to convert a authzdb configuration file to the equivalent
omnisession plugin.

It should work out-of-the-box, but the source and destination paths may
be overridden.

The gplazma.conf file is NOT updated, but suggested instructions are
included.

Result:

Sites now have a utility to help them migrating from authzdb to
omnisession.

Target: master
Requires-notes: yes
Requires-book: perhaps
Patch: https://rb.dcache.org/r/13026/
Acked-by: Tigran Mkrtchyan
Request: 7.1
Request: 7.0
Request: 6.2


mksahakyan:
retest this please

Issue #3827, opened 2018-02-14
Modernize PoolStatisticsV0


kofemann:
The poolStatisticsV0 class periodically collects statistics and generates raw and html report files. By doing so it uses old java SimpleDateFormat class. Since java8, there is time java.time API which is thread  save and more efficient. A custom CellCron scheduler can be replaced ScheduledExecutorService.

enragedshamy:
I will give my best on this one

PR #6036, opened 2021-08-30
dcache: Modernize PoolStatisticsV0

iNeele:
Motivation:
The poolStatisticsV0 class periodically collects statistics and
generates raw and html report files. By doing so it uses the old
java SimpleDateFormat class. Since java8, there is the
java.time API available, which is thread-safe and more efficient.
Moreover, the class uses a custom CellCron-scheduler, which can
be replaced by a scheduledExecutorService.

Modification:
Implements a scheduledExecutorService and replaces the
SimpleDateFormat class with the DateTimeFormatter class of the
java.time.format package in poolStatisticsV0.

Result:
Using the ScheduledExecutorService allows the CellCron-class in
the module cells to be removed. The exchange of the time-classes
results in more thread-safe and efficient code.
No user observable changes.

Issue: #3827
Target: master


modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
```diff
@@ -148,87 +148,87 @@
 
     private static final String DEFAULT_AUTHOR = "&copy; dCache.org ";
 
-    private static final ThreadLocal<SimpleDateFormat> _pathFromDate =
-            new ThreadLocal<SimpleDateFormat>()
+    private static final ThreadLocal<DateTimeFormatter> _pathFromDate =
```
kofemann, line 151: as DateTimeFormatter is thread-safe you don't need to to ThreadLocal anymore.

kofemann (changes requested):
The benefit of new time API is thread-safety. You should try to make use of it.

modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
iNeele: All ThreadLocal-variables have been removed and changed

modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
```diff
@@ -345,19 +284,19 @@ public void getInfo(PrintWriter pw) {
     public void messageArrived(CellMessage message) {}
 
     private String getNiceDayOfCalendar(Calendar calendar) {
-        return _dayOfCalendar.get().format(calendar.getTime());
+        return _dayOfCalendar.format(calendar.getTime().toInstant());
```
kofemann, line 287: The tima API always requires the timezone. Thus you should use something like:
`Calendar.getInstance().toInstant().atZone(ZoneId.systemDefault()`

This is easy to test with `jshell` (a java on command line):

```
[nairi] ~ $ jshell
|  Welcome to JShell -- Version 11.0.12
|  For an introduction type: /help intro

jshell> import java.time.*

jshell> import java.time.format.*

jshell> var formatter = DateTimeFormatter.ofPattern("yyyy" + File.separator + "MM" + File.separator + "dd" + File.separator + "yyyy-MM-dd-HH'.raw'");
formatter ==> Value(YearOfEra,4,19,EXCEEDS_PAD)'/'Value(MonthOf ... -'Value(HourOfDay,2)'.raw'

jshell> formatter.format(Calendar.getInstance().toInstant());
|  Exception java.time.temporal.UnsupportedTemporalTypeException: Unsupported field: YearOfEra
|        at Instant.getLong (Instant.java:602)
|        at DateTimePrintContext.getValue (DateTimePrintContext.java:308)
|        at DateTimeFormatterBuilder$NumberPrinterParser.format (DateTimeFormatterBuilder.java:2702)
|        at DateTimeFormatterBuilder$CompositePrinterParser.format (DateTimeFormatterBuilder.java:2341)
|        at DateTimeFormatter.formatTo (DateTimeFormatter.java:1843)
|        at DateTimeFormatter.format (DateTimeFormatter.java:1817)
|        at (#4:1)

jshell> formatter.format(Calendar.getInstance().toInstant().atZone(ZoneId.systemDefault()));
$5 ==> "2021/09/02/2021-09-02-10.raw"

jshell> 
```

kofemann (changes requested):
The java date and time related API is not at all friendly ....
 Have a look here

https://www.oracle.com/technical-resources/articles/java/jf14-date-time.html

modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
iNeele: Yes, you're right. I've already noticed this problem in IntelliJ. I fixed it by implementing a method, which converts the calendar datatype to LocalDateTime. LocalDateTime can be used by the formatter:

    private LocalDateTime toLocalDateTime (Calendar calendar) {
        return calendar.getTime().toInstant().atOffset(ZoneOffset.UTC).toLocalDateTime();
    }

    private String getNiceDayOfCalendar(Calendar calendar) {
        return _dayOfCalendar.format(toLocalDateTime(calendar));
    }

But your solution is probably the better one, isn't it?

lemora: You should probably use the system default zone instead of UTC, at least.

modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
```diff
@@ -241,13 +183,14 @@ protected SimpleDateFormat initialValue()
 
     private final CellNucleus      _nucleus;
 
-    private CellCron.TimerTask _hourly;
+    private final ScheduledExecutorService executor = Executors.newScheduledThreadPool(1);
+    private final Runnable hourlyRunner = new HourlyRunner();
```
lemora, line 187: Since you only use the hourlyRunner object in a single place, there is no need to store it globally.

modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
```diff
@@ -241,13 +183,14 @@ protected SimpleDateFormat initialValue()
...
+    private Calendar _calendar;
```
lemora, line 188: Better name the Calendar object "lastRun" or "previousRun" to make clear what it is used for. The current name does not convey anything.

modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
```diff
@@ -418,22 +360,25 @@ public void createDiffFile(File today, File yesterday, File resultFile) throws I
...
-            _calendar = calendar;
-            _nucleus.newThread(this,"FreeRunner").start();
+        private void updateCalendar() {
+            _calendar = new GregorianCalendar();
+            _calendar.set(Calendar.SECOND, 0);
         }
 
         @Override
         public void run() {
+            updateCalendar();
```
kofemann, line 370: 1. Why we need to update the calendar on each run?
2. if we need it, the update happens in a different thread, thus access must be thread-safe

kofemann (changes requested):
well, probably this is due to historical code, however the situation with the `Calendar` is not clear to me.

modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
iNeele: I don't know if I got that right, but I think this run-method is called once an hour (by the executorService).
Then files are created with the current date and time in their name, so that the time has to be updated.
Or did i misunderstand your point?

kofemann: Ok. looks like it was required to print when the a time the task will run. You already use `scheduleAtFixedRate` which returns a `ScheduledFuture`. This is you store it, then you can query it:

```
scheduledTask =  executor.scheduleAtFixedRate(...);

...
long delayInMillis = scheduledTask.getDelay();
....
```

kofemann (changes requested):
The formatter changes looks good, but scheduler needs some changes.

modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
```diff
@@ -338,26 +269,34 @@ public void getInfo(PrintWriter pw) {
         pw.println("   Stat Base : "+_dbBase);
         pw.println("   Html Base : "+_htmlBase);
         pw.println("      Images : "+ _images);
-        pw.println("    Next Run : "+_hourly);
+        pw.println("    Next Run : "+ new Date(Calendar.getInstance().getTime().getTime() +
```
kofemann, line 272: I think `Instant.now().plusMillis()' does the same, isn't it? 

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/Instant.html#plusMillis(long)

iNeele: yes, it does! But the output does not look the same in the println:
If you use Instant.now().plusMillis(...) it looks like:
2021-09-10T14:14:55.922033900Z
If you use new Date(...) it shows:
Fri Sep 10 16:14:56 CEST 2021

I don't know if this is important, but the "old" cellCron-scheduler printed the second one...

kofemann: This is only the matter of formatter that is used:
```
jshell> DateTimeFormatter.ofPattern("E L HH:mm:ss z yyyy").format(ZonedDateTime.now())
$10 ==> "Fri 9 16:29:34 CEST 2021"
```

iNeele: ohhhh.... okay :D
I will change it.

kofemann: javadoc is your friend :)

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/format/DateTimeFormatter.html

modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
```diff
@@ -338,26 +270,33 @@ public void getInfo(PrintWriter pw) {
         pw.println("   Stat Base : "+_dbBase);
         pw.println("   Html Base : "+_htmlBase);
         pw.println("      Images : "+ _images);
-        pw.println("    Next Run : "+_hourly);
+        pw.println("    Next Run : "+ Instant.now().plusSeconds(scheduledTask.getDelay(TimeUnit.SECONDS)));
```
kofemann, line 273: As `scheduledTask` get assigned in `started` method,  calling `getInfo` before scheduledTask is initialized will result in NullPointerException. Thus you have to check that `scheduledTask` is not null. Moreover, the `scheduledTask` should be `volatile` as it initialiaed and used from different threads.

kofemann (changes requested):
With minor fix the patch is ready to be merged. :tada:

modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
iNeele: thanks for your feedback :) Do you think this simple if statement is already enough?

kofemann: yes, or something like:

```
 pw.println("    Next Run : " + (scheduledTask == null ? "-" : ZonedDateTime.now()...toString());
```

paulmillar: [non-blocking]

Please bear in mind that the output here is for admins (i.e., human consumption).  An ISO 8601 representation is certainly valid, but not necessarily easy to digest.  You could use `org.dcache.util.TimeUtils#relativeTimestamp` to provide something easier for the admin to understand.

kofemann (approved):

modules/dcache/src/main/java/diskCacheV111/services/PoolStatisticsV0.java
iNeele: pw.println("    Next Run : " +
                (scheduledTask == null ? "-" :
                TimeUtils.relativeTimestamp(Instant.now().plusSeconds(scheduledTask.getDelay(TimeUnit.SECONDS)))));

Thanks for your feedback, too, Paul. I have now taken both of your comments into account. The output should now be more readable. However the line of code is very long... Should I commit again or change something else?

lemora (approved):

kofemann:
Thanks, @iNeele ! :tada:

PR #6045, opened 2021-09-01
 webdav: add 'http-tpc ls' admin command

paulmillar:
NB. This patch requires an additional (earlier) patch, which is included in this pull request.

webdav: add 'http-tpc ls' admin command

Motivation:

Provide admins with some idea what's happening.

Modification:

Add the 'http-tpc ls' admin command. This provides a table of summary
information about transfers that have been submitted and are on-going.
The command includes options for selecting which fields to show,
filtering transfers on different criteria, and sorting the resulting
table in different ways.

The progress reports messages from TransferManager are updated to
optionally include the pool name, if the request has been assigned to a
pool and transfer-manager has been updated. The command will use the
information, if available.

The IoJobInfo class is updated to optionally include the expected
transfers size. This information is optional because non-upgraded pools
will not include it and, under some circumstances, the information is
unknowable (e.g., downloading when HTTP chunked encoding is used).

Result:

The webdav admin interface has been extended to include monitoring
information about on-going transfers. The command can optionally show
pool information if transfer-manager is update; updating
transfer-manager is not required by this change. Percent transfer
progress for PULL requests is optionally available if pools are updated;
updating pools is not required by this change.

Target: master
Requires-notes: yes
Requires-book: no
Request: 7.1
Request: 7.0
Request: 6.2
Patch: https://rb.dcache.org/r/13150/
Acked-by: Albert Rossi

### DIFF
--- a/modules/common/src/main/java/org/dcache/util/TimeUtils.java
+++ b/modules/common/src/main/java/org/dcache/util/TimeUtils.java
@@ -377,6 +377,24 @@ public class TimeUtils
      * {@code <number> <space> <units>}, where {@code <number>}
      * is an integer and {@code <units>} is defined by the value of unitFormat.
      */
+    public static StringBuilder appendDuration(StringBuilder sb, Duration duration,
+                    TimeUnitFormat unitFormat)
+    {
+        // FIXME: this method is a wrapper around the overloaded method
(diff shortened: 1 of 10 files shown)


paulmillar:
retest this please

PR #6083, opened 2021-09-29
dcache: Unit tests of messageReply

iNeele:
Motivation:
The class messageReply is used in many services and allows a reply
to be send securely from the message delivery thread.
However, many of the cell communication tests are still missing,
as are those for messageReply.

Modification:
Adds one test file, in which all methods of the messageReply class
are tested.

Result:
The tests increase code coverage.
No user observable changes.

Target: master

### DIFF
--- a/modules/dcache/src/test/java/org/dcache/cells/MessageReplyTest.java
+++ b/modules/dcache/src/test/java/org/dcache/cells/MessageReplyTest.java
@@ -1,0 +1,348 @@
+package org.dcache.cells;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+import diskCacheV111.util.CacheException;
+import diskCacheV111.vehicles.Message;
+import dmg.cells.nucleus.CellAddressCore;
+import dmg.cells.nucleus.CellEndpoint;
+import dmg.cells.nucleus.CellMessage;
+import dmg.cells.nucleus.UOID;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import org.junit.Test;
+
+
+public class MessageReplyTest {
+
+    @Test(expected = NullPointerException.class)
+    public void shouldThrowAnExceptionIfDeliveringWithoutEndpointAndEnvelope() {
+        MessageReply messageReply = new MessageReply();
+        messageReply.deliver(null, null);
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void shouldThrowAnExceptionIfDeliveringWithoutEndpoint() {
+        MessageReply messageReply = new MessageReply();
+        CellMessage envelope = new CellMessage();
+
+        messageReply.deliver(null, envelope);
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void shouldThrowAnExceptionIfDeliveringWithoutEnvelope() {
+        MessageReply messageReply = new MessageReply();
+        CellEndpoint endpoint = mock(CellEndpoint.class);
+
+        messageReply.deliver(endpoint, null);
+    }
+
+    @Test
+    public void shouldNotSendIfDeliveringWithoutMessage() {
+        MessageReply messageReply = new MessageReply();
+        CellMessage envelope = new CellMessage();
+        CellEndpoint endpoint = mock(CellEndpoint.class);
+
+        messageReply.deliver(endpoint, envelope);
+        verify(endpoint, times(0)).sendMessage(envelope);
+    }
+
+    @Test
+    public void shouldSendIfDeliveringWithAMessage() {
+        MessageReply<Message> messageReply = new MessageReply<>();
+        Message message = new Message();
+        CellMessage envelope = new CellMessage();
+        CellAddressCore pathSource = new CellAddressCore("foo", "source");
+        CellEndpoint endpoint = mock(CellEndpoint.class);
+
+        envelope.addSourceAddress(pathSource);
+
+        messageReply.reply(message);
+        messageReply.deliver(endpoint, envelope);
+        verify(endpoint, times(1)).sendMessage(envelope);
+    }
+
+    @Test
+    public void shouldBeValidIfEnvelopeIsNull() {
+        MessageReply messageReply = new MessageReply();
+        assertTrue(messageReply.isValidIn(0));
+    }
+
+    @Test
+    public void shouldBeValidIfDelayIsSmallerThanTimeToLiveOfEnvelope() {
+        MessageReply messageReply = new MessageReply();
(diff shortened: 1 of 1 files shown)


iNeele:
I have a few questions :)

1.) First, I wonder if I should use "random" or if I should pass fixed values.
(the reason why I used random numbers is that the tested methods has to work with all positive numbers, not just one...).

2.) I'm also a little bit unsure of the time the tests need to run.
For example to simulate that a thread is waiting I let the programme sleep for 100 milliseconds several times.
Moreover, in isValidIn I increased the timeToLive by a second... is it enough to run the test in every scenario correctly?

3.) And last, I probably need to edit the layout including the name of all tests, right?

thanks for your feedback and further comments!

paulmillar:
I don't think anyone has all the answers for unit-tests, but here's my take.

1. In generally, using anything random in a unit-test is a bad idea.

A key point of unit tests is repeatability: a test that fails should do so repeatedly.  How else do you know if a patch fixes the problem?  (and, in my experience, intermittent failures are always a complete pain to diagnose and fix.) 

You also mentioned testing that it works for all positive numbers.  There are ways of doing this (JUnit Theories spring to mind), but perhaps a "better" goal (for some definition of better) is just to aim for 100% code-branch coverage (`-P.code-coverage` is your friend)  This will reduces the risk of adding pointless unit-tests, while will still making it quite likely to catch any regression.

2. Time is a difficult concept in unit-tests (as are any environmental factors).  In general, sleeping in a test is a bad idea.

It makes tests fragile because the exact timing depends on the CPU, competing other activity, etc.  In general, a better approach is to use synchronisation between threads (you may need to be inventive).  If you want to test how something behaves with the passing of time, Java v8 introduced the Clock class that may be pressed into service.  For inspiration, see `AdvanceableClock` and `ParsableFileTest`.

3. In terms of layout, note that Java methods don't use `snake_case_name`, but `camelCaseNames`.

You might like to look at BDD (Behaviour Driven Development) for a guide on how to lay out a unit test; in particular, the "given-when-then" concept.  Unfortunately, we have to use `assert`, rather than "then", both otherwise the framework goes something like:

    void shouldDoX() throws Exception {
        // set up initial conditions.
        givenX();
        givenY();

        whenA();  // something happens

        assertThat(xxx, ...); // verify the consequences
        assertThat(yyy, ...);
    } 


For what I consider half-way reasonable examples,  take a look at `ParsableFileTest` and `RequestContainerV5Test` (if you're feeling brave!)

Some general comments -- my own opinion (so nothing that's set in stone) but perhaps a good starting point.

a. if you're testing class X then create a new (real) class for each test and use mock objects for all its dependencies.  This gives you maximum control over setting up the initial conditions.

b. try to put all the information describing the scenario being tested in the unit-test method itself.  I tend to use builder pattern to do this (I think it makes the code easier to read), but that isn't essential.

c. think very carefully before using Mockito spy!  In general, don't use it (spy is mostly for working with third-party code and/or legacy code). Instead, think what is the "test surface" you wish to test (usually the public methods) and focus more on the given-when-then way of structuring test: set up the initial conditions, trigger something happening, check it has the right consequences.

iNeele:
Hey Paul, thanks for your answer.

I had already thought to myself that random numbers are not a good idea (and Lea had also pointed this out to me). As soon as the test fails, the reason for the failure cannot be tracked, that makes sense to me. I should also revise the test-names, thank you for the tip!
Though I am little bit stuck with your second mentioned point to synchronize the threads and to use mock instead of spy. I think the main problem is that there are three private variables in messageReply (the class that I want to test) and I cannot mock one of them because it is a final class (CellMessage). In addition, there are no setters or getters in messageReply.

To create individual and independent tests that should only check one method at a time, I used a spy, because that allows me to check whether a single method is being invoked by another method. For example, in the method deliver, which calls the method send if message is not null:
        if (_msg != null) {
            send();
        }

Here I didn’t want to test whether everything works like I would invoke send, but just if send is generally invoked if the message is not null. So, I thought I would avoid code-duplication (because I’m testing the method send elsewhere) by using: verify(messageReplySpy, times(1)).send()).

I found out that I could also mock final classes, but to do so I have to add a new test file to the resources folder (https://www.baeldung.com/mockito-final). Do you think this would be a possible alternative to replace the spy with a mock?

What is more, I have a similar problem with wait():
while (_msg == null) {
    wait();
}

I can test the state of a thread via:
assertEquals(Thread.State.WAITING, thread.getState());
But the getState-request must be in a different thread. However, I understood it that way, that threads run parallel. So I get an error, because the thread that calls wait() has not yet called it and have the state RUNNABLE. That’s why the second thread is only allowed to make the getState-request AFTER the first thread has invoked wait(). I don’t know exactly how to synchronize that... I hope that was to some degree understandable 😊

modules/dcache/src/test/java/org/dcache/cells/MessageReplyTest.java
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    @Test
+    public void shouldFailWithCacheException() {
+        MessageReply<Message> messageReply = new MessageReply<>();
+        Message message                    = new Message();
+        Throwable e                        = new CacheException("ExceptionMessage");
+
+        MessageReply<Message> messageReplySpy = spy(messageReply);
+        doNothing().when(messageReplySpy).fail(any(Message.class), any(int.class), any(Serializable.class));
```
kofemann, line 135: possibly `anyInt()` is better than `any(int.class)`

modules/dcache/src/test/java/org/dcache/cells/MessageReplyTest.java
```diff
@@ -0,0 +1,422 @@
...
+            } catch (InterruptedException | ExecutionException e) {
+                e.printStackTrace();
+            }
+        });
+        thread.start();
+
+        TimeUnit.MILLISECONDS.sleep(100);
+        assertEquals(Thread.State.WAITING, thread.getState());
+
+        Field messageField = messageReply.getClass().getDeclaredField("_msg");
```
kofemann, line 332: Why you  don't simply `reply(message)`? The test checks that external notiffyAll will wake up the thread, but is it happens in the code?

modules/dcache/src/test/java/org/dcache/cells/MessageReplyTest.java
```diff
@@ -0,0 +1,422 @@
...
+            }
+        });
+        thread.start();
+
+        TimeUnit.MILLISECONDS.sleep(100);
+        assertEquals(Thread.State.TIMED_WAITING, thread.getState());
+
+        Field messageField = messageReply.getClass().getDeclaredField("_msg");
+        messageField.setAccessible(true);
+        messageField.set(messageReply, message);
```
kofemann, line 389: same here

kofemann (changes requested):
Nice work! However, some usage of reflection looks to me over kill. Do you really need it?

modules/dcache/src/test/java/org/dcache/cells/MessageReplyTest.java
iNeele: Yes, I agree with you and have changed the tests accordingly.

iNeele:
I have removed all reflections except one (on this one I will have a closer look again).
In addition, I was able to remove some of the spy-objects.
Do you know, how I can handle the rest of them?

kofemann:
@iNeele Some of the some of changes probably not too hard to fix. However, I suggest we commit merge as-is and leave improvements for the next generation :)

PR #6095, opened 2021-10-06
acl: set IDENTIFIER_GROUP flag if WHO is GROUP or OWNER_GROUP

kofemann:
Motivation:
The dCache (chimera) t_acl table have two possibilities to specify the
principal as a group identifier - through `who` field, with a defined
value 1 (one), or as a bit in ACE corresponding flag. The flag is used
NFSv4 server, as required by spec. Nevertheless, when dCache admin
command (or chimera-cli) sets group ACE, the flag is not set, thus
such ACEs apply to users principals instead of groups.

Modification:
Update ACE class constructor to enforce IDENTIFIER_GROUP flag if WHO is GROUP or OWNER_GROUP.

Result:
The group ACEs apply to desired group instead of to a users with the
same numeric id.

Ticket: #10191
Acked-by:
Target: master, 7.2, 7.1, 7.0, 6.2
Require-book: no
Require-notes: yes
(cherry picked from commit 04d8ae96ebdeed85add4c72d18109f711c304b70)

### DIFF
--- a/modules/acl-vehicles/src/main/java/org/dcache/acl/ACE.java
+++ b/modules/acl-vehicles/src/main/java/org/dcache/acl/ACE.java
@@ -60,7 +60,7 @@ public class ACE implements Serializable
      */
     public ACE(AceType type, int flags, int accessMsk, Who who, int whoID) {
         _type = type;
-        _flags = flags;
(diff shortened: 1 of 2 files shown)


svemeyer:
retest this please

mksahakyan:
retest this please

PR #6100, opened 2021-10-11
cells: Unit tests of cellMessage

iNeele:
Motivation:
Some of the methods of the cellMessage class has not yet been tested.

Modification:
Adds some tests to the existing test file for the class cellMessage.

Result:
The tests increase code coverage.
No user observable changes.

Target: master


lemora:
@kofemann If you can live with `assertFalse` rather than `assertNotEquals`... although I guess we would need to change this after committing in any case.
Sure, I'll merge the patch then?

PR #6107, opened 2021-10-13
Update install.md

ufalke:
More clarity on systemd units (the multi-unit example confuses the first-time installer who should expect exactly one systemd unit ...)


ufalke:
K,  changed 7.2 to master in #6107, but now there are 22 commits there of which mine is only one  (324d79b<https://github.com/dCache/dcache/pull/6107/commits/324d79bfde9d49f68b63929eec75d930f95e8e23>).

I do not want to break things ...

________________________________

ufalke:
changing back to 7.2 and going to open a similar PR for master. thx for your advice ...

PR #6225, opened 2021-11-08
pnfsmanager: fix confusing error message.

paulmillar:
Motivation:

When creating a file, certain failure modes trigger misleading or
otherwise confusing error messages.  Here are the two specific examples.

First, when attempting to upload `/A/my-file` and `/A` is a file, the
error message is:

    Not a directory: /A/my-file

This is misleading because the problem is with `/A`, not with
`/A/my-file`.

Second, when attempting to upload `/A/my-file` and directory `/A` is
missing, the error message is:

    No such file or directory: /A/my-file

This is misleading on two counts: it is `/A` that is missing and `/A`
should be a directory and not a file.

Modification:

Update these error messages to use the parent path.  In the second case,
"file or directory" is replaced by "directory".

Result:

Attempts to create a file where the parent directory is either missing
or not a directory now provide clearer error messages.

Target: master
Requires-notes: yes
Requires-book: no
Request: 7.2
Request: 7.1
Request: 7.0
Request: 6.2
Patch: https://rb.dcache.org/r/13243/
Acked-by: Lea Morschel
Acked-by: Svenja Meyer


paulmillar:
retest this please

Issue #6194, opened 2021-10-28
examples for needrestart


calestyo:
Hey.

Do we have something like Debian's `/usr/share/doc/dcache/` in dcache’s git repo?

In Debian it's typical, that one may have an `examples` subdir in that, and for that I'd propose to add:

`needrestart/conf.d/exclude-dcache.conf`:
`$nrconf{override_rc}{qr(^dcache@.+\.service)} = 0;`

`needrestart/conf.d/exclude-postgresql.conf`:
`$nrconf{override_rc}{qr(^postgresql@.+\.service)} = 0;`

These can be used to cause needrestart ignore dcache/postgresql per default and might therefore be beneficial to anyone who runs dcache with needrestart.
I don't think however, that these files should be automatically installed into `/etc`

Neither do I think that such example should go to the book, because no one will ever find them there ;-)

Cheers,
Chris.

kofemann:
Hi @calestyo ,

A funny tool. I never heard about it. Sounds like a workaround for broken packages. Anyway, if you think it would be a good idea to include examples with the dcache package, we can do that. The /usr/share/doc/dcache sounds good.

PR #6242, opened 2021-11-10
skel: added example configuration for needrestart

calestyo:
Motivation:

needrestart is “semi-default” on Debian systems (and may be used with any other
distributions as well.
It allows to detect which services need to be restarted after upgrades and, per
default, automatically selects them for being so.

For dCache and it’s dependencies like PostgreSQL, this is typically undesired on
production clusters.

Modification:

Added example configuration and documentation on how to specifically prevent
that behaviour for dCache and PostgreSQL.

Target: master
Requires-notes: no
Requires-book: no
Issue: #6194


paulmillar:
ok to test

PR #6250, opened 2021-11-12
Update cookbook-transport-security.md

onnozweers:
Updated some things like redirection of WebDAV, Java versions and their security settings, and the security-policies framework which is new and a bit annoying in EL8.


paulmillar:
OK to test

kofemann:
Thanks Onno!

Issue #6251, opened 2021-11-12
Pin request TTL exceeded [10006]


qiulan2021:
We got Pin request TTL exceeded issue.

[...]

PR #6266, opened 2021-11-17
pinmanager: provide more details information on timeout

paulmillar:
Motivation:

Issue #6251 describes a problem with pinmanager where requests are
caught in a retry loop until the pin request message timeout results in
the request being terminated.

Unfortunately, the current error message provides insufficient
information to learn what is going wrong.  It indicates that the request
has been retried until it eventually fails, but not why.

Modification:

In lieu of a full pin-request activity log, we can throw a
CacheException with a message that includes why the final retry was
attempted, the one that resulted in the request exceeding the message
lifetime.

This is in keeping with how explicit failures are handled, where more
detailed information is provided.

Result:

A more detailed pinmanager log entry is provided for pin requests that
are caught in a retry loop, and eventually time out.

Target: master
Request: 7.2
Request: 7.1
Request: 7.0
Request: 6.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/13266/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/pinmanager/PinRequestProcessor.java
+++ b/modules/dcache/src/main/java/org/dcache/pinmanager/PinRequestProcessor.java
@@ -209,17 +209,18 @@ public class PinRequestProcessor
         return RequestContainerV5.allStatesExceptStage;
     }
 
-    private void retry(final PinTask task, long delay) {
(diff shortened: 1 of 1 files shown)


paulmillar:
retest this please

Issue #5958, opened 2021-07-05
Update `dcache-star` script to use Python v3


paulmillar:
The `dcache-star` Python script is currently written for Python v2.  As per #5957, this should be updated to use Python v3

PR #6334, opened 2021-12-06
dcache-star: conversion to Python3

calestyo:
Motivation:

Python2 has reached end-of-life and is no longer supported.

Modification:

Converted `dcache-star` to Python3

Target: master
Requires-notes: no
Requires-book: no
Issue: #5958

### DIFF
--- a/packages/fhs/src/main/deb/control
+++ b/packages/fhs/src/main/deb/control
@@ -9,7 +9,7 @@ Package: @PackageName@
 Depends: ssh-client, adduser, ssl-cert, rsyslog
 Conflicts: dcache-server
 Architecture: all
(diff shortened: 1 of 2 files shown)


calestyo:
Someone should really double check this, especially the line which sets the filename:
```
filename = '%08x'%int(now) + '%05x'%int((now - int(now))*1000000) + '%x'%randint(0,15)
```

I wasn't really sure what it's intended to be... there there is some guesswork with the conversions ;-)

calestyo:
Maybe this should actually go to the release notes, cause it requires people to have Python3 rather than.

paulmillar:
Thanks, Chris.

Looks good.

PR #6377, opened 2021-12-15
common-security: prevent NPE on password protected cert

lemora:
Motivation:

When attempting to use a password-protected certificate with GridFTP, a NullPointerException is logged without any further information. This is a bug and does not help discover the underlying problem.

Modification:

The CanlContextFactory attempts to get the credential by creating a new instance of PEMCredential, which expects a key password as a char array. Because it is assumed that the key is not password protected, a value of null is passed, leading to a NullPointerException later on. This is changed, for every code occurrence, to instead pass an empty char array, which will result in a more helpful error message.

Result:
A better error message is logged when attempting to use a password-protected credential:
java.io.IOException: Error decrypting private key: the password is incorrect or the PEM data is corrupted.

Target: master
Target: 7.2
Target: 7.1
Target: 7.0
Target: 6.2
Ticket: #10265
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/13337/
Acked-by: Paul Millar

### DIFF
--- a/modules/common-security/src/main/java/org/dcache/ssl/CanlContextFactory.java
+++ b/modules/common-security/src/main/java/org/dcache/ssl/CanlContextFactory.java
@@ -269,9 +269,14 @@ public class CanlContextFactory implements SslContextFactory {
 
         public Callable<SSLContext> buildWithCaching() throws IOException {
             final CanlContextFactory factory = build();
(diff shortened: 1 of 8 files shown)


svemeyer:
retest this please

PR #6384, opened 2022-01-03
gplazma: scitoken validate 'wlcg.ver' claim

paulmillar:
Motivation:

The WLCG AuthZ JWT profile defines the 'wlcg.ver' claim that all tokens
must provide.  It provides a smooth upgrade path; for example, by
allowing services to support multiple profile versions.

Although the 'scitoken' plugin doesn't require 'wlcg.ver' as it also
supports SciToken tokens, it should check the value if one is present.

Modification:

Add code to validate the 'wlcg.ver' claim, if present.

Add unit tests to verify correct behaviour.

Result:

dCache will now reject WLCG-AuthZ-JWT profile tokens with a 'wlcg.ver'
claim that it does not support.  Other tokens (e.g., SciTokens) are
unaffected by this change.

Target: master
Requires-notes: yes
Requires-book: no
Request: 7.2
Request: 7.1
Request: 7.0
Request: 6.2
Patch: https://rb.dcache.org/r/13343/
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/gplazma2-scitoken/src/main/java/org/dcache/gplazma/scitoken/SciTokenPlugin.java
+++ b/modules/gplazma2-scitoken/src/main/java/org/dcache/gplazma/scitoken/SciTokenPlugin.java
@@ -29,6 +29,7 @@ import java.security.Principal;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


paulmillar:
retest this please

PR #6428, opened 2022-01-26
dcache-core: fix admin command to query pins by state

lemora:
Motivation:

When querying pins by state, `count pins unpin-failed` also reports the number of pins in state PINNING, which is not intended.

Modification:
Result:

When asking for `count pins unpin-failed`, only report the number of pins in state FAILED_TO_UNPIN and not PINNING.

Target: master
Target: 7.2
Target: 7.1
Requires-notes: no
Requires-book: no
Patch: https://rb.dcache.org/r/13395/
Acked-by: Paul Millar


kofemann:
retest this please

mksahakyan:
retest this please

mksahakyan:
retest this please

PR #6430, opened 2022-01-26
skel: minor improvements to example configuration for needrestart

calestyo:
Modification:

Properly anchor the regular expression and escape the `@` which Perl may
interpret as variable when followed by a valid identifier.

Target: master
Requires-notes: no
Requires-book: no


kofemann:
Thanks!

PR #6436, opened 2022-01-28
TransferManager: log bugs with corresponding stack-trace

paulmillar:
Motivation:

There are a few places where TransferManager catches a RuntimeException
but logs this exception without any stack-trace.  A dCache admin might
not realise a log entry without a stack-trace is the result of a bug,
and so not report it.  Similarly, without a stack-trace, understanding
the bug becomes considerably harder.

Modification:

Update log entries to include a meaningful message and to log the
stack-trace.

Result:

TransferManager will now log bugs with the corresponding stack-trace,
making fixes any such bug easier.

Target: master
Requires-notes: yes
Requires-book: no
Request: 7.2
Request: 7.1
Request: 7.0
Request: 6.2
Patch: https://rb.dcache.org/r/13373/
Acked-by: Lea Morschel


mksahakyan:
retest this please

PR #6439, opened 2022-01-28
CopyManager: reply success when the target transfer completes.

paulmillar:
Motivation:

The CopyManager initiates a transfer between pools to duplicate the
contents of a file.  It does this by creating two HTTP transfers: one on
the source pool and one on the destination pool.

Currently, it waits until both transfers have completed successfully
before indicating that the copy request was processed successfully.

This is unnecessary, as the successful completion of the download
transfer (on the destination pool) is sufficient to know that the
transfer was successful; e.g., a failure on the source pool during
post-processing does not effect the success of the copy request.

Moreover, for optimisation reasons, the source pool's transfer might not
complete immediately.  Therefore, delaying the successful response until
both transfers have completed could introduce a significant delay in the
(apparent) time needed to copy a file.

Modification:

Rework the code that sends the reply to the door, indicating whether the
copy request was processed successfully.

Update the 'copy' method, so that the successful cell message response
is sent as soon as the upload transfer (on the destination pool)
completes successfully.  The method still waits for the download
transfer (on the source pool) to complete and logs any problem, but now
such problems no longer influences the success of the copy request.

Result:

dCache internal file copying is slightly faster, under some
circumstances.

Target: master
Request: 7.2
Requires-notes: no
Requires-book: no
Patch: https://rb.dcache.org/r/13360/
Acked-by: Lea Morschel

### DIFF
--- a/modules/dcache/src/main/java/diskCacheV111/doors/CopyManager.java
+++ b/modules/dcache/src/main/java/diskCacheV111/doors/CopyManager.java
@@ -38,6 +38,7 @@ import org.dcache.cells.CellStub;
 import org.dcache.poolmanager.PoolManagerStub;
 import org.dcache.util.Args;
 import org.dcache.util.RedirectedTransfer;
+import org.dcache.util.TimeUtils;
 import org.dcache.util.Transfer;
 import org.dcache.util.URIs;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


mksahakyan:
retest this please

mksahakyan:
retest this please

Issue #6449, opened 2022-02-01
frontend:  bogus request hash returns 403, not 404


alrossi:
A non-existent bulk request that is queried should provoke a "Not Found" error code.

7357d257-0c0d-4691-aaaa-e8860fb7e4c1 does not exist:

```
[litvinse@fnisd1 ~]$  curl  -L --capath /etc/grid-security/certificates --cert /tmp/x509up_u8637 --cacert /tmp/x509up_u8637 --key  /tmp/x509up_u8637  https://fndca3b.fnal.gov:3880/api/v1/bulk-requests/7357d257-0c0d-4691-aaaa-e8860fb7e4c1
{"errors":[{"message":"Forbidden","status":"403"}]}[litvinse@fnisd1 ~]$ 
```

alrossi:
https://rb.dcache.org/r/13430/

PR #6482, opened 2022-02-09
bulk, frontend-rest: return 404, not 403, when request id does not exist

alrossi:
Motivation:

See GH #6449 https://github.com/dCache/dcache/issues/6449

When GET or PATCH is issued for the bulk-request resource
(requesting the status or cancelling the request), if the
submitted id is not recognized by the service, 404 should
be returned.  Currently, we are seeing 403 (permission denied)
instead.

Modification:

The ordering of the underlying checks for existence and
then for permissions needs to be reversed.

Result:

The existence check overrides the permission check in
terms of the error return, as it logically should.

Target: master
Request: 8.0
Request: 7.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/13430/
Closes: #6449
Acked-by: Tigran


svemeyer:
retest this please

PR #6488, opened 2022-02-15
pool: try to forcefully kill HSM process if needed

kofemann:
Motivation:
When nearline interface needs to kill HSM script then `Process#destroy`
is called to interrupt the running HSM migration process. However, if
the migration script catches SIGTERM (15) and ignores it, then request
to HSM will be blocked in CANCELED state in the pool.

Modification:
If HSM script survives `Process#destroy` switch to
`Process#destroyForcibly` which sends SIGKILL (9).

Result:
less situations when expired/canceled requests stay in CANCELED.

Acked-by: Paul Millar
Target: master, 8.0, 7.2
Require-book: no
Require-notes: yes
(cherry picked from commit 982f89b0ffb5e1b6c4920c131e86fefe5d379e58)


svemeyer:
retest this please

PR #6490, opened 2022-02-15
dcache-bulk: pin by request id and unpin optionally by request

alrossi:
Motivation:

Currently PIN/UNPIN in the bulk service is implemented
so as to pin without a request id.   This is actually
an implementation flaw.

What should be happening is that the pin should be (by default)
made on behalf of the specific request, and UNPIN should
then (usually) be issued with respect to the original
PIN request.

Modification:

Add the id arguments to both PIN and UNPIN, with
appropriate descriptors.

If the argument is absent from the PIN request,
it defaults to the request id of the PIN target.
If it is missing from UNPIN, it is left NULL.

Result:

Request-specific pinning is enabled, but generic
unpinning is still possible.

Target: master
Request: 8.0
Patch:  https://rb.dcache.org/r/13433/
Requires-notes: yes
Requires-book: no
Acked-by: Lea


svemeyer:
retest this please

svemeyer:
retest this please

PR #6491, opened 2022-02-15
dcache-bulk: aborted request gets stuck in the STARTED state

alrossi:
Motivation:

On FNAL production we have found that file permission or existence errors on
paths/targets at the root of the request leave the request stuck in the
STARTED state.

Modification:

The initial bulk request job needs to check for request completion after
unregistering itself from the completion handler.

Result:

Jobs which originally get stuck now complete; their failure information
contains the reason for premature completion.

(NOTE:  the 'null' DEPTH seems to appear on FNAL production [7.2], but
but I have not yet been able to reproduce it using 7.2 on the testbed.)

Target: master
Request: 8.0
Request: 7.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/13442/
Acked-by: Tigran


svemeyer:
retest this please

PR #6506, opened 2022-02-24
pool: http-tpc prefer IPv6 address

paulmillar:
Motivation:

WLCG would like HTTP-TPC transfers to use an IPv6 address if both the
pool and the remote site support IPv6.

Modification:

Out-of-the-box, Apache HTTP client will try connections in the order
returned by the system's DNS resolver.  This order is not guaranteed,
but seems to favour IPv4 in real-world environments.

Add a custom DnsResolver class that prefers IPv6 addresses.

Result:

HTTP-TPC transfers now prefer IPv6 address, if both endpoints support
it.

Target: master
Requires-notes: yes
Requires-book: no
Request: 8.0
Request: 7.2
Request: 7.1
Request: 7.0
Request: 6.2
Patch: https://rb.dcache.org/r/13437/
Acked-by: Tigran Mkrtchyan
Acked-by: Dmitry Litvintsev


svemeyer:
retest this please

PR #6532, opened 2022-03-08
dcache-core: prevent ocasional message ttl test failure

lemora:
Motivation:
By default a CellMessage's TTL is Long.MAX_VALUE. The unit test code in methods `shouldBeValidIfDelayIsSmallerThanTimeToLiveOfEnvelope` and `shouldBeInvalidIfDelayIsBiggerThanTimeToLiveOfEnvelope` calls setTtl(Long.MAX_VALUE) and then isValidIn(Long.MAX_VALUE). This will fail if the code takes longer than (on average) 0.5 ms from calling the CellMessage constructor to the test.

As is, this test is nonsense.

Modification:
Fix the unit tests by setting the message test lifetime to a fixed, smaller than maximal value to ensure that the test setup and execution work as intended. Test against a larger difference in both the positive and negative case.

Result:
Fix unit tests and prevent accidental MessageReplyTest test failures due to high system load.

Target: master, 8.0
Requires-notes: no
Requires-book: no
Patch: https://rb.dcache.org/r/13457/
Acked-by: Paul Millar

### DIFF
--- a/modules/dcache/src/test/java/org/dcache/cells/MessageReplyTest.java
+++ b/modules/dcache/src/test/java/org/dcache/cells/MessageReplyTest.java
@@ -85,12 +85,11 @@ public class MessageReplyTest {
 
         messageReply.deliver(endpoint, envelope);
 
-        long timeToLive = envelope.getTtl();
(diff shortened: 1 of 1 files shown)


svemeyer:
retest this please

Issue #6181, opened 2021-10-26
Open srr api to external access


XMol:
Hello dCache.org,

as became obvious by the feedback on the user-forum thread "The new SRR module is GA", multiple sites would like you to lift the restriction on the [new SRR REST-API](https://github.com/dcache/dcache/commit/24d79fdd765c568ac4fc3ceabce502d997eb7379) such that it may be queried from other places besides localhost.

The benefits are quite obvious (to me):
* Consumers - most prominently WLCG's CRIC - may retrieve up-to-date information any time it wants to.
* Site admins are no longer obligated to set up a cronjob that downloads the report and then upload it to a location where consumers may read it from.

Thank you for all your effort that went into this service so far!

Best regards,
Xavier.

XMol:
Hello dCache.org,

are there other open issues regarding SRR that you have to work on in parallel? If not, may I ask you kindly to include the necessary changes in one of the upcoming releases soon? According to Tigran, actually very little work would go into it, so I'm surprised this request is pending for so long, despite the feedback on the user-forum.

Best regards,
Xavier.

paulmillar:
Hi Xavier,

The changes in behaviour you describe are available with dCache v7.2.6, v7.1.16, v7.0.25, v6.2.36.

These dCache versions were released on 2021-12-10 (last Friday).

Cheers,
Paul.

kofemann:
@paulmillar The direrd change is not in the dcache code base and still in RB: https://rb.dcache.org/r/13273/

XMol:
Thank you for the quick update!

XMol:
Hello @kofemann, @paulmillar,

again two months have passed and the issue in RB that Tigran linked has seen no updates either. I trust you did not forget about this request? 😉 

Ciao,  
Xavier.

paulmillar:
I believe this is done (or almost done).  You should see something "soon".

XMol:
Thank you @kofemann,

adding a new property to toggle the behavior is fine for me/us.

Ciao,  
Xavier.

ageorget:
Hi,
I'm trying to enable public access to SRR with the 6.2.41 dCache version but I still got 403 forbidden error if I ask anything else than localhost.
My Frontend configuration looks like : 

```
[FrontendDomain]
[FrontendDomain/frontend]
frontend.authn.basic=true
frontend.authn.protocol=https
frontend.authz.anonymous-operations=READONLY
frontend.srr.public=true
```
[...]

PR #6551, opened 2022-03-17
frontend: fix problem with making SRR resources public

lemora:
Motivation:
Commit 3351df28ab introduced an option that allows making the SRR resources public. The patch contained a mistake, which inverts the logic of the property, so that public access is allowed when `frontend.srr.public` option is set to false, not true.

Modification:
Invert the faulty logic so that now `frontend.srr.public=true` means "allow public access" and false does not.

Result:
The SSR public access property works as intended.

Target: master
Request: 8.0, 7.2, 7.1, 7.0, 6.2
Fixes: #6181
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/13488/
Acked-by: Tigran Mkrtchyan


svemeyer:
retest this please

PR #6573, opened 2022-03-31
pnfsmanager: avoid leaking whether or not file exists

paulmillar:
Motivation:

The WebDAV door (among other doors) queries pnfsmanager to discover the
PNFS-ID of a target. Within pnfsmanager, the NameSpaceProvider plugin is
used to discover information about the target.

When looking up the inode from a path, the Chimera plugin
(ChimeraNameSpaceProvider) checks whether the user can 'cd' into the
target's parent directory.  To do this, it looks up the inode
information for the target and all it's parents.

Currently, if the target does not exist then this immediately throws
FileNotFoundChimeraFsException.  This is true even if the user does not
have permission to 'cd' into a parent directory.  Therefore, the user
can discover whether a file exists within a directory that user cannot
enter.

Modification:

Update ChimeraNameSpaceProvider so that, after discovering the target
does not exist, it next checks whether the user can 'cd' into the
deepest parent directory that exists.  If the user does not have
permission to 'cd' into this deepest parent directory then the
FileNotFoundChimeraFsException is replaced with a
PermissionDeniedCacheException.

Result:

Doors no longer allow a user to discover whether or not a file or
directory exists within directories they cannot access.

Target: master
Request: 8.0
Request: 7.2
Request: 7.1
Request: 7.0
Request: 6.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/13505/
Acked-by: Marina Sahakyan
Acked-by: Tigran Mkrtchyan

### DIFF
--- a/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraNameSpaceProvider.java
+++ b/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/ChimeraNameSpaceProvider.java
@@ -227,23 +227,51 @@ public class ChimeraNameSpaceProvider
         _uploadSubDirectory = path;
     }
 
+    private void checkLookupPermissions(Subject subject, List<FsInode> inodes, String path)
+          throws ChimeraFsException, CacheException {
+        for (FsInode inode : inodes) {
+            if (inode.isDirectory()) {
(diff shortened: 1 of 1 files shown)


svemeyer:
retest this please

Issue #6557, opened 2022-03-21
transfers failing with rejected PUT: 500 Server Error


cfgamboa:
Hello all,

we have seen many transfers failing with ``rejected PUT: 500 Server Error``
https://fts308.usatlas.bnl.gov:8449/var/log/fts3/transfers/2022-03-17/webdav.aglt2.org__dcsrm.usatlas.bnl.gov/2022-03-17-0609__webdav.aglt2.org__dcsrm.usatlas.bnl.gov__1277490583__bc5672c6-a5b6-11ec-82c5-b49691292f84

At the billing log the file appears to be available for copy, here the  source is PNSFID 0000EEC369BA76464902A124C6842E7DE7CA, file name:

```
/pnfs/usatlas.bnl.gov/BNLT1D0/data16_13TeV/RAW/other/data16_13TeV.00300800.physics_Main.daq.RAW/data16_13TeV.00300800.physics_Main.daq.RAW._lb0419._SFO-2._0002.data
/pnfs/usatlas.bnl.gov/BNLT1D0/data16_13TeV/RAW/other/data16_13TeV.00300800.physics_Main.daq.RAW/data16_13TeV.00300800.physics_Main.daq.RAW._lb0419._SFO-2._0002.data
```

Looking at the Billinglogs for that transfer

1. Not working transfer
```
grep data16_13TeV.00300800.physics_Main.daq.RAW._lb0419._SFO-2._0002.data billing-2022.03.21

03.21 17:08:42 [door:WebDAV2-dcdoor12-2@webdav2-dcdoor12_httpsDomain:request] ["usatlas1":6435:31152:192.12.15.232] [0000EEC369BA76464902A124C6842E7DE7CA,2621829888] [/pnfs/usatlas.bnl.gov/BNLT1D0/data16_13TeV/RAW/other/data16_13TeV.00300800.physics_Main.daq.RAW/data16_13TeV.00300800.physics_Main.daq.RAW._lb0419._SFO-2._0002.data] bnlt1d0:BNLT1D0@osm 5 0 {10011:"Task was cancelled."}

[...]

PR #6575, opened 2022-03-31
pool: describe why migration job was cancelled

paulmillar:
Motivation:

We have had reports (see #6557) where a migration job was cancelled;
however, the reason the job was cancelled is not clear.  Currently, the
pool logs only `Task was cancelled`.

Modification:

Update PoolMigrationCancelMessage to include a reason (as a simple
String), explaining the motivation behind cancelling the migration job.

The controlling job is updated to populate this explanation.  Note that
this information is already available if the task is explicitly
cancelled (i.e., outside of the FSM).

The target pool is updated to log the explanation from the
PoolMigrationCancelMessage if one is provided.

Result:

The pool now provides more information when a migration job was
cancelled.

Target: master
Request: 8.0
Request: 7.2
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/13501/
Acked-by: Lea Morschel


svemeyer:
retest this please

PR #6596, opened 2022-04-20
info: fix broken unit-test, remove redundant one

lemora:
Motivation:
A unit test logs an InterruptedException from an internally used lambda. Another test tests almost the same thing but uses a faulty wait call outside of a monitor, thus throws an IllegalMonitorStateException.

Modification:
Fix the first test by catching the (expected) exception.
Remove the second faulty and redundant test case.

Result:
Tests shouldn't fail anymore.

Target: master
Request: 8.0
Request: 7.2
Request: 7.1
Request: 7.0
Request: 6.2
Requires-notes: no
Requires-book: no
Patch: https://rb.dcache.org/r/13534/
Acked-by: Albert Rossi


svemeyer:
retest this please

Issue #6603, opened 2022-04-22
Bug statisticsDomain


cristinamanzano:
When trying to start dCache the following output is logged in the statisticsDomain log:

```
Apr 22 13:44:56 zam288 dcache@statisticsDomain: 22 Apr 2022 13:44:56 (System) [] Bug detected in dCache; please report this to <support@dcache.org> with the following information.
Apr 22 13:44:56 zam288 dcache@statisticsDomain: java.lang.IllegalArgumentException: Either <baseDirectory> or <htmlBase> doesn't exist
Apr 22 13:44:56 zam288 dcache@statisticsDomain: at diskCacheV111.services.PoolStatisticsV0.starting(PoolStatisticsV0.java:231)
Apr 22 13:44:56 zam288 dcache@statisticsDomain: at dmg.cells.nucleus.CellAdapter.prepareStartup(CellAdapter.java:681)
Apr 22 13:44:56 zam288 dcache@statisticsDomain: at dmg.cells.nucleus.CellNucleus.doStart(CellNucleus.java:899)
Apr 22 13:44:56 zam288 dcache@statisticsDomain: at dmg.cells.nucleus.CellNucleus.lambda$wrapLoggingContext$3(CellNucleus.java:735)
Apr 22 13:44:56 zam288 dcache@statisticsDomain: at com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.runInterruptibly(TrustedListenableFutureTask.java:125)
Apr 22 13:44:56 zam288 dcache@statisticsDomain: at com.google.common.util.concurrent.InterruptibleTask.run(InterruptibleTask.java:57)
Apr 22 13:44:56 zam288 dcache@statisticsDomain: at com.google.common.util.concurrent.TrustedListenableFutureTask.run(TrustedListenableFutureTask.java:78)
Apr 22 13:44:56 zam288 dcache@statisticsDomain: at org.dcache.util.BoundedExecutor$Worker.run(BoundedExecutor.java:247)
Apr 22 13:44:56 zam288 dcache@statisticsDomain: at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
Apr 22 13:44:56 zam288 dcache@statisticsDomain: at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[...]

PR #6612, opened 2022-04-28
dcache-core: improve error message for inexistent statistics path

lemora:
Motivation:
The statistics domain writes files into a configurable directory, which it expects to be present. If it is not, it will throw an exception that does not make clear were the problem is coming from.

Modification:
Change the error message to give hints to the admin as to why the cell fails to start.

Result:
Hopefully easier to interpret error message in case the statistics cell fails to start due to path issues.

Target: master
Request: 8.0
Request: 7.2
Request: 7.1
Request: 7.0
Request: 6.2
Fixes: #6603
Requires-notes: yes
Requires-book: no
Patch: https://rb.dcache.org/r/13546/
Acked-by: Paul Millar


svemeyer:
retest this please

PR #6622, opened 2022-04-28
dcache-xrootd: fix TPC rendezvous to work with token authorization

alrossi:
Motivation:

A bit of history.

When xrootd first implemented TPC, it used a schema
whereby the initiating client would do an open on both
the source and destination servers, passing to
them a generated "rendezvous key"; when the third-party
client then connects to the source, it should have that
key in its possession; the source server validates that
the key is the same as the one the client used to on
open, and then allows the third-party client to proceed
to open the file (in our case, start the mover).

After delegation was implemented, this strategy could
be short-circuited (the client avoids calling open on
the source); designated "TPC Lite."

Because the rendezvous token carries only implicit
authorization and no authentication, in order to
support a third-party client that connects without
authenticating (say, via a certificate), the code was
modified to make the TPC Subject = ROOT, since it
would only be reading the file, never writing.

However, when JWT token authorization was introduced,
this strategy accidentally got defeated by indicating
that the presence of a token meant the open could
take place immediately.  While this may be true for
the TPC client, it is not true for the initiating
client.  In the case where the TPC client has no
token but the initiating client does, the former
will sit there waiting for the rendezvous key forever.

Modification:

Change the logic to create the rendezvous key even in
the presence of the authz CGI, except on the TPC client.
This will allow for the rendezvous authorization of the third-party
client without a token even if the initiator originally was
authorized/authenticated or is using a JWT token.

If the TPC client is in fact presenting a JWT token,
the rendezvous store-and-wait is aborted.

Result:

Rendezvous TPC without requiring a JWT token to be passed
by the third-party client is possible (again).

Target: master
Request: 8.0
Request: 7.2
Request: 7.1
Request: 7.0
Request: 6.2
Patch: https://rb.dcache.org/r/13502/
Requires-notes: yes
Requires-book: no
Ackd-by: Dmitry


alrossi:
Depends on update of xrootd4j to 4.0.12

svemeyer:
retest this please

mksahakyan:
retest this please

mksahakyan:
retest this please

PR #6661, opened 2022-05-17
Updated instructions for ssh2 keys

MartinLip:
Updated instructions to access the admin interface with ssh2 public keys. The old instructions used DSA keys. Also it is required that the key comment matches the admin user name.

Signed-off-by: Martin Lipinski <martin.lipinski@rwth-aachen.de>


kofemann:
Thanks @MartinLip

PR #6706, opened 2022-06-28
dcache-frontend: bulk-requests POST, allow either string or boolean f…

alrossi:
…or boolean values

Motivation:

Bulk version 1, prior to #13257 master@df7627b6d1cf6802267400980b487329214f9ca2,
accepted JSON boolean values for clearOnSuccess, clearOnFailure and cancelOnFailure.
When the multiple format changes were introduced, the type was changed to a string.

While use of bulk version 1 was not extensive, it is still probably a good idea
to support a strict boolean type for these attributes (as well as the new 'prestore'
attribute).  At the same time, since we have released Bulk v2 in 8.1 with
string definitions, we should continue to accept these for backward compatibility as well.

Modification:

Allow both boolean and string values for these attributes.

Result:

Avoids potential incompatibilities with scripts and supports the more natural
boolean type expression for those fields.

Target: master
Request: 8.1 (for v1)
Request: 8.0 (for v1)
Request: 7.2 (for v1)
Patch: https://rb.dcache.org/r/13596/
Requires-book: no
Requires-notes: yes
Acked-by: Tigran

### DIFF
--- a/modules/dcache-frontend/src/main/java/org/dcache/restful/resources/bulk/BulkResources.java
+++ b/modules/dcache-frontend/src/main/java/org/dcache/restful/resources/bulk/BulkResources.java
@@ -373,6 +373,9 @@ public final class BulkResources {
         String string = removeEntry(map, String.class, "activity");
         request.setActivity(string);
 
+        string = removeEntry(map, String.class, "target");
+        request.setTarget(string);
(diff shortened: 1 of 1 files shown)


svemeyer:
retest this please

PR #6716, opened 2022-06-30
Add clarification on authorized keys format

pcolladosoto:
Despite what one might expect, the comment added to entries in the authorized keys file (which is usually `/etc/dcache/admin/authorized_keys2`) must include `admin` as the user of a common SSH destination string.

This commit is based on the contributions made by Martin Lipinski, Cristoph Anton Mitterer, Thomas Hartmann and Xavier Mol through the <user-forum@dcache.org> mailing list on the `Cannot access dCache administration interface with public key anymore' thread.

### DIFF
--- a/docs/TheBook/src/main/markdown/config-admin.md
+++ b/docs/TheBook/src/main/markdown/config-admin.md
@@ -79,7 +79,12 @@ breaks) and should have a standard format, such as:
 
 > **NOTE**
 >
-> You may omit the part behind the equal sign as it is just a comment and not used by dCache.
+> The standard format for SSH keys as documented on [`sshd(8)`](https://man7.org/linux/man-pages/man8/sshd.8.html#AUTHORIZED_KEYS_FILE_FORMAT) states that a trailing
+> comment field must be included with each key and that it isn't considered for anything. An example of such comment would be `/Users/JohnDoe/.ssh/id_dsa` above.
+> As specified in `sshd(8)`, the main use for the comment is aiding in the identification of the public key contained on that very same line. The catch with the
+> SSH access to dCache's admin shell is that, despite what one might expect, this comment **must** adhere to a specific format for the key to be considered.
+> This comment **must** be of the form `admin@foobar`, where `foobar` can be any string. In other words, the comment has to specify `admin` as the user of a
+> typical SSH [destination](https://man7.org/linux/man-pages/man1/ssh.1.html#DESCRIPTION) string.
 
 Now you can login to the admin interface by
 


calestyo:
I'd still suggest to rather simply drop this non-standard behaviour of dcache instead.

pcolladosoto:
I'd also advocate for that solution. The thing is I believe it's useful to note the behaviour as it is now: if it were fixed we could just revert the documentation to its original state.

kofemann:
@pcolladosoto Thanks for the PR. Please update it to be against master, otherwise the next release will not have it.
@calestyo drorring this behavior is a valid request. However, this functionality was added with security in the mind. Can you
start a conversation on user forum to help us with the decision?

pcolladosoto:
Thanks @kofeman! The thing is these changes had already been merged to `master` on #6661. This PR contains more information and references to standard `sshd` behaviour. Should this information be added on top of #6661 or would you rather have me close the PR without merging it to master?

Sorry for the noise O_o

kofemann:
Ah, makes sense. Then you need to make PR for 8.1 and 8.0 as well. We have your backport rules : top -> down. So, yes, if here more info, then it should fo into master first.

pcolladosoto:
Hi @kofemann! If I understand correctly I need to open PRs to both `8.1` and `8.0` and also to `master`. Is that so or am I missing something?

I'd rather be sure before messing everything up...

calestyo:
@kofemann What exactly is the security benefit respectively the use case?

AFAIU, the key is anyway used just for the admin interface... and there shouldn't it be an "either-you're allowed in and can do anything - or not"?

calestyo:
But apart from that, *if* there's a need for security to specify the user, then the comment filed is IMO still an unfortunate choice.

Give that users/admins will consider it to be just that, a comment, they may by accident put information in, that cause harm to security.

Why not using the options field of `authorized_keys`? Either you could just add a custom option like `org,dcache_user=` or if your ssh lib doesn't support that, include in a well-defined one like in `environment="NAME=value"` or perhaps abuse `principals=`?

Then the comment field could be ignored again, and the ssh cell could be made printing a nice descriptive error, if - whatever option is used in the end - is missing.
That should help people to more quickly find out what’s missing.

paulmillar:
Just a quick reply, @calestyo

The security use-case is "partial" administration.

The idea is that a dCache instance might have several people who maintain different parts of an instance, but some of those people have only limited access, and cannot access to other parts of dCache.  As a concrete example, a set of pools might be administered by members of a research group (the group that provided funding for those storage nodes) but the doors and central nodes should be "off limits" for those people.

dCache's admin interface includes an ACL model, where it may be configured to control which admins are allowed to do which operations on any given cell within the dCache cluster.  The default is everyone can do everything, but it is configurable.

Naturally, having an ACL model for admin commands only makes sense if you know who people are.  That's the reason why simply having the public key isn't enough: we need also to map that public key to a specific admin account.

Your suggestions (options and environment variables) seem to be client-side hints, describing which account the client would like to log into.  The problem is rather server-side: the mapping is something dCache needs to enforce, in order for the ACL model to work.

calestyo:
> dCache's admin interface includes an ACL model, where it may be configured to control which admins are allowed to do which operations on any given cell within the dCache cluster. The default is everyone can do everything, but it is configurable.

Is that even documented somewhere?

> Your suggestions (options and environment variables) seem to be client-side hints, describing which account the client would like to log into. The problem is rather server-side: the mapping is something dCache needs to enforce, in order for the ACL model to work.

No these are not client-side, they'd also be on the server side’s `authorized_keys` just not the comment-field (last one), but the options-field (the optional first one).

For the hack idea of defining an env var in the `authorized_keys` keys file, one would need to check whether the client could override this (e.g. via `ssh`’s `SendEnv` or `SetEnv`) - but my idea would have been anyway to define a new option.

If using the option field should really prove not possible, then one could make the use of the comment field smarter:<br>
E.g. unless the comment starts with something like `org.dcache.admin.user=` the comment would be ignored and the key be mapped to an allow-everything user. Only if it matches, the set user is taken to suit your acl model.<br>
But I'd do that really only if the proper way cannot be made working (e.g. if your ssh library doesn't support adding other options. Abusing a comment field which is explicitly defined to be ignored may even harm security in the end.

The whole benefit of all this would at least be, that the question why SSH doesn't work, which pops up at least a handful of times a year, would go away ;-)


Cheers,
Chris.

kofemann:
> 

>>    dCache's admin interface includes an ACL model, where it may be configured to control which admins are allowed to do which operations on any given cell within the dCache cluster. The default is everyone can do everything, but it is configurable.

> Is that even documented somewhere?

Yes, it is: 
https://github.com/dCache/dcache/blob/master/docs/TheBook/src/main/markdown/config-admin.md#creating-a-new-user

kofeman:
Sorry I am kofeman, not kofemann.
Konstantin

mksahakyan:
@[pcolladosoto](https://github.com/pcolladosoto) could you please add the changes to 8.0 and 8.1 branches as weel.

pcolladosoto:
Hi @mksahakyan! I have just opened PRs for `8.0` and `8.1` on #6741 and #6742, respectively.

I'd really appreciate it if those two could be merged along with this one.

I'm still in doubt as to wether add these changes to the `master` branch too. Is that something you'd like?

Thanks for your time. Fell free to `@` me if anything comes up.

PR #6731, opened 2022-07-11
UPDATE_METADATA is needed for staging

onnozweers:
This was already updated on the webdav.md user guide page, but not here.


kofemann:
Thanks Onno!

PR #6771, opened 2022-09-06
common: remove orphaned test file

mothlyme:
Motivation:
Test file is no longer needed

Modification:
remove README file

Target: master

Require-book: no
Require-notes: no
Signed-off-by: Maximilian Bichel <maximilian@bichel.cc>


kofemann:
ok to test

Issue #5410, opened 2020-04-27
drop Apache HttpClient library


kofemann:
Java 9 introduced a brand new HTTP client there are no reasons to use Apache HttpClient any more.

paulmillar:
This affects the following components:
- SRM client
- delegation client
- OIDC refresh token support
- scitoken gPlazma plugin
- xacml gPlazma plugin
- HTTP-TPC support in pool

paulmillar:
Should we create a milestone for "Dropping Apache HttpClient" and create individual issues for each affected component?

PR #6774, opened 2022-09-07
Drop usage of Apache's HttpClient in dcache-webdav.

LukasMansour:
Motivation:
Starts some progress on issue #5410. 

Modification:
Modify the single class in the module dcache-webdav to no longer use Apache's HttpClient but instead use the JDK HttpClient available from JDK 9 onwards.

**N.B.** No real testing environment available, assuming dCache-tests suffice, otherwise someone should test it before throwing it on prod!

Result:
The dcache-webdav has its dependency on apache-httpclient removed (please note it is still transitively available until other modules also drop their usage)

Signed-off-by: Lukas Mansour <lukas.mansour@desy.de>

### DIFF
--- a/modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/CredentialServiceClient.java
+++ b/modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/CredentialServiceClient.java
@@ -188,11 +188,7 @@ public class CredentialServiceClient
                         )
                 )
                 .header("Authorization", "Basic " +
-                        Base64.getEncoder().encodeToString(
-                                UTF_8.encode(
-                                        clientId + ":" + (clientSecret == null ? "null" : clientSecret)
-                                ).array()
-                        )
+                        Base64.getEncoder().encodeToString((clientId + ":" + clientSecret).getBytes(UTF_8))
                 )
                 .header("Content-Type", "application/x-www-form-urlencoded")
                 .build();
--- a/modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/CredentialServiceClient.java
+++ b/modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/CredentialServiceClient.java
@@ -47,6 +47,7 @@ import java.security.cert.X509Certificate;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


kofemann:
ok to test

kofemann:
Thanks @LukasMansour . The Jenkis will take care of testing.

modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/CredentialServiceClient.java
```diff
@@ -164,48 +156,58 @@ public StaticOpenIdCredential getDelegatedCredential(String token,
...
+                                                "&scope=%s",
+                                        GRANT_TYPE,
+                                        clientId,
+                                        token,
+                                        TOKEN_TYPE,
+                                        SCOPE)
+                        )
+                )
+                .header("Authorization", "Basic " +
+                        Base64.getEncoder().encodeToString(
```
kofemann, line 191: This looks too complicated to me. Why not simple:  

```
Base64.getEncoder().encodeToString((clientId + ":" + clientSecret).getBytes(UTF_*));
```

modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/CredentialServiceClient.java
```diff
@@ -164,48 +156,58 @@ public StaticOpenIdCredential getDelegatedCredential(String token,
...
-        return post;
+    private HttpRequest buildRequest(String token, String host, String clientId, String clientSecret)
+            throws UnsupportedEncodingException{
+
+        return HttpRequest.newBuilder()
+                .uri(URI.create(tokenEndPoint(host)))
+                .POST(
+                        HttpRequest.BodyPublishers.ofString(
+                                String.format(
+                                        "grant_type=%s" +
```
kofemann, line 178: you cannot simply concatenate strings as they must be URL encoded

modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/CredentialServiceClient.java
LukasMansour: I didn't know the interaction of "String" + null = "Stringnull", you learn something new every day.

Charset.encode() and String.getBytes() do different things (but should achieve the same result) and I was keeping parity with what Apache HTTP had done. I'll change to getBytes() since from JDK 18 onwards UTF_8 is standard ([See JEP 400](https://openjdk.org/jeps/400)), which means getBytes() is probably quicker thereafter and most likely the same until then.

modules/dcache-webdav/src/main/java/org/dcache/webdav/transfer/CredentialServiceClient.java
LukasMansour: I thought the BodyPublisher took care of it, ofc in Hindsight this is not the case and cannot be the case.

LukasMansour:
I will work on some new tests for actually ensuring the output is behaving the same as before.

PR #6779, opened 2022-09-13
gplazma: vo-group plugin

DmitryLitvintsev:
Motivation:
vo-group plugin reads json file containing fqan to GID/name mapping and creates an in-memory map. The file habitually contains glob patterns instead of fqans. To setup most "specific match first" the map key has to be sorted (in reverse order).

Modification:
Sort map by glob key.

Result:
The mapping is robust against order in which dictionaries are specified in the input file.

Patch: https://rb.dcache.org/r/11575/
Target: trunk
Request: 8.2


svemeyer:
retest this please

PR #6781, opened 2022-09-15
Book: describe `r` and `g` inheritance flags

onnozweers:
Sources: discussion with Tigran on the user forum on October 26 2021, and the source: https://github.com/dCache/dcache/blob/master/modules/acl-vehicles/src/main/java/org/dcache/acl/enums/AceFlags.java

As far as I can see, this has been in the source since 8 years ago, so I think it can be backported to all currently supported releases. https://github.com/dCache/dcache/blame/d9226110996ff71e69bc54260dd7f2959e862edf/modules/acl-vehicles/src/main/java/org/dcache/acl/enums/AceFlags.java#L33


kofemann:
Thanks, Onno!

PR #6803, opened 2022-09-29
dcache-xroot:  select proxy door address on basis of reachability for…

alrossi:
… client

Motivation:

On dual stack deployments like BNL's, it is possible that only the IPv4 address is exposed to the outside
world.   In this case, we need to make sure the proxy is
reachable.

Modification:

Instead of using the default local address, bind the proxy to a local address that respects the network of the client and is likely reachable by it.  This is done by a simple call to NetworkUtils.getLocalAddress(client).

Result:

"Special" network configurations seem to work with this change.

Target: master
Request: 8.2
Patch: https://rb.dcache.org/r/13696/
Closes: RT 10350
Requires-notes: yes
Requires-book: no
Acked-by: Dmitry


alrossi:
I pushed https://rb.dcache.org/r/13697/ to this pull-request rather than having to fix the merge manually.

PR #6856, opened 2022-11-11
dcache-bulk:  fix admin command options (uppercase)

alrossi:
Motivation:

There is a wrong option for target type (REGULAR, not FILE).

Also, the options are not consistently indifferent to UC/LC.

Modification:

Fix type and catch the instances where toUpperCase need to be applied.

Result:

Admin commands are now consistent.

Target: master
Request: 8.2
Patch: https://rb.dcache.org/r/13758/
Requires-notes: yes
Requires-book: no
Acked-by: Tigran


mksahakyan:
retest this please

PR #6876, opened 2022-11-22
dcache-bulk:  add id,rid index on request target table

alrossi:
Motivation:

Retrieval of potentially long lists in bulk, e.g., of targets or even requests, use implicit (admin)
or explicit (REST) paging to fetch batches at
a time.  This means that they need to be ordered
sequentially and a limit applied on the database
query.

Without indexing on Y, for a query like
"SELECT ... FROM ... WHERE X=x ORDER BY Y LIMIT Z" the Postgres query optimizer elects to
do a full scan with a filter on the X
condition (without the limit, it will use a
bitmap scan on X).

Modification:

Add the necessary index (actually a composite here) on the request_target table to facilitate this query.

Result:

Significant speed-up on retrieval of target rows;
this is essential once the table grows to > 100M.

Target: master
Request: 8.2
Requires-notes: yes
Patch: https://rb.dcache.org/r/13781/
Acked-by: Lea


svemeyer:
retest this please

PR #6878, opened 2022-11-22
frontend.srr.shares text rewritten

onnozweers:
I had some difficulties understanding what to do with frontend.srr.shares; I only figured it out after reading the Book and finding a real life example. I took the liberty of rewriting the documentation here so that it hopefully is a little bit more clear.


kofemann:
thanks, @onnozweers

PR #6883, opened 2022-11-25
dcache-bulk: initialize/load and reset on executor thread

alrossi:
Motivation:

Bulk reload on restart munges the state of all the non-terminal requests, which means resetting / deleting targets.  This can take a bit of time.

Currently, this is all done on the cell startup thread, which is not optimal, because it can block access to the cell.  What needs to take place is something similar to what happens in resilience/ QoS:  an initializer executes these routines on a separate thread, and then starts the manager.  This allows for logging into the cell, and even for submission of new requests (which will be queued).

Modification:

Execute the second half of the intialization on an incoming executor thread.  Add an executor (single threaded) to the BulkServiceCommands and execute reset on that queue as well.

Result:

The bulk service only locks at the beginning until the namespace has been successfully pinged.  All reloading of previous tasks takes place under the covers.  Note that the manager still starts only after the reloading activity has completed.  Incoming requests will of course be deprioritized as their timestamps are posterior to the reloaded requests.

Target: master
Request: 8.2
Patch: https://rb.dcache.org/r/13787/
Requires-notes: yes
Acked-by: Tigran

### DIFF
--- a/modules/dcache-bulk/src/main/java/org/dcache/services/bulk/BulkService.java
+++ b/modules/dcache-bulk/src/main/java/org/dcache/services/bulk/BulkService.java
@@ -140,32 +140,7 @@ public final class BulkService implements CellLifeCycleAware, CellMessageReceive
          */
         waitForNamespace();
 
-        /*
-         * See store specifics for how reload is handled, but the minimal contract is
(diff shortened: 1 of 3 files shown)


svemeyer:
retest this please

PR #6894, opened 2022-11-30
dcache-bulk: insert targets immediately on submit

alrossi:
Motivation:

The TAPE API expects to be able to do a GET
immediately following a POST of a stage request;
the current FTS implementation fails the request
if all the paths are not returned for that GET.
The spec provides that any waiting target/path
have the SUBMITTED state.

Unfortunately, the current implementation does
not insert the target paths into the table with
their appropriate state until the request actually runs (done either before processing with `-prestore` or after it without).  This is an error,
particularly in consideration that requests may
be queued up a long time.

Modification:

Insert all explicit initial paths directly into
the table upon arrival.  To facilitate this,

- a batched version of insert is used there
- subsequent calls in the container to insert (with attribute info from the namespace) are done using upsert for uniformity with discovered paths (from recursion).

Result:

We adhere to the reasonable spec requirements.

Target: master
Request: 8.2
Requires-note:  yes
Patch: https://rb.dcache.org/r/13782/
Depends-on: #13779
Acked-by: Tigran

### DIFF
--- a/modules/common/src/main/java/org/dcache/db/JdbcUpdate.java
+++ b/modules/common/src/main/java/org/dcache/db/JdbcUpdate.java
@@ -30,7 +30,7 @@ import javax.annotation.Nullable;
  */
 public abstract class JdbcUpdate {
 
-    final Map<String, Object> updates = new LinkedHashMap<>();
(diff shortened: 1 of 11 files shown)


svemeyer:
retest this please

PR #6918, opened 2022-12-09
dcache-frontend: handle parse exceptions in bulk and tape resources

alrossi:
Motivation:

JSONException and the various exceptions thrown by GSON are RuntimeExceptions; if they are not caught and converted into HTTP exceptions, they generate
stack traces.

Modification:

Wrap the relevant code with try ... catch blocks.

Result:

No stack traces when  badly formed objects are
sent by users.

Target: master
Request: 8.2
Patch: https://rb.dcache.org/r/13816/
Requires-notes: yes
Acked-by: Lea


lemora:
retest this please

Issue #6914, opened 2022-12-07
API: changing QoS says "success" but doesn't work (master snapshot 9.0)


onnozweers:
Dear dCache devs,

Changing the QoS of a file through the API seems broken in the master 9.0 snapshot. The API reports back "success" but the QoS of the file is not changed.
In our production instance (version 7.2) it works as expected.

Here's what I did:

Getting a macaroon with these properties:
```
location Optional.empty
identifier dIGgJazo
cid iid:Ru2zIwqu
cid id:31029;31040,40304,30013;onno
cid before:2022-12-07T22:03:45.516158Z
cid home:/users/onno
cid root:/users/onno
cid activity:LIST,DOWNLOAD,UPLOAD,MANAGE,DELETE,READ_METADATA,UPDATE_METADATA
cid ip:145.38.0.0/16,145.100.5.0/27,145.100.5.210/26,145.100.32.0/22,145.100.48.0/23,145.100.50.0/23,145.100.200.0/21,145.100.9.64/29,145.101.32.0/21,145.100.56.0/22,2001:610:108::/48
```
Listing a file:
```
11:07 ui.grid.surfsara.nl:/home/onno 
onno$ curl --config /home/onno/.ada/headers/authorization_header_3yL6CRUz2IfM -H 'accept: application/json' --fail --silent --show-error --ipv4 -X GET 'https://dcachetest.grid.surfsara.nl:20443/api/v1/namespace/%2Ftape%2Ftest?locality=true&qos=true'
{
  "fileMimeType" : "application/octet-stream",
  "fileLocality" : "ONLINE_AND_NEARLINE",
  "currentQos" : "tape",
  "labels" : [ ],
  "size" : 10485760,
  "creationTime" : 1669736063762,
  "fileType" : "REGULAR",
  "pnfsId" : "0000EC9AEB2B04304A8D8E34F530B55198F5",
  "nlink" : 1,
  "mtime" : 1669736064395,
  "mode" : 384
}
```
Changing the QoS returns "success":
```
11:08 ui.grid.surfsara.nl:/home/onno 
onno$ curl --config /home/onno/.ada/headers/authorization_header_3yL6CRUz2IfM -H 'accept: application/json' --fail --silent --show-error --ipv4 -H 'content-type: application/json' -X POST https://dcachetest.grid.surfsara.nl:20443/api/v1/namespace/%2Ftape%2Ftest -d '{"action":"qos","target":"disk+tape"}'
{"status":"success"}
```
[...]

PR #6919, opened 2022-12-09
dcache-qos,frontend:  add exception to remote client method and retur…

alrossi:
…n 500 if fails

Motivation:

See GitHub issue https://github.com/dCache/dcache/issues/6914. If there are no qos services running, the namespace qos transition should fail.

Modification:

Change the send of that message to sendAndWait with message return; catch and handle the checked exceptions.

Result:

Desired behavior (returns 500 with error message when services are not reachable).

Target: master
Request: 8.2
Request: 8.1
Patch: https://rb.dcache.org/r/13810/
Requires-notes: yes
Requires-book: no
Closes: #6914
Acked-by: Tigran


lemora:
retest this please

Issue #6909, opened 2022-12-04
Proxy xrootd door keep restarting


vingar:
Hello,

The Xrootd doors enabled with `xrootd.net.proxy-transfers=true` keep restarting on one of our instances.

The dCache version is 8.2.2.

Some observed error in the logs (see file 
[xrootd2-dcdndoor03Domain.log](https://github.com/dCache/dcache/files/10148865/xrootd2-dcdndoor03Domain.log)
in copy) are:

```04 Dec 2022 04:18:49 (Xrootd-dcdndoor03-externalsubnet) [] Thread Group "Xrootd-dcdndoor03-externalsubnet-threads":
Thread: pool-14-thread-1 [A--] (5) TIMED_WAITING
    java.base@11.0.16/jdk.internal.misc.Unsafe.park(Native Method)
    java.base@11.0.16/java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:234)
    java.base@11.0.16/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.awaitNanos(AbstractQueuedSynchronizer.java:2123)
    java.base@11.0.16/java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.take(ScheduledThreadPoolExecutor.java:1182)
    java.base@11.0.16/java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.take(ScheduledThreadPoolExecutor.java:899)
    java.base@11.0.16/java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1054)
    java.base@11.0.16/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1114)
    java.base@11.0.16/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
    java.base@11.0.16/java.lang.Thread.run(Thread.java:829)
04 Dec 2022 04:18:49 (Xrootd-dcdndoor03-externalsubnet) [] Forcefully interrupting thread pool-14-thread-1 during shutdown.
04 Dec 2022 04:18:50 (Xrootd-dcdndoor03-externalsubnet) [] Thread Group "Xrootd-dcdndoor03-externalsubnet-threads":
Thread: pool-14-thread-1 [A--] (5) TIMED_WAITING
    java.base@11.0.16/jdk.internal.misc.Unsafe.park(Native Method)
[...]

PR #6945, opened 2023-01-09
dcache-xroot: handle outbound errors on channel promise

alrossi:
Motivation:

Refer to the final comments in
GH 6909 "Proxy xrootd door keep restarting"
https://github.com/dCache/dcache/issues/6909

Occasionally we will find unexpected errors on the pipeline being reported as with the example below

```
04 Dec 2022 04:07:55 (Xrootd-dcdndoor03-externalsubnet) [] An exceptionCaught()
event was fired, and it reached at the tail of the pipeline. It usually means
the last handler in the pipeline did not handle the exception.
io.netty.channel.StacklessClosedChannelException: null
	at io.netty.channel.AbstractChannel$AbstractUnsafe.write(Object, ChannelPromise)(Unknown Source)
```

The problem arises from the asymmetical nature of
exception propagation on the Netty pipeline.  For
exceptions on inbound/read, all that is needed is
for the last handler in the pipeline (the "TOP")
to implement exceptionCaught.  This is done by
the final handlers in the door, proxy and pool
pipelines.

However, for outbound exceptions, one needs to
add a listener to the channel future or channel
promise; this needs to be done at the very
beginning of the outbound pipeline ("BOTTOM");
the listener allows exceptions on channel write
to invoke the uncaughtException method on the
inbound handlers.

Modification:

Create an OutboundHandler that simply adds the
listener and add it first to all pipelines.

Result:

We should no longer see exceptions such as the one above reporting no exception handler to handle
the exception.

Target: master
Request: 8.2
Request: 8.1 (Without proxy)
Request: 8.0 (Without proxy)
Request: 7.2 (Without proxy)
Closes: #6909
Patch: https://rb.dcache.org/r/13833/
Requires-notes: yes
Acked-by: Tigran

### DIFF
--- a/modules/dcache-xrootd/src/main/java/org/dcache/xrootd/OutboundExceptionHandler.java
+++ b/modules/dcache-xrootd/src/main/java/org/dcache/xrootd/OutboundExceptionHandler.java
@@ -1,0 +1,36 @@
+/* dCache - http://www.dcache.org/
+ *
+ * Copyright (C) 2014 Deutsches Elektronen-Synchrotron
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


svemeyer:
retest this please

PR #6972, opened 2023-01-24
fix binary download URL in README.md

jamesbraid:
Update the binary download URL to remove the version, as the original URL is 404


kofemann:
Thanks, @jamesbraid !

PR #6998, opened 2023-02-04
dcache-xroot:  return kXR_ItExists error code when file exists

alrossi:
Motivation:

From v4 to v5, there were some changes in
the xrootd protocol affecting error codes.
One of these was the addition of kXR_ItExists.

It was not immediately apparent that the
code that we were returning (kXR_unsupported
in 2014, then kXR_NotAuthorized) would
cause some clients to react differently;
but in any case, the neglect to change
this has provoked potential data loss
with gfal, which aggressively removes
the file/directory and tries to overwrite
in this situation when doing `copy`.

Modification:

Make sure that when a file exists we
return kXR_ItExists.

Result:

Aggressive clients should be blocked
from pre-emptive removal.

Target: master
Request: 8.2
Request: 8.1
Request: 8.0
Request: 7.2
Patch: https://rb.dcache.org/r/13873
Requires-notes: yes
Acked-by: Dmitry


svemeyer:
retest this please

PR #7004, opened 2023-02-08
dcache-bulk,dcache-frontend:  fix bug in tape/stage and in argument e…

alrossi:
…xtraction from DB

Motivation:

See RT 10426 Inconsistencies in Pin Lifetime between dCache-like pinning https://rt.dcache.org/Ticket/Display.html?id=10426

It was noted that the diskLifetime value was always default for /tape/stage, unlike for /bulk-requests with -activity=PIN.

It turns out the fundamental problem was that the frontend resource was neglecting to set the map of arguments on the request.

Once this was done, however, a second problem was discovered. The serialized value in the database was not being parsed back
into the correct JSON string.   This has to do with the fact
that that arguments is a <String, String> map, but the
stringified value was now being used to store a JSON object
string instead of a simple value, meaning that it contains
the same delimiters as the top level map does (: and ,).

Modification:

Adding a surrounding '{' '}' to the value fetched from the database and parsing it as a JSONObject solved the problem.

Result:

The frontend resource now adds the arguments map as it should.

Target: master
Request: 8.2
Bug: #10426
Requires-notes: yes
Patch: https://rb.dcache.org/r/13877/
Acked-by: Dmitry

### DIFF
--- a/modules/dcache-bulk/src/main/java/org/dcache/services/bulk/activity/plugin/pin/StageActivity.java
+++ b/modules/dcache-bulk/src/main/java/org/dcache/services/bulk/activity/plugin/pin/StageActivity.java
@@ -133,8 +133,14 @@ public final class StageActivity extends PinManagerActivity {
     @Override
     protected void configure(Map<String, String> arguments) {
         if (arguments != null) {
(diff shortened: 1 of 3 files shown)


svemeyer:
retest this please

PR #7015, opened 2023-02-11
dcache-qos,psu:  allow getStorageUnit to resolve regex

alrossi:
Motivation:

The Resilience service unpacked the PoolSelectionUnit into its own internal map and had its own storage unit resolution code.

When transposed into QoS, the decision was made that the full special processing of PSU information was no longer necessary, so QoS relies on the PoolManager code more than Resilience did.

However, an oversight of this migration occurred in not realizing the `getStorageUnit` method did not
resolve regular expressions (when that option is on) as does the `match` method, and as Resilience did on its own.

Modification:

Change the `getStorageUnit` method to attempt unit name resolution in the event that it cannot find
the literal unit name mapped.

We also make a small correction to the qos-engine
(unnecessary class cast).

Result:

QoS can now function as it should when the PSU
regex option is `on`.

Target: master
Request: 8.2
Request: 8.1
Patch: https://rb.dcache.org/r/13888/
Requires-notes: yes
Acked-by: Dmitry


svemeyer:
retest this please

PR #7040, opened 2023-03-03
Allow for getting Migration Information per REST.

LukasMansour:
Motivation:
Allows the status of a migration to be requested with the poolname and ID. This will most likely be necessary for projects in the KAI Cooperation.

Modification:
Makes submitMigrationCopy method to return 401 and 403 correctly.
Adds getMigrationInfo(Pool, ID) which will return a JSON Object containing the migration information (status, completion, targets, etc.).

Result:
Now we can not only start but also see the information of migrations on pools.

There are small TODOs left to do with formatting, one could discuss if a more JSON-like encoding is even necessary for these fields (see: 'running tasks' and 'most recent errors' in the JSON Response).

### DIFF
--- a/modules/dcache-frontend/src/main/java/org/dcache/restful/resources/migration/MigrationResources.java
+++ b/modules/dcache-frontend/src/main/java/org/dcache/restful/resources/migration/MigrationResources.java
@@ -245,10 +245,10 @@ public final class MigrationResources {
     }
 
     /**
-     * Gets migration information to a specified migration ID on a pool.
+     * Gets migration information for the specified migration job id on the specified pool.
      */
     @GET
-    @ApiOperation(value = "Gets information to a migration. (See Pool Operator Commands 'migration info')")
+    @ApiOperation(value = "Gets migration information for the specified migration job id on the specified pool. (See Pool Operator Commands 'migration info')")
     @ApiResponses({
           @ApiResponse(code = 200, message = "OK"),
           @ApiResponse(code = 400, message = "Bad request"),
@@ -342,7 +342,6 @@ public final class MigrationResources {
             throw new InternalServerErrorException(e);
         }
     }
-
 
     private static void conditionalAppendString(String cmdParam, String jsonKey, StringBuilder sb,
           JSONObject jsonPayload) {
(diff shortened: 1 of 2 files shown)


modules/dcache-frontend/src/main/java/org/dcache/restful/resources/migration/MigrationResources.java
```diff
@@ -232,9 +242,108 @@ public Response submitMigrationCopy(@ApiParam(
...
+
+                }
+            }
+            return migInfoOutput;
+        } catch (CacheException | NoRouteToCellException | InterruptedException e) {
+            throw new InternalServerErrorException(e);
+        }
     }
 
+
```
lemora, line 346: Please remove the new line

modules/dcache-frontend/src/main/java/org/dcache/restful/resources/migration/MigrationResources.java
```diff
@@ -232,9 +242,108 @@ public Response submitMigrationCopy(@ApiParam(
         } catch (InterruptedException | CacheException | NoRouteToCellException e) {
             throw new InternalServerErrorException(e);
         }
+    }
+
+    /**
+     * Gets migration information to a specified migration ID on a pool.
+     */
+    @GET
+    @ApiOperation(value = "Gets information to a migration. (See Pool Operator Commands 'migration info')")
```
lemora, line 251: Maybe  "information to a" -> "information for a"

lemora (commented):
Small nitpicky, non-blocking things :)

modules/dcache-frontend/src/main/java/org/dcache/restful/resources/migration/MigrationResources.java
LukasMansour: Agreed. Fixed in [#e0a1266](https://github.com/dCache/dcache/pull/7040/commits/e0a1266c9d7b00f1699821fa31979c19c6f89668)

lemora (approved):

kofemann (approved):

Issue #6932, opened 2022-12-21
API/namespace: mv on a dir with relative target makes it disappear


onnozweers:
Dear dCache devs,

I'm trying to rename a directory /users/onno/disk/testdir1/ to /users/onno/disk/testdir2/. When I use a relative target, the directory disappears. I tested it with Swagger and basic auth (so it's not a problem with macaroons with a root: property). dCache fairly recent master snapshot 9.0.

This is the curl command listed by Swagger:

```
curl -X POST "https://dcachetest.grid.surfsara.nl:20443/api/v1/namespace/%2Fusers%2Fonno%2Fdisk%2Ftestdir1" -H  "accept: application/json" -H  "content-type: application/json" -d "{\"action\":\"mv\",\"destination\":\"testdir2\"}"
```

I would expect that a mv from `testdir1` to `testdir2` would rename the dir. If that is not possible, I would expect an error message.

If I specify the full path as target, like this:

```
curl -X POST "https://dcachetest.grid.surfsara.nl:20443/api/v1/namespace/%2Fusers%2Fonno%2Fdisk%2Ftestdir1" -H  "accept: application/json" -H  "content-type: application/json" -d "{\"action\":\"mv\",\"destination\":\"/users/onno/disk/testdir2\"}"
```

... it works as expected.

Is it supposed to happen that a mv from `testdir1` to `testdir2` makes testdir1 disappear?

There is a slightly related ticket https://github.com/dCache/dcache/issues/5331, but I think this case is different.

Cheers,
Onno

alrossi:
Onno,

so sorry this got lost in the shuffle!

I have replicated this bug.   Will investigate what is going on.   

Cheers, Al

alrossi:
Onno,

could you also maybe create a junk directory on production and try to rename it using a relative path?   (You are running 8.2, I take it.)

That confirmation would be useful.

Thanks, Al

PR #7042, opened 2023-03-03
dcache-frontend: resolve on parent of source when renaming

alrossi:
Motivation:

Use of relative path when issuing `mv` through
the RESTful `/api/v1/namespace` is broken.

Modifcation:

Call resolve with the parent of the path.

Result:

Fixes problem.

Target: master
Request: 8.2
Request: 8.1
Request: 8.0
Request: 7.2
Requires-notes: yes
Closes: #6932
Patch: https://rb.dcache.org/r/13905/
Acked-by: Lea


mksahakyan:
retest this please

Issue #7095, opened 2023-03-30
[The Book] Revocation entries with grid-vorolemap and gridmap plugin do not work


XMol:
Hello dCache.org,

this week's EGI security drill revealed something that flew under the radar, which is that the revocation entries as [documented in The Book to this day](https://www.dcache.org/manuals/Book-8.2/config-gplazma.shtml#revocation-entries) actually do not have the intended effect!

Please refer to dCache RT \#7834 for the full story. The short of it is, that the ban plugin was introduced precisely because no other plugin could address the need to explicitly revoke authentication of specific users/vos. Prompted by the security drill, I tried a revocation entry for the reported DN and found that gPlazma (dCache v8.2.10) ignored it. An acceptable work around for now is to map to a bogus virtual account (e.g. "ban"), for which the map session won't yield useful results.

/label documentation

Best regards,
Xavier.

XMol:
All pull requests for the active branches have been merged.

PR #7102, opened 2023-03-30
Revocation entries are defunct since gplazma-1 is discontinued

XMol:
Addresses #7095 for dCache version master.


kofemann:
Thanks, @XMol !

PR #7106, opened 2023-03-31
bulk: let bulk propagate Subject for pin/stage

lemora:
Motivation:
When making pin, stage or unpin requests via `PinManager`, the bulk service currently does not propagate the subject to PinManager. If it did, that information would be used by PinManager/PoolManager to enforce permissions such as if a subject is allowed to stage.

Modification:
Add the known subject to `PinManagerPinRequest` and `PinManagerUnpinMessage` messages issued by bulk.

Result:
Better enforcement of subject-based permissions.

Target: master
Request: 9.0
Request: 8.2
Request: 8.1
Request: 8.0
Request: 7.2
Requires-notes: no
Requires-book: no
Patch: https://rb.dcache.org/r/13949/
Acked-by: Albert Rossi


svemeyer:
retest this please

PR #7118, opened 2023-04-12
dcache-bulk: fix prefix-stripping logic when returning request info

alrossi:
Motivation:

The logic whereby the decision to strip the prefix from recursive targets in https://rb.dcache.org/r/13937/ `dcache-frontend: support VO- or user-root relative paths in requests` is incorrect.

Modification:

Check to see if any initial target has the prefix; if so, it should not be stripped from the recursive targets.

Result:

Correct paths for recursively found targets are returned; i.e., absolute if the user submitted absolute paths, relative is the user used relative paths on the
initial targets.

Target: master
Request: 9.0
Requires-notes: yes
Patch: https://rb.dcache.org/r/13955/
Acked-by: Tigran


lemora:
retest this please

PR #7119, opened 2023-04-12
webdav: preserve request URL in protocol info

kofemann:
Motivation:
When http request redirected to a pool, then new location contains the dcache internally resolved file path, which doesn't match the prefix anymore. Thus if a pool redirects a client back, then client will get 404.

Modification:
Update HttpTransfer#createProtocolInfo to use the original request path then creating ProtocolInfo

Result:
correct behaviour on redirect with prefix.

Ticket: #10446
Acked-by: Paul Millar
Target: master, 9.0, 8.2
Require-book: no
Require-notes: yes
(cherry picked from commit 8a4273fa8ce478fd5dffa425728fe505b367d629)


kofemann:
/
retest this please

Issue #7113, opened 2023-04-09
Symlink behavior broken for SE-tokens


vokac:
Since dCache 8.2.17 (https://github.com/dcache/dcache/commit/18fa39ee2789ec82c3cd92dd973b5b7b82828e9f) SE issued macaroon tokens no longer work correctly and it is impossible to access resources via symlink using WebDAV doors, e.g.  with this configuration
```
[root@dcache /]# chimera ls /dpm/farm.particle.cz/home/
total 3
drwxr-xr-x 3 0 0 512 Apr 09 11:59 .
drwxr-xr-x 3 0 0 512 Apr 09 11:57 ..
lrwxrwxrwx 1 0 0   6 Apr 09 11:59 atlas -> /atlas
[root@dcache /]# chimera ls /atlas/1M
total 1025
-rw-rw-r--  1 2002 2000 1048576 Apr 12  2020 .
drwxrwxr-x 42 2000 2000     512 Jan 23 17:05 ..
[root@dcache /]# chimera ls /dpm/farm.particle.cz/home/atlas/1M
total 1025
-rw-rw-r--  1 2002 2000 1048576 Apr 12  2020 .
drwxrwxr-x 42 2000 2000     512 Jan 23 17:05 ..
```
I'm not able download file from https://dcache.farm.particle.cz/dpm/farm.particle.cz/home/atlas/1M
```bash
$ export SRC=https://dcache.farm.particle.cz/dpm/farm.particle.cz/home/atlas/1M
$ export TSRC=$(curl --silent --cert /tmp/x509up_u$(id -u) --key /tmp/x509up_u$(id -u) --cacert /tmp/x509up_u$(id -u) --capath /etc/grid-security/certificates -X POST -H 'Content-Type: application/macaroon-request' -d '{"caveats": ["activity:DOWNLOAD"], "validity": "PT30M"}' "$SRC" | jq -r '.macaroon')
$ python -c "import macaroons; print(macaroons.deserialize('$TSRC').inspect())"
location Optional[/dpm/farm.particle.cz/home/atlas/1M]
identifier 6Am25+k/
cid iid:e1mJvSZH
cid id:2011;2011,2100,2011,2000;atlas_production
cid before:2023-04-09T11:10:34.374580Z
cid path:/dpm/farm.particle.cz/home/atlas/1M
cid activity:DOWNLOAD
signature bff40688dc72c20c78f3bfebb00298d89197c97953980e8e1bb29481bfc87ef3
[...]

PR #7135, opened 2023-05-01
commons,dcache:  alternative symlinks resolution on restrictions

alrossi:
Motivation:

See GitHub #7113 Symlink behavior broken for SE-tokens https://github.com/dCache/dcache/issues/7113

commit master@7cb0cd0c5f5c1debe97ee0ff868e5455f33fefdb https://rb.dcache.org/r/13908/

introduced symlink resolution for all targets.
The assumption was that configurations (including
token scopes) would not use symlinks to express
path restrictions.

However, macaroons constitute a special case and
were not tested.  The upgrade to stable branches
incorporating that commit thus break when the
user provides a symlinked path to construct a
macaroon.

Modification:

Inject a resolver function into Restrictions
and have the restrictions resolve both their
constraints as well as target.

The PnfsManager handling of restrictions
is refitted to inject the resolver.  Note
also that the resolution method now
walks up the path tree in case the
symlink check is done on a non-existent
path.

Result:

Symlinks on both constraints and targets
are permitted; thus macaroons will also
work.

Target: master
Request: 9.0
Request: 8.2
Patch: https://rb.dcache.org/r/13970/
Closes: #7113
Acked-by: Tigran

### DIFF
--- a/modules/chimera/src/main/java/org/dcache/chimera/FsSqlDriver.java
+++ b/modules/chimera/src/main/java/org/dcache/chimera/FsSqlDriver.java
@@ -1799,7 +1799,7 @@ public class FsSqlDriver {
     String resolvePath(FsInode root, String path) throws ChimeraFsException, SQLException {
         FsInode pathInode = path2inode(root, path);
         if (pathInode == null) {
-            return path;
+            return null;
         }
         return inode2path(pathInode, root);
     }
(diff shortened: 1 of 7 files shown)


lemora:
retest this please

alrossi:
retest this please

alrossi:
builds and all unit tests pass on my dev testbed SL 7 node:

```
arossi@fndcatemp1 dcache]$ uname -a
Linux fndcatemp1.fnal.gov 3.10.0-1160.83.1.el7.x86_64 #1 SMP Tue Jan 24 08:34:19 CST 2023 x86_64 x86_64 x86_64 GNU/Linux


[arossi@fndcatemp1 dcache]$ java -version
java version "11.0.8" 2020-07-14 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.8+10-LTS)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.8+10-LTS, mixed mode)

```

Issue #7129, opened 2023-04-19
Transfer with xroot using POSC fails for token with storage.modify


vokac:
With latest dCache 8.2.20 I'm getting same error as discussed in #6952 (xroot with POSC transfers using scope based authorization fails), e.g.
```shell
$ xrdfs roots://dcache.farm.particle.cz rm /wlcg/x
rm /wlcg/x : [SUCCESS] 
$ xrdcopy /etc/services roots://dcache.farm.particle.cz:1094//wlcg/x
[654.6kB/654.6kB][100%][==================================================][163.6kB/s]  
$ xrdfs roots://dcache.farm.particle.cz rm /wlcg/x
rm /wlcg/x : [SUCCESS] 
$ xrdcopy -P /etc/services roots://dcache.farm.particle.cz:1094//wlcg/x
[0B/0B][100%][==================================================][0B/s]  
Run: [ERROR] Server responded with an error: [3010] Restriction MultiTargetedRestriction[Authorisation{allowing [LIST, MANAGE, UPLOAD, DELETE, READ_METADATA, UPDATE_METADATA] on /wlcg}] denied activity UPLOAD on /upload/0/79d40e08-9643-45bf-bc9c-77885ba86fb1/x (destination)
```
for WLCG JWT token
```json
{
  "wlcg.ver": "1.0",
  "sub": "58280cfd-ed7f-4954-90c7-cfde610cb963",
  "aud": "https://wlcg.cern.ch/jwt/v1/any",
  "nbf": 1681947649,
  "scope": "storage.modify:/",
  "iss": "https://wlcg.cloud.cnaf.infn.it/",
  "exp": 1681951249,
  "iat": 1681947649,
  "jti": "c1cd1fb7-7819-4c06-aaf3-8a3fb390d80d",
  "client_id": "a6d4bb06-a7ef-40c8-92de-bc5b5e140609"
}
```
Our dCache journal show following details
```
...
Apr 20 01:47:50 dcache.farm.particle.cz dcache@doorsDomain[18606]: 20 Apr 2023 01:47:50 (Xrootd-dcache) [door:Xrootd-dcache@doorsDomain:AAX5uQ0/ceg] XrootdSessionHandler.getResponse: Request 3010
[...]

PR #7140, opened 2023-05-03
Motivation:

alrossi:
https://rb.dcache.org/r/13855/
master@40dc86223b1f664f6dad5eff4ec222fae5631d3e

was intended to allow upload (for xrootd, POSC)
when there are Multitargeted Restrictions from
tokens.  However, as

https://github.com/dCache/dcache/issues/7129#issuecomment-1531658518 `Transfer with xroot using POSC fails for token with storage.modify`

indicates, there is a bug which is erroneously
excluding user roots equivalent to '/'.

Modification:

Remove the conditional excluding '/'.

Result:

POSC works with tokens when user root is '/'.

Target: master
Request: 9.0
Request: 8.2
Request: 8.1
Request: 8.0
Request: 7.2
Closes: #7129
Requires-notes: yes
Patch: https://rb.dcache.org/r/13973/
Acked-by: Dmitry


lemora:
retest this please

Issue #7164, opened 2023-05-15
BULK:   in-memory counter is not decrementing when RELEASE activity completes


alrossi:
Example:

```
------------------ TARGETS BY STATE ------------------
         (cumulative from last service start)
CANCELLED            :             0
COMPLETED            :           603
SKIPPED              :             0
FAILED               :             0

------------ REQUEST TOTALS (since start) ------------
Requests received    :             3
Requests completed   :             3
Requests cancelled   :             0

--------------- REQUESTS (since start) ---------------
RELEASE              :             2
STAGE                :             1

---------------- REQUESTS  (current) -----------------
Active               :             0

----------------- TARGETS  (current) -----------------
RUNNING              :           400

```

There were 2 release requests on the STAGE request, each of 200 file paths.

alrossi:
I will fix this before the demo.

This does not happen with UNPIN.   The reason for this lies in the fact that the RELEASE request is ephemeral (not a separate request) so the decrement has to take place in a different code location.

alrossi:
Actually, this occurs for PIN/UNPIN as well, so the bugginess is generalized.

alrossi:
Looks like Cancel now double decrements :(

------------------ TARGETS BY STATE ------------------
         (cumulative from last service start)
CANCELLED            :             8
COMPLETED            :         71522
SKIPPED              :             0
FAILED               :             0

------------ REQUEST TOTALS (since start) ------------
Requests received    :             2
Requests completed   :             0
Requests cancelled   :             2

--------------- REQUESTS (since start) ---------------
LOG_TARGET           :             1
UNPIN                :             1

---------------- REQUESTS  (current) -----------------
Active               :             0

----------------- TARGETS  (current) -----------------
RUNNING              :            -3

PR #7168, opened 2023-05-17
dcache-bulk: fix double counting of RUNNING targets

alrossi:
Motivation:

See GH #7164
BULK: in-memory counter is not decrementing when RELEASE activity completes

Modification:

The offending call was in storeOrUpdate. Its call to the underlying update method should not increment for RUNNING.

Removed the incrementing from the dao class and distributed it properly.

Result:

Counts for RUNNING targets are now sane.

Target: master
Request: 9.0
Patch: https://rb.dcache.org/r/13986/
Closes: #7164
Requires-notes: yes
Acked-by: Tigran


lemora:
retest this please

PR #7203, opened 2023-06-29
dcache-qos(verifier,engine): fix incompatibility issues with Subject …

alrossi:
…and attributes

Motivation:

See RT 10469 QoS change is not working.

This brought to my attention two incompatibilities which require a bit more defensive code.

1.  https://github.com/dCache/dcache-security-fixes/tree/fix/master/qos-propagate-subject-to-adjuster introduced a subject field/column in the `qos-operation` table to enforce certain kinds of
    restrictions.   However, it neglected to consider the case where a pre-existing table
    would be modified by Liquibase and the update would fill that field with `null` values,
    causing a deserialization error when refetching the metadata.

2.  The `ALRPStorageUnitQoSProvider` does not know how to handle undefined AL or RP values (it expects the file attributes always to be defined).

Modification:

1.  Just return a null for the Subject and let the operation fail if it is a QoSModify.  The operation should be removed from the database, and a subsequent retry should work.

2.  Inject the default values into the class and use those when the FileAttributes do not contain them.

Result:

Two small incompatibilities fixed.

Target: master
Request: 9.1
Request: 9.0
Request: 8.2
Patch: https://rb.dcache.org/r/13997/
Bug: #10469 (RT)
Requires-notes: yes
Requires-book: no
Acked-by: Marina

### DIFF
--- a/modules/dcache-qos/src/main/java/org/dcache/qos/services/engine/provider/ALRPStorageUnitQoSProvider.java
+++ b/modules/dcache-qos/src/main/java/org/dcache/qos/services/engine/provider/ALRPStorageUnitQoSProvider.java
@@ -159,8 +159,8 @@ public class ALRPStorageUnitQoSProvider implements QoSRequirementsProvider, Cell
         }
 
         FileAttributes attributes = descriptor.getAttributes();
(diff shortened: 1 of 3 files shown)


lemora:
retest this please

PR #7228, opened 2023-07-06
dcache-frontend: add symlink resolution to /api/v1/id GET and /api/v1…

alrossi:
…/namespace

Motivation:

If the user root contains a symlink, or
the target path contains a symlink pertaining
to the user root, the following currently
fail:

`GET, POST or DELETE /api/v1/namespace`
`GET /api/v1/id`

In the former case, we see:

```
{
  "detail": No such file or directory /pnfs/fnal.gov/usr/pnfs/fs/usr/fermilab/users/arossi/volatile/data_1b",
  "title":"Not Found",
  "status":"404"
}

```
in the latter,

```
{
  "detail": "/pnfs/fnal.gov/usr is not a prefix of /pnfs/fs/usr",
  "title": "Bad Request",
  "status": "400"
}
```

This stems from the need for a special call to
the pnfs handler to resolve symlinks on path and root, as we do, for instance, in the `xroot` door.

Modification:

So as not to affect code in other areas which seems to be working, we add two methods to `pathMapper`
which have an extra parameter for passing in
a configured `PnfsHandler`.   We use the
same message as the `xroot` door does to
retrieve resolved root and path.

Result:

The above REST calls no longer fail.

Target: master
Request: 9.1
Request: 9.0
Request: 8.2
Patch: https://rb.dcache.org/r/14013/
Requires-notes: yes
Acked-by: Tigran
Acked-by: Dmitry

### DIFF
--- a/modules/dcache-frontend/src/main/java/org/dcache/restful/resources/namespace/FileResources.java
+++ b/modules/dcache-frontend/src/main/java/org/dcache/restful/resources/namespace/FileResources.java
@@ -67,7 +67,6 @@ import org.dcache.namespace.FileType;
 import org.dcache.pinmanager.PinManagerPinMessage;
 import org.dcache.pinmanager.PinManagerUnpinMessage;
 import org.dcache.poolmanager.PoolMonitor;
-import org.dcache.qos.QoSException;
 import org.dcache.qos.QoSTransitionEngine;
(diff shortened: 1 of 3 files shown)


alrossi:
Symlink resolution message Etc.  was not backported to 8.2 (https://rb.dcache.org/r/13971).   Discarding.

PR #7232, opened 2023-07-07
gplazma-roles:  add QoS role support

alrossi:
Motivation:

We would like a way to allow non-admin users
to be authorized to request QOS_MODIFIED for
a file or set of files.

The most convenient way to do this is to
add a new role, `qos`, which does not alter
whatever user ROOT the user has from the
user's login, but which is visible to
the QoS engine and which can be checked
there.

Modfication:

Create the new role.

Result:

Necessary step to authorization of qos
modifications.

This is a new feature but needs to be
in 8.2 so we do not oblige users to
upgrade (yet) to the 9-series.

Target: master
Request: 9.1
Request: 9.0
Request: 8.2
Patch: https://rb.dcache.org/r/14017/
Requires-notes: yes
Acked-by: Tigran


alrossi:
Removing because of incompatibilities

PR #7236, opened 2023-07-13
dcache-qos:  improve exception reported on aborted verification

alrossi:
Motivation:

When the verifier aborts an operation, the
error code is always the same (10011, unexpected
system error), and the cause may not reflect
the root cause.

Modification:

Fix this so that more specific error info
is registered.

Result:

Better understanding of what went wrong.

Target: master
Request: 9.1
Request: 9.0
Request: 8.2
Patch: https://rb.dcache.org/r/14020/
Requires-notes: yes
Acked-by: Lea


svemeyer:
retest this please

Issue #7190, opened 2023-06-16
bulk recursive QoS change 


DmitryLitvintsev:
dCache version 8.2.23 

command executed to bring to tape all files in directory tree starting from `/pnfs/fnal.gov/usr/nova/persistent/users/annah1`
```
curl ...  -X POST "https://example.org:3880/api/v1/bulk-requests" -H "accept: application/json" -H "content-type: application/json" -d '{"activity":"UPDATE_QOS","expandDirectories":"ALL","arguments":{"targetQos":"tape"},"target":["/pnfs/fnal.gov/usr/nova/persistent/users/annah1"]}'
```
(Url changed on purpose) 

The request started and is currently in this state:

```
request ls f1fcfb7d-db6e-46b8-a8a1-2384ab530dd2 
SEQNO        | ARRIVED             |            MODIFIED |        OWNER |     STATUS | ID
829          | 2023/06/15-18:02:53 | 2023/06/15-18:02:53 |   44776:9553 |    STARTED | f1fcfb7d-db6e-46b8-a8a1-2384ab530dd2
```

But nothing seems to be happening since 6 PM (~ 4 hours). 

I have checked the pinboard:

```
15 Jun 2023 18:26:59 [pool-8-thread-7] [] handleTarget f1fcfb7d-db6e-46b8-a8a1-2384ab530dd2, path /pnfs/fnal.gov/usr/nova/persistent/users/annah1, error Timeout during directory listing..
15 Jun 2023 18:27:12 [bulk-766] [PnfsManager PnfsListDirectory] Received list result for an unknown request. Directory listing was possibly incomplete.
```

I look at DB: 

```
bulk=# select * from request_target where rid = 'f1fcfb7d-db6e-46b8-a8a1-2384ab530dd2' order by created_at asc ;
  id   | pid |                 rid                  |                pnfsid                |                                                                             
                                        path                                                                                                                     |  type 
  |  activity  |  state  |       created_at        |       started_at        |      last_updated       |            error_type             |           error_message     
      
-------+-----+--------------------------------------+--------------------------------------+-----------------------------------------------------------------------------
-----------------------------------------------------------------------------------------------------------------------------------------------------------------+-------
--+------------+---------+-------------------------+-------------------------+-------------------------+-----------------------------------+-----------------------------
------
[...]

PR #7239, opened 2023-07-13
dcache-qos: fix adjuster task cloning to reset state

alrossi:
Motivation:

See principally GH Issue #7190 `bulk recursive QoS change`.

The final problem solved there was the non-completion of bulk requests because the QoS Verifier would go into an infinite loop of retrying the FLUSH, since it did not see the `p2p` as failed and simply moved on to recheck the status of the file, finding no precious replica or tape location.

Modification:

The retry with a cloned object needs to reset the
state of the task to INITIALIZED.  Otherwise,
the map will see it as "DONE" (even on the
READY queue) and report it as such.  Since
the exception is not carried over on the clone,
it reports it erroneously as successful.

Result:

The adjuster no longer swallows Migration task
exceptions and the verifier will abort the
operation as it should, allowing the bulk
request to complete.

Target: master
Request: 9.1
Request: 9.0
Request: 8.2
Closes: #7190
Bug: #7190
Requires-notes: yes
Acked-by: Tigran


svemeyer:
retest this please

PR #7244, opened 2023-07-14
xrootd: throw FileNotFoundException if vomsdir doesn't exist

svemeyer:
Motivation:

If /etc/grid-security/vomsdir doesn't exist, dCache throws a stacktrace when xrootd is starting.

Modification:

Check, if vomsdir is an existing directory before setting the variable and throw a FileNotFoundException if it doesn't exist.

Result:

Error logging without a stacktrace if the directory doesn't exist.

Target: master
Request: 9.1
Request: 9.0
Request: 8.2
Require-notes: yes
Require-book: no
Acked-by: Al


svemeyer:
retest this please

PR #7264, opened 2023-08-02
pool: handle double remove of IdleStateHandler

kofemann:
Motivation:
If http client issues two GET requests then zero-copy enabled http mover will try to remove the IdleStateHandler twice and fail  with NoSuchElementException.

Modification:
Check the presence of IdleStateHandler before removal.

Result:
No failed transfers on GET+GET

Acked-by:
Target: master, 9.1, 9.0, 8.2
Require-book: no
Require-notes: yes


kofemann:
retest this please

PR #7278, opened 2023-08-03
Remove all remaining references of Guava's Files class.

### LINKED ISSUE #3808: Replace guava's com.google.common.io.Files implementation with analog from jdk7
dCache code base heavily  uses Googl's guava library. However, newer JDK releases providing similar functionality. For instance, com.google.common.io.Files can be replaced with java.nio.file.Files to reduce dependency on external library.

LukasMansour:
Motivation:
Closes #3808 

Modification:
Removes the last instances of com.google.common.io.Files and replaces most of them with the JDK alternative. (Also required updating one or two guava classes, like the ```TreeTraverser```)

Result:
The remaining references of Guava's ```Files``` class are removed and some small tweaks to reduce the usage of Guava were done.

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/services/billing/text/Indexer.java
+++ b/modules/dcache/src/main/java/org/dcache/services/billing/text/Indexer.java
@@ -1,22 +1,20 @@
 package org.dcache.services.billing.text;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.io.Files.isFile;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Arrays.asList;
 import static java.util.stream.Collectors.toList;
 import static org.dcache.util.ByteUnit.KiB;
 
 import com.google.common.base.Function;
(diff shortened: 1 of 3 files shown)


modules/dcache/src/main/java/org/dcache/services/billing/text/Indexer.java
```diff
@@ -102,20 +103,18 @@ public class Indexer {
...
-            return Collections.emptyList();
-        }
-    };
+    private static final Traverser<File> SORTED_FILE_TREE_TRAVERSER = Traverser.forTree(
+          file -> {
+              // check isDirectory() just because it may be faster than listFiles() on a non-directory
+              if (file.isDirectory()) {
+                  File[] files = file.listFiles();
+                  if (files != null) {
+                      return Ordering.natural().sortedCopy(asList(files));
```
kofemann, line 112: I think this can be de-guavaed as well

kofemann (changes requested):
Thanks for looking at this!

PR #7331, opened 2023-09-06
dcache-webdav:  fix incorrect parameter value given to DcacheDirector…

alrossi:
…yResource constructor

Motivation:

https://rb.dcache.org/r/14085/

passes the incorrect parameter value in
to the `DcacheDirectoryResource` contstructor.
It should be `true` if this is not a `PROPFIND`
call.

Modification:

Use the `isFetchAllAttributes` method
instead of the variable set from the
property value.

Result:

Correct resource configuration in the
case it is not `PROPFIND`.

Target: master
Patch: https://rb.dcache.org/r/14088/
Request: 9.1
Request: 9.0
Request: 8.2
Requires-notes: no (fixes unreleased change)
Acked-by: Svenja


svemeyer:
retest this please

Issue #7280, opened 2023-08-03
Kafka TLS settings for ftp (and dcap)


paurkedal:
We need some more settings for the producers of these cells to complete our setup, since we are using TLS-enabled Kafka.  Our configuration is using `*.kafka.producer.configs` for other cells, and this works well.  However, I can see this parameter-bundle is only available for cells which are composed with the spring framework.  I am not familiar with the spring framework (or Java), so I don't know feasible it would be to set up also the ftp and dcap in the same manner.  Although we're only using 7 extra parameters, [there are many](https://kafka.apache.org/10/javadoc/org/apache/kafka/common/config/SslConfigs.html), so maybe the easiest alternative solution would be to load the properties from a separate file, instead of passing them individually as with `kafka-max-block`?

I made the following incomplete sketch, but wanted to ask if it's an acceptable approach before looking into how to complete the auxiliary files and test it.
```diff
diff --git a/modules/dcache-ftp/src/main/java/org/dcache/ftp/door/FtpDoorSettings.java b/modules/dcache-ftp/src/main/java/org/dcache/ftp/door/FtpDoorSettings.java
index 5142513e29..b484121d4a 100644
--- a/modules/dcache-ftp/src/main/java/org/dcache/ftp/door/FtpDoorSettings.java
+++ b/modules/dcache-ftp/src/main/java/org/dcache/ftp/door/FtpDoorSettings.java
@@ -4,6 +4,9 @@ import diskCacheV111.util.PnfsHandler;
 import dmg.cells.nucleus.CellAddressCore;
 import dmg.cells.nucleus.CellEndpoint;
 import dmg.cells.nucleus.CellPath;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 import org.apache.kafka.clients.producer.KafkaProducer;
@@ -14,6 +17,8 @@ import org.dcache.poolmanager.PoolManagerStub;
 import org.dcache.util.Option;
 import org.dcache.util.OptionParser;
 import org.dcache.util.PortRange;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Object holding configuration options for FTP doors.
@@ -21,6 +26,7 @@ import org.dcache.util.PortRange;
  * Settings can be injected using {@link OptionParser}.
  */
 public class FtpDoorSettings {
[...]

PR #7334, opened 2023-09-07
Support additional Kafka parameters for ftp and dcap.

paurkedal:
This is my proposal to solve #7280.  Originally I indicated that the new configuration parameters would be called `configs`, but I changed it to `config-file` due to the observable difference in how it is used.

I have only tested gsiftp, since the implementation for dcap is almost identical and I already spent most of the time getting gsiftp to work.  Though that mainly caused by a bug in globus, so I can maybe get dcap working faster if you think it's essential to also test it.  I did the test by cherry-picking the commit onto the 9.1 branch, building a deb and installing it on some experimental headnodes we're running, adding in our TLS parameters in the separate file and triggering a billing record with arccp.


kofemann:
/
ok to test

kofemann:
retest this please

Issue #4087, opened 2018-08-05
Replace guava's ListanableFuture with j.u.c.CompletableFuture


kofemann:
Googls's guava library provides ListanableFuture - a building block for async processing. Starting from java8 j.u.c.CompletableFuture is available, which covers almost all use cases of ListanableFuture.

LukasMansour:
In order to solve this and truly remove almost all ```ListenableFuture```s, we would need to change our caches from Guava to [Caffeine](https://github.com/ben-manes/caffeine). 

For more Information:
Caffeine is the recommended successor library from Guava itself to improve caching speeds.
See the official Guava Javadocs [here](https://guava.dev/releases/31.1-jre/api/docs/com/google/common/cache/CacheBuilder.html)


Caffeine's caches return ```CompletableFuture```s, which is the main reason why this is significant.
Furthermore Caffeine's caches are more performant than Guava's.

This comes with it's own set of challenges, as Guava and Caffeine share a very similar API syntax, however their internals can be a little different (mainly how Exceptions are handled in the async threads for the cache). In some cases this might lead to behavioural changes that we need to be careful of.

kofemann:
One of the projects have a nice road map for migration

https://github.com/bitcoinj/bitcoinj/issues/2254

Funny enough, they have used my notes. The word is small 😎

PR #7346, opened 2023-09-25
Use Caffeine instead of Guava for caching.

LukasMansour:
Motivation:
Work towards #4087 

Modification:
Replace's Guava's caching with Caffeine. Caffeine is the google-supported alternative to guava's caches, that importantly uses CompletableFutures instead of ListenableFutures. Guava does not plan to ever add this support and only does bug-fixes for Guava's caches.

It's important to remember, that the 'async' terminology used in Guava/Caffeine varies from the JDK's 'async' terminology. As such some issues may arise during translation from one library to the other. But for most cases, one will see that the translation is very straight-forward. 

I recommend this for the release after 9.2.

Result:
Replaces all uses of Guava's cache with Caffeine's cache and #4087 is now feasible without having to wrap ListenableFutures in CompletableFutures. With these commits It's possible to completely remove any import on ListenableFuture, this is however not done yet in this commit.

### DIFF
--- a/modules/cells/src/main/java/dmg/cells/services/CoreRoutingManager.java
+++ b/modules/cells/src/main/java/dmg/cells/services/CoreRoutingManager.java
@@ -23,9 +23,7 @@ import static java.util.stream.Collectors.toMap;
 
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Iterables;
-import com.google.common.collect.Lists;
 import com.google.common.collect.Multimap;
-import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
@@ -329,7 +327,7 @@ public class CoreRoutingManager
                           .map(CellDomainInfo::getCellDomainName)
                           .forEach(domain -> domains.put(domain, new ArrayList<>()));
                     queueRoutes.asMap().forEach(
-                          (domain, cells) -> domains.put(domain, Lists.newArrayList(cells)));
+                          (domain, cells) -> domains.put(domain, new ArrayList<>(cells)));
                 }
                 msg.revertDirection();
                 msg.setMessageObject(new GetAllDomainsReply(domains));
@@ -504,9 +502,9 @@ public class CoreRoutingManager
     public synchronized Object ac_ls_$_0(Args args) {
         return new Object[]{
               getCellDomainName(),
-              Sets.newHashSet(localConsumers.values()),
+              new HashSet<>(localConsumers.values()),
               queueRoutes.asMap().entrySet().stream().collect(
-                    toMap(Map.Entry::getKey, e -> Sets.newHashSet(e.getValue())))
+                    toMap(Map.Entry::getKey, e -> new HashSet<>(e.getValue())))
         };
     }
 }
--- a/modules/cells/src/main/java/dmg/util/logback/FilterThresholdSet.java
+++ b/modules/cells/src/main/java/dmg/util/logback/FilterThresholdSet.java
@@ -4,23 +4,24 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
 import ch.qos.logback.classic.Level;
+import com.github.benmanes.caffeine.cache.CacheLoader;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.base.Throwables;
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
 import com.google.common.collect.HashBasedTable;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
 import com.google.common.collect.Table;
+import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.ExecutionException;
+import java.util.concurrent.CompletionException;
+import java.util.function.Function;
 import org.slf4j.Logger;
 
 /**
@@ -41,7 +42,7 @@ public class FilterThresholdSet {
 
     private final FilterThresholdSet _parent;
 
-    private final Set<String> _appenders = Sets.newHashSet();
+    private final Set<String> _appenders = new HashSet<>();
 
     private final Set<LoggerName> _roots = new HashSet<>();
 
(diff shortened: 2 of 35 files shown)


kofemann:
ok to test

modules/cells/src/main/java/dmg/cells/services/CoreRoutingManager.java
```diff
@@ -329,7 +327,7 @@ public void messageArrived(CellMessage msg) {
                           .map(CellDomainInfo::getCellDomainName)
                           .forEach(domain -> domains.put(domain, new ArrayList<>()));
                     queueRoutes.asMap().forEach(
-                          (domain, cells) -> domains.put(domain, Lists.newArrayList(cells)));
+                          (domain, cells) -> domains.put(domain, new ArrayList<>(cells)));
```
kofemann, line 330: Maybe `List.copyOf()` is a better

modules/cells/src/main/java/dmg/cells/services/CoreRoutingManager.java
```diff
@@ -504,9 +502,9 @@ private Optional<CellTunnelInfo> getTunnelInfo(CellAddressCore tunnel) {
     public synchronized Object ac_ls_$_0(Args args) {
         return new Object[]{
               getCellDomainName(),
-              Sets.newHashSet(localConsumers.values()),
+              new HashSet<>(localConsumers.values()),
```
kofemann, line 505: same here, `Set.copyOf`

modules/cells/src/main/java/dmg/util/logback/FilterThresholdSet.java
```diff
@@ -93,7 +93,7 @@ public synchronized void addAppender(String name) {
      */
     public synchronized Collection<String> getAppenders() {
         if (_parent == null) {
-            return Lists.newArrayList(_appenders);
+            return new ArrayList<>(_appenders);
```
kofemann, line 96: and so on :)

modules/chimera/src/main/java/org/dcache/chimera/JdbcFs.java
```diff
@@ -725,36 +719,40 @@ public FsInode path2inode(String path, FsInode startFrom) throws ChimeraFsExcept
     @Override
     public String inode2id(FsInode inode) throws ChimeraFsException {
         try {
-            return _idCache.get(inode.ino(), () -> {
+            return _idCache.get(inode.ino(), (key) -> {
                 String id = _sqlDriver.getId(inode);
                 if (id == null) {
-                    throw FileNotFoundChimeraFsException.of(inode);
+                    throw new RuntimeException(FileNotFoundChimeraFsException.of(inode));
```
kofemann, line 725: this looks quite dirty....

kofemann (commented):
First pass...

Lot of changes, but mostly looks ok. The flow with RuntimeException looks quite dirty to me. 

I will digest the PR and look at it a bit later again.

modules/chimera/src/main/java/org/dcache/chimera/JdbcFs.java
LukasMansour: The issue if I recall, is that you can't throw an exception in a CompletableFuture like you can in a ListenableFuture. Perhaps we can find a nicer way to propagate the FileNotFoundException upwards , I am almost sure, there is a nicer way to solve this issue, I just have not found it yet.

modules/cells/src/main/java/dmg/cells/services/CoreRoutingManager.java
LukasMansour: It might be, however List.copyOf() will return an immutable List. I will see if this is adequate in this case, if we have any modification then this is of course not adequate.

modules/chimera/src/main/java/org/dcache/chimera/JdbcFs.java
LukasMansour: This should be the case now.

modules/cells/src/main/java/dmg/util/logback/FilterThresholdSet.java
LukasMansour: Not doable in this case. As they cascade in the children and add elements as they do so.

modules/chimera/src/main/java/org/dcache/chimera/JdbcFs.java
```diff
@@ -725,36 +719,40 @@ public FsInode path2inode(String path, FsInode startFrom) throws ChimeraFsExcept
...
                 }
                 return id;
             });
-        } catch (ExecutionException e) {
-            Throwables.throwIfInstanceOf(e.getCause(), ChimeraFsException.class);
-            Throwables.throwIfInstanceOf(e.getCause(), DataAccessException.class);
-            Throwables.throwIfUnchecked(e.getCause());
-            throw new RuntimeException(e.getCause());
+        } catch (CompletionException e) {
+            if (e.getCause() != null) {
```
kofemann, line 730: I think the check is unnecessary. CompletionException should always have a cause.

modules/chimera/src/main/java/org/dcache/chimera/JdbcFs.java
```diff
@@ -725,36 +719,40 @@ public FsInode path2inode(String path, FsInode startFrom) throws ChimeraFsExcept
...
+                        throw new CompletionException(FileNotFoundChimeraFsException.ofPnfsId(id));
                     }
                     return ino;
                 }));
-            } catch (ExecutionException e) {
-                Throwables.throwIfInstanceOf(e.getCause(), ChimeraFsException.class);
-                Throwables.throwIfInstanceOf(e.getCause(), DataAccessException.class);
-                Throwables.throwIfUnchecked(e.getCause());
+            } catch (CompletionException e) {
+                if (e.getCause() != null) {
```
kofemann, line 751: ditto

modules/dcache-nfs/src/main/java/org/dcache/chimera/nfsv41/door/proxy/NfsProxyIoFactory.java
```diff
@@ -78,29 +78,36 @@ public ProxyIoAdapter getOrCreateProxy(Inode inode, stateid4 stateid, CompoundCo
...
+                  (key) -> {
+                try {
+                    final NFS4Client nfsClient;
+                    if (context.getMinorversion() == 1) {
+                        nfsClient = context.getSession().getClient();
+                    } else {
+                        nfsClient = context.getStateHandler().getClientIdByStateId(stateid);
+                    }
+
+                    final NFS4State state = nfsClient.state(stateid);
```
kofemann, line 90: the sateid in this context is the `key`, isn't it?

modules/dcache-srm/src/main/java/diskCacheV111/srm/dcache/CanonicalizingByteArrayStore.java
```diff
@@ -217,8 +218,8 @@ public byte[] readBytes(Token token) {
      */
     private Token makeToken(long id) {
         try {
-            return canonicalizationCache.get(id, () -> new Token(id));
-        } catch (UncheckedExecutionException | ExecutionException e) {
+            return canonicalizationCache.get(id, (key) -> new Token(id));
```
kofemann, line 221: same here (if I correctly read caffeine API)

kofemann (changes requested):
pretty close :)

modules/dcache-nfs/src/main/java/org/dcache/chimera/nfsv41/door/proxy/NfsProxyIoFactory.java
LukasMansour: Yes it is, however I do not see any significant advantage (one can argue performance benefits for either or) of using the lambda parameter or the parameter from the higher scope. The idea is that you could use a getter or a return method in the statement and don't need to store it another variable

Advantage in the following case:
```java
cache.get(someinstance.getStringIDontWantToStoreInAVariable(), (string) -> "defaultValue" + string)
```

No real advantage in the following case (our case):
```java
public getFromCacheOrDefault(String key) {
    cache.get(key, (key_) -> "defaultValue" + key_) // I could just as easily use key and not key_ here.
}
```

Maybe I am mistaken, but I thought in this case it would be easier to just wait for JDK 21 where we have unnamed variables ("_"), hence I just called it key.

kofemann:
retest this please

kofemann:
@LukasMansour could you please rebase to the current master

PR #7370, opened 2023-10-02
dcache-bulk: container rewrite to optimize threading

alrossi:
Motivation:

Benchmarking Bulk PIN requests against SRM's
bringonline requests revealed that the current
bulk threading implementation was nearly an order
of magnitude slower (e.g., 10K target paths in
10 minutes rather than 75 seconds).

Investigation determined there were two main issues:

Bulk was executing preparatory calls to the
PnfsManager synchronously on the same thread
as that allocated for the main activity operation. Thread pools were unnecessarily multiple,
and all of the BoundedCachedExecutor variety,
which does not allow for unbounded pool expansion
(unlike SRM, which uses the unbounded cached thread pool).

The second point actually makes a huge difference in performance.

All of this suggested that the container job needed significant rewriting to get the performance SRM does.

Modification:

The container job classes have been collapsed into a single concrete implementation class. This has
been refactored to use ContainerTask objects.

All jobs are given to the unbounded cached pool
  for execution, but they must acquire a semphore
  permit before running, releasing it when they have
  finished.

PnfsManager communication is now asynchronous.
  Callbacks are handled by the calling thread.
  The callback from the activity itself is resubmitted
  to the executor pool since it involves a bit more work,
  and always a database update.

Activity semaphores and executors have been removed, along with their related properties.

There is one minor internal interface change (parameter type) on the activity perform method.

The properties and documentation have been updated. The defaults are set from the testing reported here, though further tweaking may be desirable after thorough stress-testing has been done.

Result:

See summary under testing.  Performance is now very good.
  Code is also cleaner (IMO).

Target: master
Request: 9.2
Patch: https://rb.dcache.org/r/14115
Requires-book: yes (changes included)
Requires-notes: yes
Acked-by: Tigran


alrossi:
**_Please note that when applying this patch, there are two conflicts which will need to be resolved against the HA pull request._**

alrossi:
Conflicts resolved

PR #7453, opened 2023-12-05
dcache-xroot:  check that descriptor is not null before calling close

alrossi:
Motivation:

```
05 Dec 2023 16:29:55 (dcache-cms199-01) [] An exception java.lang.NullPointerException
at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
at org.dcache.xrootd.pool.XrootdPoolRequestHandler.removeAllDescriptorsAtomically(XrootdPoolRequestHandler.java:948)
at org.dcache.xrootd.pool.XrootdPoolRequestHandler.exceptionCaught(XrootdPoolRequestHandler.java:277)
```

as reported on Slack channel.

Modification:

Should check for `null` before closing descriptor.

Result:

The NPE does not swallow the actual exception from `exceptionCaught`.

Target: master
Request: 9.2
Request: 9.1
Request: 9.0
Patch: https://rb.dcache.org/r/14180/
Requires-notes: yes
Acked-by: Lea

### DIFF
--- a/modules/dcache-xrootd/src/main/java/org/dcache/xrootd/pool/XrootdPoolRequestHandler.java
+++ b/modules/dcache-xrootd/src/main/java/org/dcache/xrootd/pool/XrootdPoolRequestHandler.java
@@ -18,6 +18,7 @@
 package org.dcache.xrootd.pool;
 
 import static java.nio.charset.StandardCharsets.US_ASCII;
+import static java.util.function.Predicate.not;
 import static org.dcache.xrootd.protocol.XrootdProtocol.UUID_PREFIX;
 import static org.dcache.xrootd.protocol.XrootdProtocol.kXR_ArgInvalid;
 import static org.dcache.xrootd.protocol.XrootdProtocol.kXR_ArgMissing;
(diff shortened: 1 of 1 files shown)


kofemann:
I think the fix is wrong. See the comments to https://rb.dcache.org/r/14180/

modules/dcache-xrootd/src/main/java/org/dcache/xrootd/pool/XrootdPoolRequestHandler.java
```diff
@@ -945,7 +946,7 @@ private FileDescriptor getDescriptorAtomically(int fd) throws XrootdException {
 
     @GuardedBy("writeLock")
     private void removeAllDescriptorsAtomically() {
-        _descriptors.forEach(FileDescriptor::close);
+        _descriptors.stream().filter(not(null)).forEach(FileDescriptor::close);
```
kofemann, line 949: Should be 'Objekts::notNull'

kofemann (changes requested):

PR #7504, opened 2024-01-30
pool: mover grizzly IO buffer initialization into NfsTransferService

kofemann:
Motivation:
the grizzly IO buffers are pre-initialized on startup. However, the initialization is bound to EDSOperationREAD class, thus, effectively, during the first NFS read request.

Modification:
mover grizzly IO buffer initialization into NfsTransferService, which is initialized at pool stat.

Result:
More predicable buffer initialization

Acked-by: Lea Morschel
Target: master
Require-book: no
Require-notes: yes
(cherry picked from commit 426d7c4f7a721c61376b09c9adab4d1f75cef2c3)


svemeyer:
retest this please

Issue #6298, opened 2021-11-24
There is no kafka door event for HTTP-TPC transfers


paulmillar:
Other doors generate a kafka event on transfers; however, the HTTP-TPC transfers do not generate such events.

Note that, currently, RemoteTransferManager sends the billing record for HTTP-TPC rather than the door.

paurkedal:
(Force-pushing tabs-to-spaces expansion.)

PR #7545, opened 2024-04-09
Add Kafka producer to transfer manager.

paurkedal:
This is a proposed solution to #6298.

### DIFF
--- a/modules/dcache/src/main/java/diskCacheV111/services/TransferManager.java
+++ b/modules/dcache/src/main/java/diskCacheV111/services/TransferManager.java
@@ -5,6 +5,7 @@ import static java.util.stream.Collectors.joining;
 import diskCacheV111.util.CacheException;
 import diskCacheV111.util.MissingResourceCacheException;
 import diskCacheV111.util.PnfsId;
+import diskCacheV111.vehicles.DoorRequestInfoMessage;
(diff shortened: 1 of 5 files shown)


mksahakyan:
retest this please

kofemann:
retest this please

kofemann:
Thanks!

paurkedal:
And thanks for reviewing!  With this and #7546, we're close to having the full billing logs for our usage in Kafka according to empirical comparison.  (The remaining missing records are from NFS, which I couldn't quite locate where to implement, but it's only internal traffic for us, anyway.)

paurkedal:
After reproducing the billing entries in our test setup, I can confirm that "request" records from NFS cells are now present also in Kafka.  My original comparison was done on our production setup, where we are running dCache 9.2.11.  I have a suspicion that "remove" records like the following are missing, still, but it could be my dCache build is too old:
```
05.02 15:45:23 [door:NFS-headnode1-x@nfs-headnode1-xDomain:remove] ["":0:0:127.0.0.1] [000042EB85512F314B498516FF443E5396CC,0] [parent:[0000A1260353F71D44819567ED4C6AD028F2]/test.txt] <Unknown> 0 0 {0:""}
```

paurkedal:
I retried the test from my previous comment with a build from today's master branch, and I can confirm that "remove" records from NFS doors are missing.

Issue #7553, opened 2024-04-18
`oidc` plugin - authentication against the OP


calestyo:
Hey.

Took me quite some time today, to find out why oidc authentication failed.

In a line like:
```
gplazma.oidc.provider!atlas = https://atlas-auth.web.cern.ch/ …
```
there needs to be some way how to actually authenticate against the OP.

What people in dCache would probably expect is that `/etc/grid-security/` CAs are used for that, which would IMO be the natural and best choice.<br>
However, undocumented as it is (or at least I couldn’t find anything about it), it uses the system JKS instead.

And the error message is pretty much encrypted:
```
Unknown kid "rsa2"
```

And even with `DEBUG` it doesn’t get much better:
```
2024-04-18T19:46:48.418110+02:00 lcg-lrz-dc50 dcache@gplazma1[742038]: 2024-04-18 19:46:48+02:00 (gPlazma) [webdav.tls_lcg-lrz-dc35 Login AUTH oidc] Failed to fetch discovery document for atlas: javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
2024-04-18T19:46:48.435721+02:00 lcg-lrz-dc50 dcache@gplazma1[742038]: 2024-04-18 19:46:48+02:00 (gPlazma) [webdav.tls_lcg-lrz-dc35 Login] Login attempt failed; detailed explanation follows:
2024-04-18T19:46:48.435958+02:00 lcg-lrz-dc50 dcache@gplazma1[742038]: LOGIN FAIL
2024-04-18T19:46:48.436091+02:00 lcg-lrz-dc50 dcache@gplazma1[742038]:  |    in: Origin[2003:cd:df19:3800:8f7e:27cc:ac69:49fc]
2024-04-18T19:46:48.436262+02:00 lcg-lrz-dc50 dcache@gplazma1[742038]:  |        JWT bearer token:
2024-04-18T19:46:48.436449+02:00 lcg-lrz-dc50 dcache@gplazma1[742038]:  |          |
2024-04-18T19:46:48.436617+02:00 lcg-lrz-dc50 dcache@gplazma1[742038]:  |          +- iss: "https://atlas-auth.web.cern.ch/"
[...]

PR #7554, opened 2024-04-21
docs: update oidc chapter to explain trust anchors

paulmillar:
Motivation:

Issue #7553 describes how it's currently undocumented that the OIDC plugin uses Java's built-in trust store.

Modification:

Document behaviour

Result:

Admins may have a better understanding of how to configure their dCache.

Target: master
Requires-notes: no
Requires-book: yes
Request: 9.2


svemeyer:
Shouldn't this be backported to 10.0, too?

Issue #7552, opened 2024-04-18
file mtime should not be updated when a new hard link is created


kofemann:
According to POSIX, file mtime should be updated only when file content has been changed, thus creating a hardlink must not modify it.

PR #7562, opened 2024-04-29
chimera: don't update file mtime on hard links creation

kofemann:
Motivation:
According to POSIX, file mtime should be updated only when file content has been changed, thus creating a hardlink must not modify it.

Modification:
added unit test to ensure posix behaviour. The inc/decNlink is split into two methods - one for directories, where nlink update associated with mtime update, and for files, where nlink update is independent of file modification.

Result:
improved POSIX compliance.

Fixes: #7552
Acked-by: Lea Morschel
Target: master, 10.0, 9.2
Require-book: no
Require-notes: yes
(cherry picked from commit bf4e865732f46a755219146359117f775913c406)


kofemann:
The demanding community migrated to 10.0.6

PR #7603, opened 2024-06-20
Update bulk.properties: add comment

onnozweers:
Added a reminder that if you change the bulk service name, you may need to update it in your frontend configuration as well.


kofemann:
Thanks @onnozweers !

PR #7610, opened 2024-06-25
chimera: add CTA HSM StorageInfo extractor

DmitryLitvintsev:
Motivation:
===========
When transitioning from Enstore to CTA we must define hsm type. And we can drop a lot of miscellaneous information typically collected by Enstore extractor

Modification:
=============
Add CtaStorageIndo and CtaStorageInfo Extractor

Result:
======
Can read/write CTA and migrated Enstore files without having to set hsmInstance tag in the full directory tree.

Patch: https://rb.dcache.org/r/14262
Target: trunk
Request: 10.x
Request: 9.x


mksahakyan:
retest this please

PR #7621, opened 2024-07-24
frontend: Add an option to set SRR path mappings

sfayer:
Hi,

This patch adds an option to populate the SRR path field, as recently discussed on the user-forum list.

I've targeted it against the master branch but I'd quite like to backport it to v9 if at all possible (it's a tiny optional feature, so I hope that isn't a problem), although I don't know how you handle backports?

Please let me know if I need to do anything further...

Regards,
Simon


kofemann:
ok to test

PR #7627, opened 2024-07-26
Add gplazma2-jython module for auth using Python

ThePhisch:
Adds a new gplazma2 module to the project which uses Jython to parse and execute Python files (only version 2.7 supported) within the dcache Java runtime. This enables site admins to quickly script authentication plugins without needing to understand or write the native dcache Java code.

This commit includes a sample "sample_plugin.py" auth plugin which is used during testing and serves as a reference.

### Implementation Details

Jython limits the Python version to version 2.7. The Python files are currently stored directly within the source
folder, i.e.
`modules/gplazma2-pyscript/gplazma2-pyscript/sample_plugin.py`. Multiple files can be inserted which will be executed
sequentially but in no particular order.

Each Python file shall be structured to contain one function `py_auth(public_credentials, private_credentials,
principals)` which returns a boolean whether the authentication was successful. Returning `False` will trigger an
`AuthenticationException` in Java. Each parameter passed to `py_auth` will be a Python set.

- `private_credentials` is a set of strings, integers or floats (implemented in Java as `String || Number`
- `public_credentials` is the same as `private_credentials`
- `principals` will be a set of strings in the form `<principal_type>:<value>`, i.e. a pair of strings separated by a
  colon. Consider `org.dcache.auth.Subjects#principalsFromArgs` to see how the principal type string maps to the
  principals classes.

The conversion from a set of principals to a set of strings is done entirely within the `authenticate` method of the
`PyscriptPlugin`-class. The Python file only ever sees a set of strings, while someone using the plugin will only ever
need to pass a set of principals from within Java. The conversion is done using existing functions.

### Tests

A test class `PyscriptPluginTest` has been added that makes use of the internal logic of the sample `sample_plugin.py`. The logic is simple and is repeated in the Java class and the Python file as a comment.

Signed-off-by: Anton Schwarz anton.schwarz@desy.de

### DIFF
--- a/modules/gplazma2-jython/pom.xml
+++ b/modules/gplazma2-jython/pom.xml
@@ -1,0 +1,40 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.dcache</groupId>
+    <artifactId>dcache-parent</artifactId>
+    <version>10.1.0-SNAPSHOT</version>
+    <relativePath>../../pom.xml</relativePath>
+  </parent>
+
+  <artifactId>gplazma2-jython</artifactId>
+  <packaging>jar</packaging>
+
+  <name>gPlazma 2 Jython Generic plugin</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+    </dependency>
+      <dependency>
+          <groupId>org.python</groupId>
+          <artifactId>jython-standalone</artifactId>
+      </dependency>
+    <dependency>
+        <groupId>org.dcache</groupId>
+        <artifactId>dcache-common</artifactId>
+        <version>${project.version}</version>
+    </dependency>
+    <dependency>
+        <groupId>org.dcache</groupId>
+        <artifactId>gplazma2</artifactId>
+        <version>${project.version}</version>
+    </dependency>
+  </dependencies>
+</project>
--- a/modules/gplazma2-jython/src/main/java/org/dcache/gplazma/jython/JythonPlugin.java
+++ b/modules/gplazma2-jython/src/main/java/org/dcache/gplazma/jython/JythonPlugin.java
@@ -1,0 +1,175 @@
+package org.dcache.gplazma.jython;
+
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.security.Principal;
+import java.util.Set;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Properties;
+import java.util.stream.Collectors;
+
+import org.dcache.auth.Subjects;
+import org.dcache.gplazma.AuthenticationException;
+import org.dcache.gplazma.plugins.GPlazmaAuthenticationPlugin;
+import org.python.core.PyFunction;
+import org.python.core.PyObject;
+import org.python.core.PySet;
+import org.python.util.PythonInterpreter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class JythonPlugin implements GPlazmaAuthenticationPlugin {
+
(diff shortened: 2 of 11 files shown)


kofemann:
ok to test

modules/gplazma2-jython/pom.xml
```diff
@@ -0,0 +1,40 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.dcache</groupId>
+    <artifactId>dcache-parent</artifactId>
+    <version>10.1.0-SNAPSHOT</version>
```
kofemann, line 7: The master is already `10.2.0-SNAPSHOT`

modules/gplazma2-jython/src/main/java/org/dcache/gplazma/jython/JythonPlugin.java
```diff
@@ -0,0 +1,175 @@
...
+            LOGGER.error("Error reading Jython directory");
+        }
+    }
+
+    private static List<Path> listDirectoryContents(String directoryPath) throws IOException {
+        List<Path> elements = new ArrayList<>();
+        Path path = Paths.get(directoryPath);
+
+        // Check if the path is a directory
+        if (Files.isDirectory(path)) {
```
kofemann, line 48: `List<Path> elements = Files.list(Path.of(directoryPath)).toList();`

modules/gplazma2-jython/src/main/java/org/dcache/gplazma/jython/JythonPlugin.java
```diff
@@ -0,0 +1,175 @@
...
+    ) throws AuthenticationException {
+
+        for (Path pluginPath : pluginPathList) {
+            // for each path in pluginPathList
+            try {
+                // ===========
+                // IO
+                // ===========
+                // read file into string
+                LOGGER.debug("gplazma2-jython running " + pluginPath.toString());
```
kofemann, line 75: Please use placeholders, like:

```
LOGGER.debug("gplazma2-jython running {}", pluginPath);
```

modules/gplazma2-jython/src/main/java/org/dcache/gplazma/jython/JythonPlugin.java
```diff
@@ -0,0 +1,175 @@
...
+                String content = Files.lines(
+                        Paths.get(pluginPath.toString())
```
kofemann, line 77: `pluginPath` is already `Path`. Can't we use it directly: `Files.lines(pluginPath)`?

modules/gplazma2-jython/src/main/java/org/dcache/gplazma/jython/JythonPlugin.java
```diff
@@ -0,0 +1,175 @@
...
+                // Only handle select datatypes
+                for (Object pubCred : publicCredentials) {
+                    if (
+                            pubCred instanceof String
+                            || pubCred instanceof Number // all number datatypes
+                    ) {
+                        pyPublicCredentials.add(pubCred);
+                    } else {
+                        LOGGER.warn(
+                            "GPlazma2-Jython: Public Credential "
```
kofemann, line 101: The same here, please use placeholders. That way, string concatenation will run only when debug logging is enabled.

modules/gplazma2-jython/src/main/java/org/dcache/gplazma/jython/JythonPlugin.java
```diff
@@ -0,0 +1,175 @@
...
+                if (!AuthenticationPassed) {
+                    throw new AuthenticationException(
+                        "Authentication failed in file " + pluginPath.getFileName().toString()
+                    );
+                }
+                // ================
+                // update principals
+                // =================
+                // Convert principals back from string representation
+                List<String> convertedPyPrincipals = pyPrincipals.stream().toList();
```
kofemann, line 158: I think `List.copyOf(pyPrincipals)` is a better option.

modules/gplazma2-jython/src/main/resources/META-INF/gplazma-plugins.xml
```diff
@@ -0,0 +1,6 @@
+<plugins>
+  <plugin>
+    <name>Jython</name>
```
kofemann, line 3: Please make plugin name lowercase to match other plugins. Maybe `pyscript` (or similar) is a better name.  `Jython` tells too little to dCache admins.

kofemann (changes requested):
Thanks for the nice work!

ThePhisch:
Completed requested changes, will add documentation and configuration in two separate PRs.

kofemann (approved):
Very nice, thanks!
One little cosmetic change. May be the wou move `gplazma2-pyscript/sample_plugin.py` into `src/main/demo` directory

PR #7639, opened 2024-08-13
GPlazma2-Pyscript: Add mapping functionality

ThePhisch:
The Gplazma2-pyscript module now also implements the GPlazmaMappingPlugin interface.

Python plugins written for pyscript are stored within the folder specified by `gplazma.pyscript.workdir`. They are read in no particular order. Each plugin file MUST contain a `py_auth` function and a `py_map` function, which implement auth and mapping functionality. Their function signatures follow those of their Java counterparts. The Python functions are read and executed at the appropriate step in the Gplazma process (i.e. either map or auth).

This PR comes with a very short example plugin for `py_map`.

Signed off by: Anton Schwarz (anton.schwarz@desy.de)

### DIFF
--- a/modules/gplazma2-pyscript/src/main/java/org/dcache/gplazma/pyscript/PyscriptPlugin.java
+++ b/modules/gplazma2-pyscript/src/main/java/org/dcache/gplazma/pyscript/PyscriptPlugin.java
@@ -1,10 +1,8 @@
 package org.dcache.gplazma.pyscript;
 
 import java.io.IOException;
-import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.security.Principal;
 import java.util.*;
 import java.util.stream.Collectors;
@@ -12,6 +10,7 @@ import java.util.stream.Collectors;
 import org.dcache.auth.Subjects;
 import org.dcache.gplazma.AuthenticationException;
 import org.dcache.gplazma.plugins.GPlazmaAuthenticationPlugin;
+import org.dcache.gplazma.plugins.GPlazmaMappingPlugin;
 import org.python.core.PyFunction;
 import org.python.core.PyObject;
 import org.python.core.PySet;
@@ -20,7 +19,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 
-public class PyscriptPlugin implements GPlazmaAuthenticationPlugin {
+public class PyscriptPlugin implements GPlazmaAuthenticationPlugin, GPlazmaMappingPlugin {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(PyscriptPlugin.class);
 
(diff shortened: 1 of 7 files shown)


modules/gplazma2-pyscript/src/test/java/org/dcache/gplazma/pyscript/PyscriptPluginTest.java
```diff
@@ -6,12 +6,15 @@
...
 import java.util.*;
 
+import org.dcache.auth.FullNamePrincipal;
 import org.dcache.auth.PasswordCredential;
 import org.dcache.auth.UserNamePrincipal;
 import org.dcache.gplazma.AuthenticationException;
 import org.junit.Before;
 import org.junit.Test;
 
+import javax.print.attribute.standard.PrinterInfo;
```
kofemann, line 16: Is it used?

modules/gplazma2-pyscript/src/main/java/org/dcache/gplazma/pyscript/PyscriptPlugin.java
```diff
@@ -42,6 +43,35 @@ private static List<Path> listDirectoryContents(String directoryPath) throws IOE
...
+                Collectors.joining("\n")
+        );
+        // execute file content, loading the function "py_auth" into the namespace
+        PI.exec(content);
+    }
+
+    private static List<String> principalSetToStringList(Set<Principal> principals) {
+        if (principals.isEmpty()) {
+            // empty set handling
+            return new ArrayList<String>();
```
kofemann, line 57: better to use `Collections.emptyList()`

modules/gplazma2-pyscript/src/main/java/org/dcache/gplazma/pyscript/PyscriptPlugin.java
```diff
@@ -42,6 +43,35 @@ private static List<Path> listDirectoryContents(String directoryPath) throws IOE
...
+        } else {
+            String principalsString = Subjects.toString(
+                    Subjects.ofPrincipals(principals)
+            );
+            String[] principalsArray = principalsString.substring(
```
kofemann, line 62: This looks quite creative :). I suggest adding a special method in the `Subjects` class that does this (with a bit of copy-paste from existing `Subjects#toString`.

kofemann (changes requested):
Thanks, looks very nice!

modules/gplazma2-pyscript/src/main/java/org/dcache/gplazma/pyscript/PyscriptPlugin.java
ThePhisch: because it returns an immutable empty list that is memory efficient?

kofemann: Right, it returns a singleton, thus, doesn't allocate new memory when an empty list is required.

kofemann (approved):
🎉

kofemann:
I have adjusted a bit the commit message :)

PR #7753, opened 2025-02-19
pool:fix outofspace error sending pool into diasbled mode

mksahakyan:
Motivation

we are seeing the this error wen there is no space on the pool left, and pool goes into disabled mode,

AB56598C55] WRITE failed : IOError
AB56598C55] Transfer failed due to a disk error: CacheException(rc=204;msg=Disk I/O Error )
AB56598C55] Pool mode changed to disabled(fetch,store,stage,p2p-client,p2p-server): Pool disabled: Disk I/O Error
AB56598C55] Pool: dcache-xfel487-01, fault occurred in transfer: Disk I/O Error . Pool disabled: , cause: CacheException(rc=204;msg=Disk I/O Error )

however we want it to go into READ-ONLY mode and files could stay accesable.

the privous patch (https://rb.dcache.org/r/14357/diff/3/#index_header) did not fix the issue.

so the new changes are still a try to fix the issue.

Acked-by: Dmitry Litvintsev
Target: master, 10.2
Require-book: no
Require-notes: yes
Patch: https://rb.dcache.org/r/14368/

### DIFF
--- a/modules/dcache-dcap/src/main/java/org/dcache/pool/movers/DCapProtocol_3_nio.java
+++ b/modules/dcache-dcap/src/main/java/org/dcache/pool/movers/DCapProtocol_3_nio.java
@@ -367,7 +367,12 @@ public class DCapProtocol_3_nio implements MoverProtocol, ChecksumMover, CellArg
                                 String errmsg = "WRITE failed : " + (ioException == null ? "IOError"
                                       : Exceptions.messageOrClassName(ioException));
                                 int rc;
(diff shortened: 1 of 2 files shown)


khys95 (commented):
Some can be turned into comments (so just remove the TODO?)

khys95 (commented):
Update: this PR to 10.2 will be merged as is (since it's already on master)

khys95 (approved):

PR #7759, opened 2025-03-19
Update gplazma.properties: explain gplazma.oidc.provider options

onnozweers:
In the Book version 10.2, I found these options:

`-profile`
`-prefix`
`-authz-id`
`-non-authz-id`

I though it best to copy some of that info to the gplazma.properties, also since the 9.2 version of the Book does not have this info yet.

Signed-off-by: Onno Zweers (for what it's worth: I did not do much more than copying from the Book to the properties file 😸)


lemora:
Makes sense. Thanks, Onno!

PR #7772, opened 2025-04-04
10.1: docs: change openjdk 11 to 17

onnozweers:
And use a more recent dCache version in the version example


kofemann:
Thanks, @onnozweers !

PR #7775, opened 2025-04-04
Scitags: fallback mechanism

marian-babik:
This brings back the fallback mechanism to generate fireflies based on subject's virtual organisation even if there are no transfer tags indicated in the protocols. It introduces a default mapping via configuration variable pool.firefly.vo-mapping

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/pool/movers/TransferLifeCycle.java
+++ b/modules/dcache/src/main/java/org/dcache/pool/movers/TransferLifeCycle.java
@@ -236,7 +236,7 @@ public class TransferLifeCycle {
      * @return the experiment ID, or -1 if it cannot be determined
      */
     private int getExperimentId(ProtocolInfo protocolInfo, Subject subject) {
-        if (!protocolInfo.getTransferTag().isEmpty()) {
+        if (protocolInfo.getTransferTag() != null && !protocolInfo.getTransferTag().isEmpty()) {
             try {
                 int transferTag = Integer.parseInt(protocolInfo.getTransferTag());
                 if (transferTag <= 64 || transferTag >= 65536) {
--- a/modules/dcache/src/main/java/org/dcache/pool/movers/TransferLifeCycle.java
+++ b/modules/dcache/src/main/java/org/dcache/pool/movers/TransferLifeCycle.java
@@ -35,6 +35,7 @@ import java.time.Instant;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.Map;
+import java.util.OptionalInt;
 import java.util.HashMap;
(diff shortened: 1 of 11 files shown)


modules/dcache/src/main/java/org/dcache/pool/movers/TransferLifeCycle.java
```diff
@@ -204,9 +227,36 @@ private String getApplication(ProtocolInfo protocolInfo) {
...
+     * @param subject the Subject representing the user or entity associated with the transfer
+     * @return the experiment ID, or -1 if it cannot be determined
+     */
+    private int getExperimentId(ProtocolInfo protocolInfo, Subject subject) {
+        if (protocolInfo.getTransferTag() != null && !protocolInfo.getTransferTag().isEmpty()) {
+            try {
+                int transferTag = Integer.parseInt(protocolInfo.getTransferTag());
+                if (transferTag <= 64 || transferTag >= 65536) {
+                    LOGGER.warn("Invalid integer range for transfer tag: {}", protocolInfo.getTransferTag());
+                    return -1;
```
kofemann, line 244: As mentioned above, you can return `OptionalInt.empty()`

modules/dcache/src/main/java/org/dcache/pool/movers/TransferLifeCycle.java
```diff
@@ -204,9 +227,36 @@ private String getApplication(ProtocolInfo protocolInfo) {
...
+                }
+                // scitag = exp_id << 6 | act_id
+                return transferTag >> 6;
```
kofemann, line 247: and here `OptionalInt.of(transferTag >> 6);`

modules/dcache/src/main/java/org/dcache/pool/movers/TransferLifeCycle.java
```diff
@@ -82,9 +87,14 @@ public void onStart(InetSocketAddress src, InetSocketAddress dst, ProtocolInfo p
             return;
         }
 
+        int expId = getExperimentId(protocolInfo, subject);
```
kofemann, line 90: One possibility to avoid magic numbers, like `-1`, is to return `OptionalInt` and then check with `isPresent()`. It doesn't make code faster, but the intent is clear.

kofemann (approved):
Thanks, @marian-babik. The change is good as-is. You can consider the suggested improvements. Otherwise, we are happy to merge it.

marian-babik:
@kofemann Thanks for review, I have done the changes and re-tested.

kofemann (approved):

kofemann:
Thanks, @marian-babik

PR #7778, opened 2025-04-14
Scitags: Firefly usage and stats

marian-babik:
This introduces usage in the fireflies, which we need to estimate direction of the transfer. It is devised from the storage statistics in the MoverInfoMessage. Optionally, based on configuration, it also adds storage statistics to the fireflies.


kofemann:
Thanks, @marian-babik !

PR #7785, opened 2025-04-29
Support EC Signed OIDC Jwt tokens

HarryKodden:
dCache gplazma2 OIDC was only supporting RSA signed JWT tokens. I had a OIDC OP that issues EC signed tokens.
I added the support for EC Signed Tokens and now my tokens are accepted.

### DIFF
--- a/modules/gplazma2-oidc/src/main/java/org/dcache/gplazma/oidc/jwt/Issuer.java
+++ b/modules/gplazma2-oidc/src/main/java/org/dcache/gplazma/oidc/jwt/Issuer.java
@@ -31,8 +31,14 @@ import java.security.KeyFactory;
 import java.security.PublicKey;
 import java.security.spec.KeySpec;
 import java.security.spec.RSAPublicKeySpec;
-import java.time.Duration;
+import java.security.interfaces.ECPublicKey;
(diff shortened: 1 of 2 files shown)


paulmillar (changes requested):
Thanks for this contribution.

Overall, the patch seems quite reasonable.

There are also some places where (in my view) the code's readability could be improved, but I believe the code is correct.  I've marked these comments "[non-blocking]"

Unfortunately, the method `transcodeJWTECDSASignatureToDER` seems to miss some of the intricacies of DER encoding, particularly in how LENGTH and INTEGER values are encoded.  I've tried to indicate these potential issues.  Given the complexity of DER-encoding, providing some unit-tests that target this specific method would seem prudent.

it would be helpful if there were also unit-tests that target the overall processing of EC-signed JWTs (see existing unit-tests for possible inspiration).  Including some real-life example tokens as unit-tests could be useful, if possible.  The existing RSA tests build the JWT dynamically, but doing this for EC-signed JWTs might be awkward.

kofemann:
It may be off-topic: Make it sense to look at https://connect2id.com/products/nimbus-jose-jwt and avoid custom JWT code?

PR #7789, opened 2025-05-21
cleaner: allow multiple pools per hsm to be cleaned in parallel

lemora:
Motivation:
Users requested that `cleaner-hsm` support deletion via multiple pools in parallel per HSM, similar to the behavior already implemented in `cleaner-disk`. The current single-pool-per-HSM limit created bottlenecks for sites with many files and multiple eligible pools. Currently, `cleaner-hsm` will wait for one active delete request per hsm on a single pool to finish or time out before requesting more deletions from that HSM.

Modification:
Refactored HsmCleaner to track active deletes per pool instead of per HSM to be able to properly track the deletion state and retry if necessary. Introduced handling of delete batches across multiple pools for the same HSM triggered in successive `cleaner-hsm` runs. It will now handle asynchronous pool responses and timeouts per pool instead of per HSM. Added unit tests for HsmCleaner.

Result:
Cleaner-hsm can now delete files from the same HSM in parallel via multiple pools. Instead of waiting for one delete request batch per HSM to finish or time out, it will send further delete requests for that HSM to other pools conneced to it on successive `cleaner-hsm` runs. By decreasing the time between successive cleaner runs, the number of parallel cleaning pools is increased (as long as there are available pools).

Target: master
Requires-notes: yes
Requires-book: yes

### DIFF
--- a/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/HsmCleaner.java
+++ b/modules/dcache-chimera/src/main/java/org/dcache/chimera/namespace/HsmCleaner.java
@@ -19,20 +19,22 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.sql.Timestamp;
 import java.time.Instant;
-import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


onnozweers:
Ahhh! You saw my wishlist! How wonderful that you took the effort to make this! Thank you!

lemora:
You're welcome :)

Issue #7836, opened 2025-07-02
Missing data in quota REST API response


nsc-jens:
dCache 10.2.11
Java 17

I think the REST API for quota has a bug. If I run a request for a project with  quota > 0 I get:

curl -X GET "https://view.swestore.se:3880/api/v1/quota/group/7737" -H "accept: application/json"

 [
  {
    "id": 7737,
    "type": "GROUP",
    "replica": 1214870414994,
    "replicaLimit": 214748364800
  }
]

This is as expected. But I set the quota to "0" I get an unexpected result:

[
  {
    "id": 8401,
    "type": "GROUP",
    "replica": 114046292506
  }
]

It does report the usage but it doesn't report the quota size of zero. Perhaps a mixup between "NONE" and 0 somewhere?

We can mitigate this by setting the quota  to "1" instead. We use it to implement a "read-delete-only" state for a project which is being decommissioned.

Additionally, a quota request for a project with no quota set reports:

{
  "detail": "No such quota exists",
  "title": "Not Found",
  "status": "404"
}

which I guess is technically correct but could also have been reported as somethings like this without a tough 404 error:

[
  {
    "id": 8402,
    "type": "GROUP",
  }
]

But I have no strong opinions about this.

/jens

DmitryLitvintsev:
Jens,

Json issue addressed, let's see what other dCache developers think about it (somebody put these conditions for some reason after all)  

At this point I am reluctant to change behavior on attempt to get non existing quota. 

My thinking is that all exising UIDs and GIDs need to be added to quota (w/o quota limits). So querying for UID/GID always return result unless UID/GID does not exist where 404 would be valid. It has benefit of easy monitoring of space used by users and groups.

PR #7847, opened 2025-07-07
dcache rest api: do not filter out fields with default and null values

DmitryLitvintsev:
when building json reply object. Enable get/set of OUTPUT quota

Motivation:
-----------

dCache rest api response filters out fields that have null or default (for that type) values. As the result, for instance, integer filed with 0 value is skipped (so is any filed with null value). This is not desirable

Modification:
-------------

Change custom Json ObjectMapper to retain fields with null and default values

Result:
-------

Request results returned by dCache rest api contain fields having null or default values

Ticket: https://github.com/dCache/dcache/issues/7836
Patch: https://rb.dcache.org/r/14479/
Target: trunk
Request: 11.0, 10.2, 10.1, 10.0, 9.2

Acked-by: Tigran

Require-book: no
Require-notes: yes


mksahakyan:
retest this please

PR #7880, opened 2025-08-26
Human readable sizes for Webadmin Interface

mcmilk:
Motivation
----------

During the last dcache hackathon meeting, there was the idea to display the rather large byte values of the web interface in a human-readable way.

Already during the meeting there was a first short version of the implementation. Today I have adjusted a few small things and now ask for a review.

Modification
------------

Replace the old JavaScript code of 'httpd/static/scripts/sorting' with a better working one: https://github.com/White-Tiger/sorttable.js

The JavaScript file 'common.js' could be used for other modifications as well.

Result
------

Fix sorting of the times in 'ms' within the '/cellInfo' page. Display all sizes in human-readable format instead of MiB. The old and new views can be switched using a button in the top left corner.

Target: trunk
Request: 11.x, 10.x, 9.x
Require-book: no
Require-notes: yes


kofemann:
Hi @mcmilk. Thanks for the contribution. Before we merge, could you please update that the table headers don't say `xxx Space/MiB`, as the unit is already displayed next the the value and not always `MiB`

mcmilk:
> We won't backport the change to older branches.

Even not v10.2.x - the Latest Golden Release?

Issue #7971, opened 2025-12-11
WebDAV returns empty body for open-ended range request (bytes=0-)


mzur:
Hi everyone, I may have discovered a bug and would like to know if you can confirm this or if I'm still doing something wrong.

**Background**

I'm working on an integration between a web application and the DESY instance of dCache. The web application accesses dCache via WebDAV and displays images or videos. For faster loading, it requests macaroons, so the browser can directly load the files from dCache without knowing the original access credentials. When used as the `src` of a HTML video element, the browser handles the request and usually uses range requests to fetch parts of the video.

**The bug**

The initial request to fetch the start of the video is usually done with the `Range: bytes=0-` header because the browser does not know the length of the video yet. However, when sent open-ended like this, dCache seems to return an empty body. This is also the case with open-ended ranges from the middle of the file (e.g. `Range: bytes=12345-`). When I manually add any end to the range (e.g. `Range: bytes=0-1023`) the same request returns data in the body.

I also see this with public files, so I don't expect this to be an issue with the macaroon.

**How to reproduce**

Here is an example cURL request with output:

```
$ curl -v 'https://dcache.host:2880/path/to/video.mp4?authz=MACAROON' \
   -H 'Range: bytes=0-' \
   -o -

> GET /path/to/video.mp4?authz=MACAROON HTTP/1.1
> Host: dcache.host:2880
> Range: bytes=0-
> Connection: keep-alive
> 
< HTTP/1.1 206 Partial Content
< Date: Thu, 11 Dec 2025 08:45:46 GMT
< Server: dCache/9.2.35
< Content-Range: bytes 0-172889459/172889460
< ETag: "..."
< Content-Disposition: attachment
< Content-Length: 172889460
< 
* transfer closed with 172889460 bytes remaining to read
* Closing connection
curl: (18) transfer closed with 172889460 bytes remaining to read
```

The same request with an end in the range responds with data:

```
$ curl -v 'https://dcache.host:2880/path/to/video.mp4?authz=MACAROON' \
   -H 'Range: bytes=0-1023' \
   -o -

> GET /path/to/video.mp4?authz=MACAROON HTTP/1.1
> Host: dcache.host:2880
> Range: bytes=0-1023
> Connection: keep-alive
> 
< HTTP/1.1 206 Partial Content
< Date: Thu, 11 Dec 2025 08:44:37 GMT
< Server: dCache/9.2.35
< Content-Range: bytes 0-1023/172889460
< ETag: "..."
[...]

PR #7973, opened 2025-12-12
webdav: fix range header formatting in relay request

mzur:
An open-ended range like `bytes=0-` produced an invalid header `bytes=0-null`, resulting in no data being transferred. This patch implements support for open-ended ranges (`bytes=0-`) and suffix ranges (`bytes=-500`).

I confirmed the fix manually using the local `system-test`. I also wanted to add a test case for this but it seemed like this requires a full integration test which could be a little over the top? Could you give me a hint how best to add a test?

While testing a suffix range like `bytes=-10` (which should return the last 10 bytes) I may have found another bug. The response returns the correct data but has the headers `Content-Range: bytes 0-10/761` and `Content-Length: 11` which seem incorrect.

References https://github.com/dCache/dcache/issues/7971

### DIFF
--- a/modules/dcache-webdav/src/main/java/org/dcache/webdav/DcacheResourceFactory.java
+++ b/modules/dcache-webdav/src/main/java/org/dcache/webdav/DcacheResourceFactory.java
@@ -1799,8 +1799,19 @@ public class DcacheResourceFactory
                 try {
                     connection.setRequestProperty("Connection", "Close");
                     if (range != null) {
-                        connection.addRequestProperty("Range",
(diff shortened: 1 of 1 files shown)


kofemann:
@mzur ~~Thanks for the contribution. Could you please sign your commit with "Signed-off-by: ..." as described at https://github.com/dCache/dcache/blob/master/CONTRIBUTING.md#external-contribution-formalities~~

kofemann:
Ah, it's already signed! Thanks!

PR #8055, opened 2026-03-27
pool: add firefly onStart marker for MoverProtocol-based transfers

ShawnMcKee:
## Problem

`AbstractMoverProtocolTransferService`, the base class for `MoverProtocol`-based transfer services (used by `RemoteHttpTransferService` for FTS third-party copy movers), never calls `transferLifeCycle.onStart()`. This means **firefly flow-start UDP markers are never emitted for TPC movers**, even though:

- `DefaultPostTransferService` correctly sends `onEnd()` markers (using the `deriveLocalEndpoint()` fallback added in 11.2.3)
- `NettyTransferService` (direct client connections via xrootd/HTTP) correctly calls `onStart()` in its `executeMover()` method

## Impact

ESnet Stardust dashboards show **incomplete flow information for TPC transfers**. Specifically:
- TPC `onEnd` markers are present with correct experiment/activity IDs from transfer tags (e.g., activityId=11, 15, 16 for ATLAS categories)
- **Zero TPC `onStart` markers** are emitted from pool nodes for TPC transfers
- Direct-read `onStart` markers (from `NettyTransferService`) are present but use FQAN fallback (`activityId=1`) since worker nodes don't send SciTag headers

Evidence from ATLAS AGLT2 pool logs:
```
# 194 TPC end markers with correct transfer-tag classifier
scitags event=marker state=end protocol=remotehttpsdatatransfer ... transferTag=143 experimentId=2 activityId=15 classifier=transfer-tag

# 0 TPC start markers
grep -c 'protocol=remote.*state=start' → 0

# 141 direct-read start markers, all FQAN/activityId=1
scitags event=marker state=start protocol=https ... experimentId=2 activityId=1 classifier=fqan
```

## Fix

Add a `startTransferLifeCycle()` call in `MoverTask.run()` (before I/O begins) that:

1. Checks `_transferLifeCycle` is configured and protocol is IP-based
2. Derives the local endpoint from the remote address using `NetworkUtils.getLocalAddress()` — the same approach used by `DefaultPostTransferService.deriveLocalEndpoint()` for `onEnd`
3. Calls `transferLifeCycle.onStart()` with the remote endpoint, derived local endpoint, protocol info, and subject
4. Handles `SocketException` gracefully with debug logging

This mirrors the pattern already established in `NettyTransferService.executeMover()` for direct connections, extending it to all `MoverProtocol`-based transfers.

## Testing

- Compilation verified (`mvn -pl modules/dcache -am compile -DskipTests`)
- Consistent with existing patterns in `NettyTransferService` and `DefaultPostTransferService`

Signed-off-by: Shawn McKee <smckee@umich.edu>

### DIFF
--- a/modules/dcache/src/main/java/org/dcache/pool/classic/RemoteHttpTransferService.java
+++ b/modules/dcache/src/main/java/org/dcache/pool/classic/RemoteHttpTransferService.java
@@ -19,6 +19,7 @@ package org.dcache.pool.classic;
 
 import com.google.common.base.Splitter;
 import diskCacheV111.util.CacheException;
+import diskCacheV111.vehicles.PoolIoFileMessage;
 import diskCacheV111.vehicles.ProtocolInfo;
 import diskCacheV111.vehicles.RemoteHttpDataTransferProtocolInfo;
 import diskCacheV111.vehicles.RemoteHttpsDataTransferProtocolInfo;
@@ -58,8 +59,11 @@ import org.apache.http.impl.client.DefaultRedirectStrategy;
 import org.apache.http.impl.client.HttpClients;
 import org.apache.http.protocol.HttpContext;
 import org.apache.http.protocol.HttpRequestExecutor;
+import org.dcache.pool.movers.Mover;
 import org.dcache.pool.movers.MoverProtocol;
+import org.dcache.pool.movers.MoverProtocolMover;
 import org.dcache.pool.movers.RemoteHttpDataTransferProtocol;
+import org.dcache.pool.repository.ReplicaDescriptor;
 import org.dcache.security.trust.AggregateX509TrustManager;
 import org.dcache.util.Version;
 import org.slf4j.Logger;
@@ -68,6 +72,8 @@ import org.springframework.beans.factory.annotation.Value;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static dmg.util.Exceptions.meaningfulMessage;
+
+import dmg.cells.nucleus.CellPath;
 
 public class RemoteHttpTransferService extends SecureRemoteTransferService {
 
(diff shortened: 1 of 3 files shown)


kofemann (changes requested):
I think that suggested change is wrong. It might preserve the pool IP, but (a) uses the wrong port and (b) will work incorrectly on multi-home hosts.

I would suggest updating the `RemoteHttpTransferService` and passing the `TransferLifeCycle` to the mover when it is created by the `createMoverProtocol` method. The `RemoteHttpDataTransferProtocol#doGet` then should call `onStart` as soon as the `local endpoint` is defined.

modules/dcache/src/main/java/org/dcache/pool/movers/RemoteHttpDataTransferProtocol.java
```diff
@@ -213,9 +216,22 @@ private enum HeaderFlags {
     private Long _expectedTransferSize;
 
     private InetSocketAddress _localEndpoint;
+    private final TransferLifeCycle _transferLifeCycle;
+    private Subject _subject;
+    private boolean _startMarkerSent;
 
     public RemoteHttpDataTransferProtocol(CloseableHttpClient client) {
```
kofemann: This constructor can be removed as all places that create a new instance of RemoteHttpDataTransferProtocol after the proposed change always provide a TransferLifeCycle

modules/dcache/src/main/java/org/dcache/pool/movers/RemoteHttpDataTransferProtocol.java
```diff
@@ -213,9 +216,22 @@ private enum HeaderFlags {
...
+        this(client, null);
+    }
+
+    public RemoteHttpDataTransferProtocol(CloseableHttpClient client,
+          @Nullable TransferLifeCycle transferLifeCycle) {
```
kofemann: With a comment above, the `transferLifeCycle` can't be null

modules/dcache/src/main/java/org/dcache/pool/movers/RemoteHttpDataTransferProtocol.java
```diff
@@ -213,9 +216,22 @@ private enum HeaderFlags {
...
         _client = requireNonNull(client);
+        _transferLifeCycle = transferLifeCycle;
```
kofemann: thus, `requireNonNull(transferLifeCycle)`

modules/dcache/src/main/java/org/dcache/pool/classic/RemoteHttpTransferService.java
```diff
@@ -128,6 +134,17 @@ public HttpUriRequest getRedirect(final HttpRequest request,
     private X509TrustManager trustManager;
     private CloseableHttpClient sharedClient;
 
+    @Override
+    public Mover<?> createMover(ReplicaDescriptor handle, PoolIoFileMessage message,
+          CellPath pathToDoor) throws CacheException {
+        Mover<?> mover = super.createMover(handle, message, pathToDoor);
+        if (mover instanceof MoverProtocolMover mpm
```
kofemann: I would just cast the `mover` to `RemoteHttpDataTransferProtocol`; any other return type indicates a code bug.

kofemann (changes requested):
looks like we are almost there!

kofemann:
@ShawnMcKee, thanks for the contribution. The updated PR looks good.

kofemann (approved):

Issue #8142, opened 2026-07-06
REST data collection frequency


calestyo:
Hey.

If I recall correctly, then Tigran said during my talk in Amsterdam, that the REST would internally query all data, when it is itself queried (via a HTTP `GET`) - like every time fresh data.

I just tried a bit and it seems that at least the `/pools/{name}/usage` data is however collected by dCache periodically once every 60s and afterwards, it returns always the same data.

E.g. the `mode=…` output of  `psu ls pool -l` updates basically instantaneously after I change a pool's mode, while the corresponding `poolMode` in the REST JSON takes well up to 60s.

I've also seen that there's:
```
# Timeout for transfer info collection
frontend.service.transfers.timeout=1
(one-of?MILLISECONDS|SECONDS|MINUTES|HOURS|DAYS)frontend.service.transfers.timeout.unit=MINUTES
frontend.service.transfers.maxCacheSize=1000

# Timeout for cell info collection
frontend.service.cell-info.timeout=1
(one-of?MILLISECONDS|SECONDS|MINUTES|HOURS|DAYS)frontend.service.cell-info.timeout.unit=MINUTES

# Timeout for restore info collection
frontend.service.restores.timeout=1
(one-of?MILLISECONDS|SECONDS|MINUTES|HOURS|DAYS)frontend.service.restores.timeout.unit=MINUTES
frontend.service.restores.maxCacheSize=1000

# Timeout for alarms info collection
frontend.service.alarms.timeout=1
(one-of?MILLISECONDS|SECONDS|MINUTES|HOURS|DAYS)frontend.service.alarms.timeout.unit=MINUTES

# Timeout for pool info collection
frontend.service.pool-info.timeout=1
(one-of?MILLISECONDS|SECONDS|MINUTES|HOURS|DAYS)frontend.service.pool-info.timeout.unit=MINUTES

# Timeout for pool history info collection
frontend.service.pool-history.timeout=1
(one-of?MILLISECONDS|SECONDS|MINUTES|HOURS|DAYS)frontend.service.pool-history.timeout.unit=MINUTES
```
And indeed, changing e.g. `frontend.service.pool-info.timeout` gives me faster updates. 🥳️

Now:
1. Can I somehow find out which of the properties correspond to which REST paths?
    I guess:
    - `/cells` is controlled by `frontend.service.cell-info.timeout` only
    - `/domains` by `frontend.service.domain-info.timeout` only
[...]

PR #8150, opened 2026-07-12
doc improvements

calestyo:
Hey.

As for 34827be79b95b488cea8a96a133b23d07ec182c4, please check someone whether there’s really no such command - I couldn't find it, but maybe `pools refresh` is simply shared by pool-info and pool-history.

As for 1e51e8b0a4dab61565a8b75cba53b8a7d0807d21, this is based on @DmitryLitvintsev’s comments and my own look at the code (though I had a hard time to actually understand which paths are affected by which timeout properties).
The problem with it is that it's incomplete:
- I haven't added the path infos to options at all, where I don't use the path in my prometheus exporter.
- Even for those that I've added, I'm not 100% sure whether the info is complete.
   in particular:
   - Are any others paths (which I don't list) affected by each of the options, e.g. `/pools/{name}/usage` somewhere contains a copy of the `/cells` data for that pool, but that for example does **not** seem to be affected by `frontend.service.cell-info.timeout` but rather by `frontend.service.pool-info.timeout`. Or `frontend.service.pool-info.timeout` might also affect `/poolgroups/`
   - Do I list any paths, where subpaths are however **not** affected by the respective option? E.g. `/pools/` contains a lot, but I've only really checked that `frontend.service.pool-info.timeout` affects `/pools/{name}/usage`.
   - Also I'm not sure whether the list for `frontend.service.pool-history.timeout` is complete/exact.

So maybe this should rather be only a draft, but in general I think these options are less useful if it's not really clear what **exactly** they do.

### DIFF
--- a/skel/share/defaults/frontend.properties
+++ b/skel/share/defaults/frontend.properties
@@ -111,6 +111,8 @@ frontend.service.transfers.timeout=1
 frontend.service.transfers.maxCacheSize=1000
 
 # Timeout for cell info collection
+#
+# Amongst others, this affects the `/cells`-resource of the REST interface.
 # These properties can also be set interactively through the admin door
 frontend.service.cell-info.timeout=1
 (one-of?MILLISECONDS|SECONDS|MINUTES|HOURS|DAYS)frontend.service.cell-info.timeout.unit=MINUTES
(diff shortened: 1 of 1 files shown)


calestyo:
See also https://github.com/dCache/dcache/issues/8142

PR #8213, opened 2026-09-17
voms: fix FQAN parsing for root /vo group

### LINKED ISSUE #8211: VomsPlugin rejects valid root-group FQANs such as /vo since 8de56cb
Since commit 8de56cb83948d6d55062845e69a15a896fd8ca42, VomsPlugin only accepts FQANs starting with:
```
"/" + voName + "/"
```

This rejects valid root-group FQANs such as:
```/biomed
```

while accepting:
```
/biomed/group
/biomed/Role=...
```

According to the VOMS specification ([GFD.182](https://ogf.org/documents/GFD.182.pdf)), `/vo` is a valid FQAN.

This causes problems with INDIGO IAM/VOMS-AA instances using the compact FQAN encoding. For example, a proxy containing:
```
/biomed
/biomed/lcg1
/biomed/team
```

results in `/biomed` being silently discarded by dCache, which may break gPlazma mappings based on the VO root group.

The issue is mostly hidden for VOs using legacy FQAN encoding, where the root FQAN is represented as:
```
/cms/Role=NULL/Capability=NULL
```

Pansanel:
Accept root VO FQANs such as:
```
/biomed
```

in addition to FQANs below the VO root, such as:
```
/biomed/group
/biomed/Role=member
```

Root-group FQANs are valid according to GFD.182, but were rejected because VomsPlugin required FQANs to start with `/vo/`.

The new check accepts either an exact `/vo` match or an FQAN starting with `/vo/`, while still preventing matches with other VO namespaces such as `/biomed2`.

Fixes #8211

Signed-off-by: Jérôme Pansanel <jerome.pansanel@iphc.cnrs.fr>


kofemann:
Hi @Pansanel , thanks for the contribution. Could you please update the patch with `Siggned-of-by:` tag, as described in: https://github.com/dCache/dcache/blob/master/CONTRIBUTING.md#external-contribution-formalities

Pansanel:
@kofemann : I have added the tag in the description. Is that fine?

